PR #17, opened 2024-01-12
feat: adding TaskRegistry

axel7083:
- Adding central place to register tasks
- Adding labels to tasks etc.

### DIFF
--- a/packages/frontend/src/pages/Models.svelte
+++ b/packages/frontend/src/pages/Models.svelte
@@ -41,8 +41,8 @@ function filterModels(): void {
     if(currentValue.state === 'success')
       return previousValue;
 
-    if(currentValue.labels) {
-      previousValue.push(currentValue.labels["model-pulling"])
+    if(currentValue.labels !== undefined) {
+      previousValue.push(currentValue.labels["model-pulling"]);
     }
     return previousValue;
   }, [] as string[]);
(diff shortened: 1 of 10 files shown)


packages/frontend/src/pages/Models.svelte
```diff
@@ -1,39 +1,106 @@
...
+  new Column<ModelInfo>('License', { width: '1fr', renderer: ModelColumnLicense }),
+];
+const row = new Row<ModelInfo>({});
+
+let loading: boolean = true;
+let intervalId: ReturnType<typeof setInterval> | undefined = undefined;
+
+$: tasks = [] as Task[];
+$: models = [] as ModelInfo[];
+$: filteredModels = [] as ModelInfo[];
```
feloy, line 36: What is the interest of using `$:` here? As there is no dependency, I feel that it is similar to:
```
let tasks = [] as Task[];
let models = [] as ModelInfo[];
let filteredModels = [] as ModelInfo[];
```

packages/frontend/src/pages/Models.svelte
```diff
@@ -1,39 +1,106 @@
...
+  // Let's collect the models we do not want to show (loading, error).
+  const modelsId: string[] = tasks.reduce((previousValue, currentValue) => {
+    if(currentValue.state === 'success')
+      return previousValue;
+
+    if(currentValue.labels !== undefined) {
+      previousValue.push(currentValue.labels["model-pulling"]);
+    }
+    return previousValue;
+  }, [] as string[]);
```
feloy, line 48: I feel it could be easier to read/maintain with a `tasks.filter(...).map(t => t.labels['model-pulling']`

packages/frontend/src/pages/Models.svelte
```diff
@@ -1,39 +1,106 @@
...
+}
+
+onMount(() => {
+  // Pulling update
+  intervalId = setInterval(async () => {
+    tasks = await studioClient.getTasksByLabel("model-pulling");
+    loading = false;
+    filterModels();
+  }, 1000);
+
```
feloy, line 59: We will want to avoid pulling, by using a `store` reactively updated by messages.

packages/frontend/src/pages/Models.svelte
axel7083: I agree on the reading level, but not on the performance side as we are iterating two time instead of one, as this method can be called a lot of time I would prefer adding some comments to improve the readability

packages/frontend/src/pages/Models.svelte
axel7083: I agree and we probably need to wait for https://github.com/projectatomic/studio-extension/issues/24

packages/frontend/src/pages/Models.svelte
axel7083: Yes, you are right, I will fix it !

feloy (approved):

PR #22, opened 2024-01-15
doc: State management proposal

feloy:
This is a proposal on how to manage the different states in AI Studio:

- catalog
- pulled applications
- errors
- etc

(as an example, the management of the state of the kube contexts in Podman Desktop is implemented in a similar way, in PR https://github.com/containers/podman-desktop/pull/5329).

### DIFF
--- a/docs/proposals/state-management.md
+++ b/docs/proposals/state-management.md
@@ -1,0 +1,52 @@
+# State management
+
+The backend manages and persists the State. The backend pushes new state to the front-end 
+when changes happen, and the front-end can ask for the current value of the state.
+
+The front-end uses `readable` stores to expose the state to the different pages. The store 
+listens for new states pushed by the backend (`onMessage`), and asks for the current state
+at initial time.
+
+The pages of the front-end subscribe to the store to get the value of the state in a reactive manner.
+
+## Catalog
+
+The catalog is persisted as a file in the user's filesystem. The backend reads the file at startup,
+and watches the file for changes. The backend updates the state as soon as changes it detects changes.
+
+The front-end uses a `readable` store, which waits for changes on the Catalog state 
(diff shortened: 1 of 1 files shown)


benoitf (approved):
👍 to have state managed on the backend

Probably would need a high-level API for handling store update, etc but basically it can be done after this extension is done so we know what should be the common stuff to extract

axel7083:
I agree with the idea of having the back managing states and front subscribing to those.

However, I would rather avoid having to replicates the hardcoded string for channels just like the podman-desktop project is doing. 

Having **a lot** of string on the front/back to specify the channel to use is a bit annoying for a few reasons

- Makes it hard to understand, we usually need to rely on text search to find where is the implementation we call
- Lack of typing, the developers is responsible of ensuring the channels and methods signature is matching on the front and back, thus leading to potentials mistakes. 
- Hard to test 

The `packages/shared/MessageProxy.ts` typescript code was a tentative at avoiding that, but it  is very limtied as it only support basic RPC. 

We clearly need to remove the `pulling update` system as it a bad design pattern in our context, I implemented it to have a quick and working experience in our time frame. 

> Probably would need a high-level API for handling store update, etc but basically it can be done after this extension is done so we know what should be the common stuff to extract

@benoitf I wonder if we should avoid postponing this and try to have a better level api, as pushing back would add debt, we would have to deal with later.

benoitf:
@axel7083 the idea is to standardize on real use-case and not try to make a standard before it's being really used

so while it can be technical debt, it avoids to try to design a pattern that it won't be used or will be a false solution and once it's inside the code we need to maintain it for backward compatibility. So it's easier to see what extensions do, what they need, if it's a well designed approach, etc. And then put it back to the core part

PR #58, opened 2024-01-16
fix: separate models from recipes

lstocchi:


feloy:
I amde a first test, but the description is not displayed in the Playground page, and I get an error with `model not found` when I try to send a query to the model (I'll have a look to the code to check where it fails)

packages/shared/StudioAPI.ts
```diff
@@ -11,7 +11,7 @@ export abstract class StudioAPI {
   abstract getCategories(): Promise<Category[]>;
   abstract getRecipesByCategory(categoryId: string): Promise<Recipe[]>;
   abstract getRecipeById(recipeId: string): Promise<Recipe>;
-  abstract getModelById(modelId: string): Promise<ModelInfo>;
+  abstract getModelById(modelId: string): ModelInfo;
```
feloy, line 14: This needs to be returning a Promise, as the value will be returned by exchanging messags

lstocchi:
@feloy thanks for pointing me to the problem. I should have fixed everything. Please give it a look 👍

feloy (approved):
works great, thanks

PR #64, opened 2024-01-16
Playground state

feloy:
With this PR, in a Playground page, the query and the response are decorrelated.

- the frontend sends a query and gets a queryId in return
- the backend waits for the query to terminate, and updates the queries state with the result of the query
- the frontend listens for the query state (through the playgroundQuery store), and updates the page with the result.
- if the user leaves the Playground page and goes back later to the page, the latest query is displayed (either it is terminated or waiting for a response)

Other related changes:
- the frontend can subscribe to messages from the RpcBrowser (which receives all messages from backend)
- the container for the playground is not restarted if already started
- a new port is used for every new playground's container, it should be possible to have several playground's containers running at the same time (to be tested)

TODO
- [x] tests


slemeur:
👍

packages/backend/src/studio-api-impl.ts
```diff
@@ -111,11 +111,15 @@ export class StudioApiImpl implements StudioAPI {
...
     const localModelInfo = this.applicationManager.getLocalModels().filter(m => m.id === modelId);
     if (localModelInfo.length !== 1) {
       throw new Error('model not found');
     }
     return this.playgroundManager.askPlayground(localModelInfo[0], prompt);
   }
+
+  async askPlaygroundQueries(): Promise<void> {
+    this.playgroundManager.sendState();
+  }
```
axel7083, line 124:
```suggestion
  async getPlaygroundStates(): Promise<QueryState[]> {
    return this.playgroundManager.getState();
  }
```

packages/backend/src/playground.ts
```diff
@@ -69,42 +97,72 @@ export class PlayGroundManager {
...
+  }
+
+  getNextQueryId() {
+    return ++this.queryIdCounter;
+  }
+  
+  sendState() {
+    this.webview.postMessage({ 
+      id: MSG_NEW_PLAYGROUND_QUERIES_STATE, 
+      body: Array.from(this.queries.values()),
```
axel7083, line 165:
```suggestion
      body: this.getState(),
```

packages/backend/src/playground.ts
```diff
@@ -69,42 +97,72 @@ export class PlayGroundManager {
...
+
+    this.queries.set(query.id, query);
+    this.sendState();
+    return query.id;
+  }
+
+  getNextQueryId() {
+    return ++this.queryIdCounter;
+  }
+
```
axel7083, line 161:
```suggestion
  getState(): QueryState[] {
    return Array.from(this.queries.values());
  }
```

packages/shared/StudioAPI.ts
```diff
@@ -23,12 +23,17 @@ export abstract class StudioAPI {
...
   /**
    * Get task by label
    * @param label
    */
   abstract getTasksByLabel(label: string): Promise<Task[]>;
+
+  /**
+   * Ask to send a message MSG_NEW_PLAYGROUND_QUERIES_STATE with the current Playground queries
+   */
+  abstract askPlaygroundQueries(): Promise<void>;
```
axel7083, line 37:
```suggestion
  abstract getPlaygroundStates(): Promise<QueryState[]>;
```

packages/frontend/src/stores/playground-queries.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { QueryState } from '@shared/models/IPlaygroundQueryState';
+import { rpcBrowser } from '../utils/client';
+import { MSG_NEW_PLAYGROUND_QUERIES_STATE } from '@shared/Messages';
+import { studioClient } from '/@/utils/client';
+
+export const playgroundQueries: Readable<QueryState[]> = readable<QueryState[]>([], (set) => {  
+  const sub = rpcBrowser.subscribe(MSG_NEW_PLAYGROUND_QUERIES_STATE, (msg) => {
+    set(msg);
+  });
+  studioClient.askPlaygroundQueries();
```
axel7083, line 12:
```suggestion
  // Initialize the store manually 
  studioClient.getPlaygroundStates().then((state) => {
    set(state);
  });
```

packages/shared/MessageProxy.ts
```diff
@@ -153,7 +166,23 @@ export class RpcBrowser {
...
     // Create a Promise
     return new Promise((resolve, reject) => {
-      this.promises.set(requestId, {resolve, reject});
+      this.promises.set(requestId, { resolve, reject });
     })
   }
+
+  // TODO(feloy) need to subscribe several times?
+  subscribe(msgId: string, f: (msg: any) => void): Subscriber {
+    console.log('add subscriber');
```
axel7083, line 175:
```suggestion
```

packages/shared/MessageProxy.ts
```diff
@@ -153,7 +166,23 @@ export class RpcBrowser {
...
+    this.subscribers.set(msgId, f);
+    return {
+      unsubscribe: () => {
+        console.log('delete subscriber');
```
axel7083, line 179:
```suggestion
```

axel7083 (approved):
Just an idea, instead of having a method firing the first event to setup the store, we could make it return the state, and initiate the store with it, it would look less _magic_ to read

lstocchi (commented):
Idk if it would be part of this PR or a separate one but the playground does not work for me. I don't have the local-ai image and i don't know what else i'm missing (still have to check the code) but as a user i clicked `send request` and nothing happens. If i look at the console i see 
image

feloy:
> Idk if it would be part of this PR or a separate one but the playground does not work for me. I don't have the local-ai image and i don't know what else i'm missing (still have to check the code) but as a user i clicked `send request` and nothing happens. If i look at the console i see image

I think it is due to the fact that you don't have the local-ai image.
I have created this issue to follow the problem: https://github.com/projectatomic/studio-extension/issues/87

lstocchi (approved):
After having downloaded local-ai, it works as expected. GJ 🚀

packages/backend/src/playground.ts
```diff
@@ -15,13 +24,33 @@ function findFirstProvider(): ProviderContainerConnection | undefined {
...
+    this.queries = new Map<number, QueryState>();
+  }
+
   async selectImage(connection: ProviderContainerConnection, image: string): Promise<ImageInfo | undefined> {
     const images = (await containerEngine.listImages()).filter(im => im.RepoTags?.some(tag => tag === image));
     return images.length > 0 ? images[0] : undefined;
   }
 
   async startPlayground(modelId: string, modelPath: string): Promise<string> {
+    // TODO(feloy) remove previous query from state?
```
axel7083, line 49: todo ?

feloy: not as part of this PR

packages/shared/src/MessageProxy.ts
```diff
@@ -183,4 +196,24 @@ export class RpcBrowser {
       this.promises.set(requestId, { resolve, reject });
     });
   }
+
+  // TODO(feloy) need to subscribe several times?
```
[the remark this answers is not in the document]
feloy: not as part of this PR

Issue #36, opened 2024-01-15
feat: convert markdown from README properly 


axel7083:
When copy past README content, a lot of issue may happen. First most of markdown in GitHub for instance use relative links for images etc. It need to be converted so we can display them.

PR #65, opened 2024-01-16
fix: markdown broken links

axel7083:
Related to #36 

Depends on https://github.com/projectatomic/studio-extension/pull/63

The markdown displayed has links, that messes with the webview routing system. Creating custom renderer for Link objects and opening externally the links when clicked.


feloy (approved):
Tested by adding a complete URL in some markdown, works fine. Thanks

PR #66, opened 2024-01-16
fix: removing search bar as logic not implemented

### LINKED ISSUE #25: fix: remove search bar where it is not necessary

axel7083:
Fix #25


feloy (approved):
When we work on externalizing the UI components, we will have to check if the NavPage layout is the good one to use here. For now, it seems ok

slemeur:
🚀

PR #72, opened 2024-01-17
fix: add ci for test, lint, format and typecheck

lstocchi:


slemeur:
🚀

feloy:
It seems you will have to run this command locally, to get some files correctly formatted (beware, files will be modified in place):

```
./node_modules/.bin/prettier -w --check "**/src/**/*.ts"
```

.github/workflows/pr-check.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+jobs:
+  lint-format-unit:
+    name: linter, formatters and unit tests / ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 40
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-2022, ubuntu-22.04]
```
benoitf, line 30: line 42 we see mac/Linux but here there is only Ubuntu, no macOS so I don't know if you omitted macOS on purpose or not

.github/workflows/pr-check.yaml
```diff
@@ -0,0 +1,87 @@
...
+        run: yarn typecheck
+
+      # Check we don't have changes in git
+      - name: Check no changes in git
+        if: ${{ matrix.os=='ubuntu-22.04'}}
+        run: |
+          if ! git diff --exit-code; then
+            echo "Found changes in git"
+            exit 1
+          fi
```
benoitf, line 87: looks like the file is not formatted (EOF missing)

package.json
```diff
@@ -16,9 +16,16 @@
     "format:fix": "prettier --write \"**/src/**/*.ts\"",
     "lint:check": "eslint . --ext js,ts,tsx",
     "lint:fix": "eslint . --fix --ext js,ts,tsx",
-    "test:backend": "vitest run -r packages/backend --passWithNoTests --coverage"
+    "test:backend": "vitest run -r packages/backend --passWithNoTests --coverage",
+    "test:shared": "vitest run -r packages/shared --passWithNoTests --coverage",
```
benoitf, line 20: it is missing frontend tests, correct ?

benoitf (approved):

.github/workflows/pr-check.yaml
lstocchi: i copied the portion i needed from the podman-desktop repo. I don't remember if there was some reasoning behind it. But i guess i can add the mac machine as well for what we do atm

lstocchi: Done 👍

packages/frontend/package.json
```diff
@@ -29,10 +30,11 @@
     "@testing-library/user-event": "^14.5.1",
     "@tsconfig/svelte": "^5.0.2",
     "@typescript-eslint/eslint-plugin": "6.15.0",
+    "jsdom": "^23.2.0",
     "svelte": "4.2.8",
     "svelte-preprocess": "^5.1.3",
     "tailwindcss": "^3.4.0",
     "postcss": "^8.4.33",
-    "postcss-load-config": "^5.0.2"
+    "postcss-load-config": "^5.0.2"
```
benoitf, line 38: it looks like the file is not formatted/sorted (usually it is ordered in alphabetical order) and there is an extra space here

lstocchi: how do you sort it automatically? Is there any npm package? 🤔

feloy: I think it is done when you run `yarn add <package>`

lstocchi: I was hoping we could add some script so that if someone copy/paste an entry it is sorted anyway

benoitf: `npx sort-package-json`

benoitf: (anyway it's not a requested change)

lstocchi: cool, i'm going to add it in a follow-up PR so i don't block @feloy 
for the moment i fixed it manually

Issue #32, opened 2024-01-15
catalog: store it as file accessible/editable by user


feloy:
The catalog can be stored as a file, and can be placed in the extension's data directory, when the extension is installed.

Ai Studio should be able to detect changes on this file and update the catalog accordingly.

feloy:
fixed by #85 and #92

PR #85, opened 2024-01-17
Load user catalog if file exists or use default one

feloy:
A default catalog is provided with the extension. The user can provide its own catalog (in ~/podman-desktop/ai-studio/catalog.json). If the user's file is valid, it will be used. Otherwise, the default catalog will be used.

For now, changes on the user's file are not detected.

Fixes partially #32

### DIFF
--- a/packages/backend/src/studio-api-impl.spec.ts
+++ b/packages/backend/src/studio-api-impl.spec.ts
@@ -18,13 +18,16 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 import content from './ai-test.json';
+import userContent from './ai-user-test.json';
 import type { ApplicationManager } from './managers/applicationManager';
 import type { RecipeStatusRegistry } from './registries/RecipeStatusRegistry';
 import { StudioApiImpl } from './studio-api-impl';
 import type { PlayGroundManager } from './playground';
 import type { TaskRegistry } from './registries/TaskRegistry';
+
+import * as fs from 'node:fs';
 
 vi.mock('./ai.json', () => {
   return {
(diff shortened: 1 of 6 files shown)


packages/backend/src/studio-api-impl.ts
```diff
@@ -1,27 +1,37 @@
...
+
   constructor(
     private applicationManager: ApplicationManager,
     private recipeStatusRegistry: RecipeStatusRegistry,
     private taskRegistry: TaskRegistry,
     private playgroundManager: PlayGroundManager,
-  ) {}
+  ) {
+    const catalogPath = path.resolve(__dirname, '..', 'catalog.json');
+    const catalogStr = fs.readFileSync(catalogPath, 'utf-8');
```
benoitf, line 31: I'm wondering here if you need to use a read of the file as it seems anyway you embed it into the extension

so could it like 
```
import catalog from '../catalog.json'
```

or with newer spec

```ts
import catalog from './catalog.json' with { type: 'json' };
```

so you don't need to add it in Containerfile, do not need to handle synchronous method/io like readFileSync instead of promises readFile, etc

benoitf (commented):
looks like there are quotes/single quote issues

packages/backend/src/studio-api-impl.ts
feloy: Probably, but I'll want in the next iteration to run a watch on this file, to detect changes, and I prefer to go one step in that direction

benoitf: note that if it's imported, vite will automatically reload the file as well and update the extension's code

benoitf: (so no need to setup a custom watcher for development mode)

feloy: But what about the production code? I would like the extension to detect changes even when the app is running in production mode

benoitf: in the current code the file is embedded and read from the extension so it doesn't change the behaviour.

if you need to read from a custom folder, you'll still need to include a default file in case it's not yet provided by the user

and in production, anyway you should not write to the source directory and monitor the source directory. You'll need to copy/read the file from a different folder (extension storage folder) b/c here if you update the extension it'll override the file anyway.

so as you need to have an 'included' version anyway it's easier to import/include it as a resource of the extension rather than a separate file.

feloy: ok, you are right, I didn't think about having an 'included' version. 
I'll proceed differently then.
Thanks

benoitf: but it's not a requested change on my side, etc.

feloy: I changed the scenario to use the imported json file as default catalog if the user does not provide any catalog.

packages/backend/src/studio-api-impl.ts
```diff
@@ -1,28 +1,72 @@
...
+          } catch (err: unknown) {
+            console.error('unable to parse catalog file, reverting to default catalog', err);
+            this.setNewCatalog(defaultCatalog);
+          }
+        })
+        .catch((err: unknown) => {
+          console.error('got err', err);
+          console.error('unable to read catalog file, reverting to default catalog', err);
+          this.setNewCatalog(defaultCatalog);
+        });
```
benoitf, line 59: not sure about chaining promises with the await

```ts
try {
  const data = await fs.promises.readFile(catalogPath, 'utf-8')
  const cat = JSON.parse(data) as Catalog;
  this.setCatalog(cat);
} catch (
  console.error(...)
  this.setCatalog(defaultCatalog);
}
```

I would check if the file exists before trying to read it as well.

packages/backend/src/studio-api-impl.ts
```diff
@@ -1,28 +1,72 @@
...
+    } catch (err: unknown) {
+      console.error('unable to read catalog file, reverting to default catalog', err);
+      this.setNewCatalog(defaultCatalog);
+    }
+  }
+
+  setNewCatalog(newCatalog: Catalog) {
```
benoitf, line 66:
```suggestion
  setCatalog(newCatalog: Catalog) {
```

packages/backend/src/studio-api-impl.ts
```diff
@@ -41,31 +85,31 @@ export class StudioApiImpl implements StudioAPI {
   }
 
   async getCategories(): Promise<Category[]> {
-    return content.categories;
+    return this.catalog.categories;
   }
```
benoitf, line 89: not sure why it's a Promise ?

benoitf: (same remark for almost all methods)

as catalog is a model, I'm not sure about the usage of async/promises for reading from it

packages/backend/src/studio-api-impl.spec.ts
```diff
@@ -18,63 +18,122 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 import content from './ai-test.json';
+import userContent from './ai-user-test.json';
```
lstocchi, line 23: couldn't you just use the `ai-test.json` file? So we avoid having multiple files for just testing.
In my case i just check the name, registry, url. So if you prefer your content it's just a matter of updating few lines in the old tests and removing the other json.

packages/backend/src/studio-api-impl.ts
feloy: It's much simpler, thanks :)

packages/backend/src/studio-api-impl.ts
feloy: I think they are async to respect the interface (class) defined in `shared/src/StudioAPI.ts`, where they are returning a Promise, and from the client side, they are really returning a Promise, due to the wrapper sending a message.

packages/backend/src/studio-api-impl.ts
feloy: IMHO, readFile correctly and efficiently checks the existence of the file, we don't have to check it before (or we would have to handle another error case)

packages/backend/src/studio-api-impl.spec.ts
feloy: What I want to test is that the user's catalog is used, when it is present. For this, I need a second, different, catalog, to be sure by reading it that it has been loaded.

packages/backend/src/studio-api-impl.ts
benoitf: ok got it.

I thought that client was getting the model object asynchronously but then it was just fields

client:
```ts
const catalog = await rpc.get(CatalogAPI.class).getCatalog();
const categories = catalog.categories;
const something = catalog.something;
const else = catalog.else;
```

and not a call each time we need something like
```ts
const categories = await rpc....getCategories();
const something = await rpc....getSomething();
const else = await rpc....getElse();

File: packages/backend/src/studio-api-impl.ts (hunk 2 above)
Comment (benoitf) [reply packages/backend/src/studio-api-impl.ts]: it's mainly b/c usually the absence of the file is not an error so you want to read it if you know user provided a custom file

and getting an exception b/c the file is not there or getting an exception b/c we're unable to read the file is not the same while here we handle it

it looks like we'll get a log trace

```
unable to read catalog file, reverting to default catalog
```

even if we don't have customized the file

packages/backend/src/studio-api-impl.ts
feloy: What I expect to do with the next PR is to have the backend (this class) post a message to the frontend with the new content of the catalog every time it changes (`in newCatalog()`). On the frontend, I wonder what would be the best approach: having a single `catalog` store, or 3 (or more) stores: `recipes`, `models`, `categories`.

Or, even, having 3 different messages (newRecipes, newModels, newCategories)? But inihs case we would have to check which ones have been modified, to push only the modified ones.

WDYT?

packages/backend/src/studio-api-impl.ts
feloy: I agree about the error log trace. Let's add the check and be silent if the file doesn't exist

lstocchi (approved):

packages/backend/src/studio-api-impl.spec.ts
lstocchi: My bad, by opening it on vscode and seeing the code without any annotation it was clear 👍

PR #88, opened 2024-01-17
fix: frontend loading issue

axel7083:
## Description

Instead of having individual page adding the LinearProgress, I added it to the NavPage component, and a new loading prop.
When loading is set to true, the linear progress is used, and the content is hidden.

## Screenshots

https://github.com/projectatomic/studio-extension/assets/42176370/de2c66b1-1b20-4659-9903-d96861f9aacc


lstocchi (approved):
LGTM, a test would be great

lstocchi (approved):
LGTM, thanks!!

PR #91, opened 2024-01-18
fix: proxy typing + copyrights

axel7083:
This PR includes

- This PR will __enforce__ at compilation the methods declared in the StudioAPI are under the format 
```ts
(...args: unknown[]): Promise<unknown>
```
- Move the `playground.ts` to the manager directory


lstocchi (approved):
Needs to be rebased.

feloy (approved):

axel7083:
Fix for https://github.com/projectatomic/studio-extension/issues/95 added

packages/backend/src/managers/catalogManager.ts
```diff
@@ -0,0 +1,53 @@
+import type { Catalog } from '@shared/src/models/ICatalog';
```
benoitf, line 1: FYI it's missing copyright headers

packages/backend/src/managers/catalogManager.ts
```diff
@@ -0,0 +1,53 @@
...
+import path from 'node:path';
+import fs from 'node:fs';
+import defaultCatalog from '../ai.json';
+import type { Category } from '@shared/src/models/ICategory';
+import type { Recipe } from '@shared/src/models/IRecipe';
+import type { ModelInfo } from '@shared/src/models/IModelInfo';
+
+export class CatalogManager {
+  private catalog: Catalog;
```
benoitf, line 10: to make it really private you could use #

```suggestion
  #catalog: Catalog;
```

axel7083: Yeah but for readability, user not knowing what # mean, will not think it is priovate, can we use both ?
```
private #catalog: Catalog
```

benoitf: no you can't

axel7083: Gonna keep private for now as we are not yet using the #

PR #92, opened 2024-01-18
push the catalog to the frontend and use a store on frontend

feloy:
This PR replaces the different RPCs to a synchro of the catalog between backend and frontend. 

The frontend now knows the frontend structure and can make requests on it.

If the structure of the catalog becomes complex in the future, we will probably want to provide a Catalog class on the frontend to centralize the logic, or add stores derived from the `catalog` one.

TODO
- [x] watch catalog file and update the catalog when changes occur
- [ ] Add tests to frontend pages using catalog

### DIFF
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -38,19 +38,54 @@ export class StudioApiImpl implements StudioAPI {
 
   async loadCatalog() {
     const catalogPath = path.resolve(this.applicationManager.appUserDirectory, 'catalog.json');
+
+    try {
+      this.watchCatalogFile(catalogPath); // do not await, we want to do this async
+    } catch (err: unknown) {
+      console.error("unable to watch catalog file, changes to the catalog file won't be reflected to the UI", err);
+    }
+
     try {
       if (!fs.existsSync(catalogPath)) {
         await this.setCatalog(defaultCatalog);
         return;
       }
-      // TODO(feloy): watch catalog file and update catalog with new content
-      const data = await fs.promises.readFile(catalogPath, 'utf-8');
-      const cat = JSON.parse(data) as Catalog;
+      const cat = await this.readAndAnalyzeCatalog(catalogPath);
       await this.setCatalog(cat);
     } catch (err: unknown) {
       console.error('unable to read catalog file, reverting to default catalog', err);
       await this.setCatalog(defaultCatalog);
     }
+  }
+
+  watchCatalogFile(path: string) {
+    const watcher = podmanDesktopApi.fs.createFileSystemWatcher(path);
+    watcher.onDidCreate(async () => {
+      try {
+        const cat = await this.readAndAnalyzeCatalog(path);
+        await this.setCatalog(cat);
+      } catch (err: unknown) {
+        console.error('unable to read created catalog file, continue using default catalog', err);
+      }
+    });
+    watcher.onDidDelete(async () => {
+      console.log('user catalog file deleted, reverting to default catalog');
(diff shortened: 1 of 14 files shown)


packages/frontend/src/pages/Recipes.svelte
```diff
@@ -1,17 +1,10 @@
...
-import type { Category } from '@shared/models/ICategory';
-
-$: categories = [] as Category[]
-
-
-onMount(async () => {
-  categories = await studioClient.getCategories()
-})
+import { RECENT_CATEGORY_ID } from '/@/utils/client';
+import { catalog } from '../stores/catalog';
```
benoitf, line 5:
```suggestion
import { catalog } from '/@/stores/catalog';
```

benoitf (approved):
feel like more robust

packages/frontend/src/lib/RecipesCard.svelte
```diff
@@ -1,26 +1,19 @@
 <script lang="ts">
 import Card from '/@/lib/Card.svelte';
-import type { Recipe } from '@shared/models/IRecipe';
 import { getIcon } from '/@/utils/categoriesUtils';
-import { onMount } from 'svelte';
-import { studioClient } from '/@/utils/client';
-import type { Category } from '@shared/models/ICategory';
+import type { Category } from '@shared/src/models/ICategory';
+import { catalog } from '../stores/catalog';
```
benoitf, line 5:
```suggestion
import { catalog } from '/@/stores/catalog';
```

packages/frontend/src/stores/catalog.ts
```diff
@@ -0,0 +1,24 @@
+import type { Readable } from 'svelte/store';
+import { readable } from 'svelte/store';
+import { MSG_NEW_CATALOG_STATE } from '@shared/Messages';
+import { rpcBrowser, studioClient } from '/@/utils/client';
+import type { Catalog } from '@shared/src/models/ICatalog';
```
benoitf, line 5: it feels odd to me to have ICatalog file with a Catalog and not a ICatalog
same for ITask.ts having Task object

Might be easier to have as convention the ICatalog for interface name as well so we know if we imports somehow an interface or an object

or you name the file Catalog not ICatalog else I don't see the point of prefixing the filename

feloy: @axel7083 what do you think? As you started with this convention, would you prefer to use `I` everywhere (file names and interface names) or remove them from everywhere?

packages/backend/src/studio-api-impl.ts
```diff
@@ -12,9 +11,11 @@ import type { PlayGroundManager } from './playground';
 import * as podmanDesktopApi from '@podman-desktop/api';
 import type { QueryState } from '@shared/src/models/IPlaygroundQueryState';
 import type { Catalog } from '@shared/src/models/ICatalog';
+import { MSG_NEW_CATALOG_STATE } from '@shared/Messages';
 
 import * as path from 'node:path';
 import * as fs from 'node:fs';
+import { watch } from 'chokidar';
```
benoitf, line 18: you should use instead createFileSystemWatcher from Podman Desktop API

packages/frontend/src/stores/catalog.ts
axel7083: I would not mind either ones.

I would have a personal preference for having file with `I` prefix and interface name as it improve readability and expectation

feloy: OK, I'll rename the interfaces to add an `I` prefix, so the file names and interafce names are the same. I'll do this in an upcoming PR

PR #93, opened 2024-01-18
feat: adding support for starting / stopping playground

axel7083:
Fixes https://github.com/projectatomic/studio-extension/issues/20

### Stopped

image

### Running

image

### Starting | Stopping

image


feloy:
I would replace the `replace none`, when no playground is running, by something more meaningful (`Playground is not running`?)

feloy (approved):

axel7083:
> I would replace the `Playground none`, when no playground is running, by something more meaningful (`Playground is not running`?)

Agree but details for now

PR #98, opened 2024-01-19
chore: add missing linting rules from Podman Desktop

benoitf:
fixes issues reported by using the same rules

it also fixes https://github.com/projectatomic/studio-extension/issues/97

avoid to see 
image

### DIFF
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -66,7 +66,7 @@ export class CatalogManager {
     try {
       this.watchCatalogFile(catalogPath); // do not await, we want to do this async
     } catch (err: unknown) {
-      console.error("unable to watch catalog file, changes to the catalog file won't be reflected to the UI", err);
+      console.error(`unable to watch catalog file, changes to the catalog file won't be reflected to the UI`, err);
     }
 
     try {
--- a/packages/backend/src/models/AIConfig.ts
+++ b/packages/backend/src/models/AIConfig.ts
@@ -32,7 +32,7 @@ export interface AIConfig {
 }
 
 export function isString(value: unknown): value is string {
-  return (value !== undefined && value !== null && typeof value === 'string') || value instanceof String;
+  return (!!value && typeof value === 'string') || value instanceof String;
 }
 
 export function assertString(value: unknown): string {
--- a/packages/shared/src/messages/MessageProxy.ts
+++ b/packages/shared/src/messages/MessageProxy.ts
@@ -42,9 +42,7 @@ export interface ISubscribedMessage {
 type UnaryRPC = (...args: unknown[]) => Promise<unknown>;
 
 export function isMessageRequest(content: unknown): content is IMessageRequest {
-  return (
-    content !== undefined && content !== null && typeof content === 'object' && 'id' in content && 'channel' in content
-  );
+  return !!content && typeof content === 'object' && 'id' in content && 'channel' in content;
 }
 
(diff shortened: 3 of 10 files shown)


packages/backend/src/managers/catalogManager.ts
```diff
@@ -66,7 +66,7 @@ export class CatalogManager {
     try {
       this.watchCatalogFile(catalogPath); // do not await, we want to do this async
     } catch (err: unknown) {
-      console.error("unable to watch catalog file, changes to the catalog file won't be reflected to the UI", err);
+      console.error(`unable to watch catalog file, changes to the catalog file won't be reflected to the UI`, err);
```
axel7083, line 69: Not a template ?

packages/backend/src/models/AIConfig.ts
```diff
@@ -32,6 +32,7 @@ export interface AIConfig {
 }
 
 export function isString(value: unknown): value is string {
+  // eslint-disable-next-line no-null/no-null
   return (value !== undefined && value !== null && typeof value === 'string') || value instanceof String;
```
axel7083, line 36:
```suggestion
  return (!!value && typeof value === 'string') || value instanceof String;
```

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -43,6 +43,7 @@ type UnaryRPC = (...args: unknown[]) => Promise<unknown>;
 
 export function isMessageRequest(content: unknown): content is IMessageRequest {
   return (
+    // eslint-disable-next-line no-null/no-null
     content !== undefined && content !== null && typeof content === 'object' && 'id' in content && 'channel' in content
```
axel7083, line 47:
```suggestion
    !!content && typeof content === 'object' && 'id' in content && 'channel' in content
```

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -210,6 +211,7 @@ export class RpcBrowser {
   isSubscribedMessage(content: any): content is ISubscribedMessage {
     return (
       content !== undefined &&
+      // eslint-disable-next-line no-null/no-null
       content !== null &&
```
axel7083, line 215:
```suggestion
      !!content &&
```

axel7083 (approved):
LGTM

lstocchi (approved):
I think we can remove the null checks, but ok for a follow-up PR. LGTM

packages/backend/src/managers/catalogManager.ts
benoitf: else we need to replace won't by will not

packages/backend/src/models/AIConfig.ts
```diff
@@ -32,7 +32,8 @@ export interface AIConfig {
 }
 
 export function isString(value: unknown): value is string {
-  return (value !== undefined && value !== null && typeof value === 'string') || value instanceof String;
+  // eslint-disable-next-line no-null/no-null
```
benoitf, line 35:
```suggestion
```

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -43,7 +43,8 @@ type UnaryRPC = (...args: unknown[]) => Promise<unknown>;
 
 export function isMessageRequest(content: unknown): content is IMessageRequest {
   return (
-    content !== undefined && content !== null && typeof content === 'object' && 'id' in content && 'channel' in content
+    // eslint-disable-next-line no-null/no-null
```
benoitf, line 46:
```suggestion
```

PR #101, opened 2024-01-19
watch catalog file even if non existing, to be able to detect new file

feloy:


axel7083:
I think I fixed it in https://github.com/projectatomic/studio-extension/pull/91 when I rebased

https://github.com/projectatomic/studio-extension/blob/bf1ee55629576b88fc8ec3036bf998ba29225c5d/packages/backend/src/managers/catalogManager.ts#L60-L67

axel7083:
Oh yeah I moved it thinking it was not necessary, but I haven't thought about new file being created!

axel7083 (approved):
LGTM, Thanks !

PR #104, opened 2024-01-22
feat: adding router state persistence

axel7083:
## Description

Fixes https://github.com/projectatomic/studio-extension/issues/99

## Screenshots

https://github.com/projectatomic/studio-extension/assets/42176370/85e2a303-ab30-4770-97a0-aa77042e0881

### DIFF
--- a/packages/shared/src/models/IRouterState.ts
+++ b/packages/shared/src/models/IRouterState.ts
@@ -1,0 +1,21 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


axel7083:
> FYI I don't see when it goes to setState/getState of the webview ?
> 
> I mean, I see you added an adhoc backend/frontend code to manage a state, but the webview already provides a setState/getState in addition to postMessage once you get the acquirePodmanDesktopApi object
> 
> so rather than adding a new layer in the App, you can get the routing state using `getState()` and each time you updated the route, you update it with `setState()`

I tried using the getState setState, it is destroyed when leaving the page, I am not able to get it back, I thought it was only for when the webview is running, maybe it was a bug ?

axel7083:
Alright, after investigation the issue was that the router update the metadata before it allows me to get the last state, so it was updating the state before I was able to get it.

packages/shared/src/models/IRouterState.ts
```diff
@@ -0,0 +1,21 @@
+/**********************************************************************
```
benoitf, line 1: AFAIK it should not be a shared information as it's only on the frontend side ?

axel7083: Yes

benoitf:
also as I reported on previous PR,  might be time to add some tests

PR #107, opened 2024-01-22
Do not download model again

### LINKED ISSUE #18: fix: Do not download models already on disk
Before downloading a models, checking if it is already on disk, avoid downloading it multiple times.

feloy:
Fixes #18

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -22,7 +22,11 @@ vi.mock('../models/AIConfig', () => ({
 test('appUserDirectory should be under home directory', () => {
   vi.spyOn(os, 'homedir').mockReturnValue('/home/user');
   const manager = new ApplicationManager({} as GitManager, {} as RecipeStatusRegistry, {} as ExtensionContext);
-  expect(manager.appUserDirectory).toMatch(/^(\/|\\)home(\/|\\)user/);
+  if (process.platform === 'win32') {
+    expect(manager.appUserDirectory).toMatch(/^\\home\\user/);
+  } else {
+    expect(manager.appUserDirectory).toMatch(/^\/home\/user/);
+  }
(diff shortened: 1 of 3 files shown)


benoitf (approved):
would be nice (as for many others PR) to add more unit tests

packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  vi.spyOn(os, 'homedir').mockReturnValue('/home/user');
+  const manager = new ApplicationManager({} as GitManager, {} as RecipeStatusRegistry, {} as ExtensionContext);
+  expect(manager.appUserDirectory).toMatch(/^\/home\/user/);
+});
+
+test('getLocalModels should return models in local directory', () => {
+  vi.spyOn(os, 'homedir').mockReturnValue('/home/user');
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  vi.spyOn(fs, 'readdirSync').mockImplementation((dir: string): any => {
+    // TODO(feloy): fix any
```
feloy, line 20: The difficulty with readdirSync is that it returns either Dirent[] or string[] depending on the options passed. I'm not sure how to handle this without using any here...

benoitf:
```ts
const readdirSyncMock: MockInstance<[path: fs.PathLike], string[]> = vi.spyOn(fs, 'readdirSync') as unknown as MockInstance<[path: fs.PathLike], string[]>;
readdirSyncMock.mockImplementation((path: fs.PathLike) => {
```

benoitf: (that said on unit test I'm fine to use `any` type)

feloy: Thanks

PR #111, opened 2024-01-22
feat: adding icons to container images managed by the extension

axel7083:
## Description

Fixes https://github.com/projectatomic/studio-extension/issues/109

## Screenshots

https://github.com/projectatomic/studio-extension/assets/42176370/b2f80913-9648-41dd-bfd7-ce85b533e04b


lstocchi (approved):
Maybe we need to fix the size on podman as it is a bit smaller than the icon above and they should have the same width but it is on a different PR.
LGTM

axel7083:
> Maybe we need to fix the size on podman as it is a bit smaller than the icon above and they should have the same width but it is on a different PR. LGTM

There an issue open I think on the podman-desktop side, I do think it is related to me

PR #112, opened 2024-01-23
refactor: move getLocalModels to ModelsManager

feloy:


packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -1,15 +1,13 @@
 import { type MockInstance, describe, expect, test, vi, beforeEach } from 'vitest';
 import { ApplicationManager } from './applicationManager';
 import type { RecipeStatusRegistry } from '../registries/RecipeStatusRegistry';
-import type { ExtensionContext } from '@podman-desktop/api';
 import type { GitManager } from './gitManager';
 import os from 'os';
-import fs, { Stats, type Dirent } from 'fs';
-import path from 'path';
+import fs from 'fs';
```
benoitf, line 6:
```suggestion
import fs from 'node:fs';
```

benoitf: using `node:` prefix is recommended for all the 'internal' packages

it's to avoid for example that you use `https://www.npmjs.com/package/fs`

packages/backend/src/managers/modelsManager.ts
```diff
@@ -0,0 +1,26 @@
+import type { LocalModelInfo } from '@shared/src/models/ILocalModelInfo';
+import fs from 'fs';
+import * as path from 'node:path';
+
+export class ModelsManager {
+  constructor(private appUserDirectory: string) {}
+
+  getLocalModels(): LocalModelInfo[] {
```
benoitf, line 8: just a question.
It seems it's a lazy initialization and without cache, do we need to read the files each time we're called or you're planning to read it eagerly or have a cache

benoitf (approved):

packages/backend/src/managers/modelsManager.ts
axel7083: We could watch the folder for changes

feloy: Good point, I created #116

PR #115, opened 2024-01-23
feat: display error if not matching containers is found

axel7083:
Fixes https://github.com/projectatomic/studio-extension/issues/114

- [x] Writing tests


packages/backend/src/managers/applicationManager.ts
```diff
@@ -132,15 +132,21 @@ export class ApplicationManager {
...
+    if (filteredContainers.length > 0) {
+      // Mark as success.
+      loadingConfiguration.state = 'success';
+      taskUtil.setTask(loadingConfiguration);
+    } else {
+      // Mark as failure.
+      loadingConfiguration.state = 'error';
+      taskUtil.setTask(loadingConfiguration);
+    }
+
```
benoitf, line 149: line 143 and 147looks the same so probably should be after the if/else

axel7083: Since I am throwing an error I can't

benoitf: before there was not the throw line 🤣 

but you could do to avoid duplicating lines

```ts
try {
  if (filteredContainers.length > 0) {
    // Mark as success.
    loadingConfiguration.state = 'success';
  } else {
    // Mark as failure.
    loadingConfiguration.state = 'error';
    throw new Error('No containers available.');
  }
} finally {
  taskUtil.setTask(loadingConfiguration);
}
```

(you can of course discard my comment :-)

axel7083: Yeah sorry I should have said I updated the code ahah.

I don't mind having one line duplicates instead of adding a try finally

feloy (approved):

packages/backend/src/managers/applicationManager.ts
```diff
@@ -126,15 +126,22 @@ export class ApplicationManager {
...
-    taskUtil.setTask(loadingConfiguration);
-
     // Filter the containers based on architecture
     const filteredContainers = aiConfig.application.containers.filter(
       container => container.arch === undefined || container.arch === arch(),
     );
 
+    if (filteredContainers.length > 0) {
+      // Mark as success.
+      loadingConfiguration.state = 'success';
```
lstocchi, line 136: I would create a new task. If there is no container for the user arch/gpu it is not that the config file failed at loading but that it does not support the user system. I would create one to specify it. Like checking if system supports app or something like that.

axel7083: This is a possibility that works, but on the long term I was thinking supporting an optional "error" property on the task, where we could provide more insight on the reason of a failing task.

lstocchi: TBH I don't know which is better but if others do not complain againts it you can go and merge.

axel7083: I think we should probably either way, start using the Taks from the podman-desktop api and just proxy them, but the api is kinda limiting for now, but discussion open

jeffmaury (approved):
LGTM

Issue #116, opened 2024-01-23
feat: Watch the models folder to reload models reactively


feloy:
For now, models are read every time we want to display the list of models (and more).

It would be more efficient to watch the models folder, and update the list of models when changes occur.

PR #118, opened 2024-01-23
Display size and creation age for local models

### LINKED ISSUE #49: Improve the metadata displayed about the list of locally loaded models
In the list of local models, we would like to display:
- Size of the model
- Date of the download
- Ability to get the checksum

feloy:
partially fixes #49 

Because computing the checksum takes time (reading the file completely), #116 should be done before to compute the checksum.


benoitf (approved):

packages/frontend/src/lib/table/model/ModelColumnSize.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
```
benoitf, line 1: need component test ?

packages/frontend/src/utils/dimensions.ts
```diff
@@ -0,0 +1,9 @@
+import moment from 'moment';
```
benoitf, line 1: missing headerss/copyright

packages/frontend/src/lib/table/model/ModelColumnCreation.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
```
benoitf, line 1: need component test ?

PR #119, opened 2024-01-24
watch local models

### LINKED ISSUE #116: feat: Watch the models folder to reload models reactively

feloy:
This PR makes the extension watch the folder containing the models, and updates the UI accordingly.

Fixes #116

### DIFF
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -1,17 +1,53 @@
 import type { LocalModelInfo } from '@shared/src/models/ILocalModelInfo';
 import fs from 'fs';
 import * as path from 'node:path';
+import { type Webview, fs as apiFs } from '@podman-desktop/api';
+import { MSG_NEW_LOCAL_MODELS_STATE } from '@shared/Messages';
+import type { CatalogManager } from './catalogManager';
 
 export class ModelsManager {
-  constructor(private appUserDirectory: string) {}
+  #modelsDir: string;
+  #localModels: Map<string, LocalModelInfo>;
 
-  getLocalModels(): LocalModelInfo[] {
-    const result: LocalModelInfo[] = [];
-    const modelsDir = path.join(this.appUserDirectory, 'models');
-    if (!fs.existsSync(modelsDir)) {
-      return [];
+  constructor(
+    private appUserDirectory: string,
+    private webview: Webview,
+    private catalogManager: CatalogManager,
+  ) {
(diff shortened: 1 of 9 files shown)


packages/backend/src/managers/modelsManager.ts
```diff
@@ -1,17 +1,53 @@
...
+    };
+    const watcher = apiFs.createFileSystemWatcher(this.#modelsDir);
+    watcher.onDidCreate(reloadLocalModels);
+    watcher.onDidDelete(reloadLocalModels);
+    watcher.onDidChange(reloadLocalModels);
+    // Initialize the local models manually
+    await reloadLocalModels();
+  }
+
+  getModelsInfo() {
```
axel7083, line 38: Do we do not display models that are not in the catalog ?

packages/shared/Messages.ts
```diff
@@ -2,3 +2,4 @@ export const MSG_PLAYGROUNDS_STATE_UPDATE = 'playgrounds-state-update';
 export const MSG_NEW_PLAYGROUND_QUERIES_STATE = 'new-playground-queries-state';
 export const MSG_NEW_CATALOG_STATE = 'new-catalog-state';
 export const MSG_NEW_RECIPE_STATE = 'new-recipe-state';
+export const MSG_NEW_LOCAL_MODELS_STATE = 'new-local-models-state';
```
axel7083, line 5: why using `new`, it is more likely to be an `updated` since the watcher will detect, deletion, update and new

axel7083 (approved):
LGTM! Great work

packages/backend/src/managers/modelsManager.ts
feloy: Right, we are displaying in the Models list only the models on the disk AND on the catalog, as the catalog is the source for many information (model name, license, etc).

packages/shared/Messages.ts
feloy: New for New state. Every time the state changes, all the state is sent, and the UI rebuilds itself from scratch with this new state. But the updates from the previous state is not taken into consideration. Hence the `New`.

packages/backend/src/managers/modelsManager.ts
axel7083: I would rather still list the models, and show some "?" or information/warning icon the missing field rather than ignoring them

feloy: @slemeur what do you think?

Issue #50, opened 2024-01-16
Provide ability to manage the local models


slemeur:
Once a model is loaded locally, as a user I'd like the following options:
- Being able to open the folder where the model is loaded
- Check if I have the latest version of the model
- Delete the model

feloy:
Needs #116

feloy:
> * Check if I have the latest version of the model

Are models versioned for now? How? @slemeur @jeffmaury

nichjones1:
@feloy this was from the prior sprint, is it still postponed?

feloy:
This question is still open: 

> Are models versioned for now? How? @slemeur @jeffmaury

slemeur:
As most of the issues have been handled, closing in favor of https://github.com/projectatomic/ai-studio/issues/466

PR #134, opened 2024-01-25
feat: delete model

feloy:
fixes partially #50

### DIFF
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -5,6 +5,7 @@ import { type Webview, fs as apiFs } from '@podman-desktop/api';
 import { MSG_NEW_LOCAL_MODELS_STATE } from '@shared/Messages';
 import type { CatalogManager } from './catalogManager';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
+import * as podmanDesktopApi from '@podman-desktop/api';
 
 export class ModelsManager {
   #modelsDir: string;
@@ -116,7 +117,7 @@ export class ModelsManager {
       await fs.promises.rm(modelDir, { recursive: true });
       this.#localModels.delete(modelId);
     } catch (err: unknown) {
-      console.error('unable to delete model', modelId, err);
+      await podmanDesktopApi.window.showErrorMessage(`Error deleting model ${modelId}. ${String(err)}`);
     } finally {
       this.#deleted.delete(modelId);
       await this.sendModelsInfo();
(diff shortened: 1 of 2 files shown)


packages/backend/src/managers/modelsManager.ts
```diff
@@ -85,7 +100,26 @@ export class ModelsManager {
...
   getLocalModels(): LocalModelInfo[] {
     return Array.from(this.#localModels.values());
   }
+
+  async deleteLocalModel(modelId: string): Promise<void> {
+    const modelDir = this.getLocalModelFolder(modelId);
+    this.#deleted.add(modelId);
+    await this.sendModelsInfo();
+    try {
+      await fs.promises.rm(modelDir, { recursive: true });
```
axel7083, line 116: The recursive option for deleting always scares me, something we could do, is listing all the files inside the model directory, ask user for confirmation something like

"The following files will be deleted
- model-a.gguff
"

Then delete them by name instead of deleting the folder ? (I am maybe just overreacting)

packages/frontend/src/lib/table/model/ModelColumnActions.svelte
```diff
@@ -0,0 +1,19 @@
...
+
+function deleteModel() {
+  console.log('delete model', object.id);
+  studioClient.deleteLocalModel(object.id);
+}
+</script>
+
+<ListItemButtonIcon
+  icon={faTrash}
+  onClick={() => deleteModel()}
```
axel7083, line 16: Yeah we need confirmation modal

lstocchi (commented):
I face this error when trying to delete the model
image

feloy:
> I face this error when trying to delete the model image

On Windows, right?

lstocchi:
> On Windows, right? I'll have a look

Yes

lstocchi:
@feloy maybe i got what happened. I had my playground running and i tried to delete the model 🤔 I'll try again, but we should inform the user in this case

feloy:
> @feloy maybe i got what happened. I had my playground running and i tried to delete the model 🤔 I'll try again, but we should inform the user in this case

Yes, I would like to display the error message to the user in this case, but for now, AI Studio doesn't have a clear way to display errors coming from the backend (as the call is async, we cannot get the error from a call to `deleteModel`).

@jeffmaury what do you think? Do we want to implement a generic error handler (kind of notifications), something else?

lstocchi:
> > @feloy maybe i got what happened. I had my playground running and i tried to delete the model 🤔 I'll try again, but we should inform the user in this case
> 
> Yes, I would like to display the error message to the user in this case, but for now, AI Studio doesn't have a clear way to display errors coming from the backend (as the call is async, we cannot get the error as the result from a call to a `deleteModel`).
> 
> @jeffmaury what do you think? Do we want to implement a generic error handler (kind of notifications), something else?

I saw Axel just pushed https://github.com/projectatomic/studio-extension/pull/152 . Maybe adding a task to the taksmanager and an error icon next to the delete icon that opens up the task manager?

lstocchi (approved):
Tried again, now the error is displayed on a dialog. Thanks!

PR #140, opened 2024-01-25
fix: use playground image based on llama-cpp-python

### LINKED ISSUE #139: Replace localAI by llama-cpp-python
Image is available at quay.io/bootsy/playground:v0

jeffmaury:
Fixes #139

### DIFF
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -188,6 +188,7 @@ export class PlayGroundManager {
       model: modelInfo.file,
       prompt: prompt,
       temperature: 0.7,
+      max_tokens: 1024,
     });
 
     const post_options: http.RequestOptions = {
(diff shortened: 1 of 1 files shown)


feloy:
I get an unauthorized access when I try to pull the image quay.io/bootsy/playground:v0

feloy:
It seems that we get only the beginning of the response from the model

model-begin

jeffmaury:
Need to verify with a curl request

feloy:
The reply does not contain anything indicating there will be a suite.

Is it an internal problem in the container, or are we using the good request?

reply

feloy:
Yes, that works for me now. 
The container gets more time to start now, I have created a related issue: https://github.com/projectatomic/studio-extension/issues/147

feloy (approved):

lstocchi (approved):
Super slow on windows and the answer is not really the one i expected but it works 😅 

image

Issue #131, opened 2024-01-25
Create the containers inside the pod


lstocchi:
https://github.com/projectatomic/studio-extension/pull/100 adds a first implementation of the pod creation but what it actually does is:
create pod -> create containers -> add them to the pod
This way you will see the containers both in the pod and outside. We should update the createContainers function so that they are created inside the pod directly

PR #144, opened 2024-01-25
feat: start pod with sample and model service

lstocchi:
This PR starts the pod and checks the model service so when it's up it can restart the sample app container


jeffmaury (commented):
LGTM. Tested on Win11 but I still have the original containers (out of the pod) being there after the pod is started. Now testing on MacOS/QEMU

lstocchi:
@jeffmaury that is a different issue https://github.com/projectatomic/studio-extension/issues/131. I think we need to update podman desktop api

feloy:
Works for me on mac/qemu

feloy (approved):

jeffmaury:
> @jeffmaury that is a different issue #131. I think we need to update podman desktop api

Can't we delete them once the pod is fully started ? But that maybe a follow up PR

jeffmaury:
Tested on MacOS/QEMU as well.

lstocchi:
> Can't we delete them once the pod is fully started ? But that maybe a follow up PR

Yes if we want to add a temporary solution until we have a cleaner one we can delete them easily after creating the pod

lstocchi:
@jeffmaury if you approve i merge it

jeffmaury (approved):

PR #152, opened 2024-01-26
feat: creating task in task manager

axel7083:
Fixes https://github.com/projectatomic/studio-extension/issues/151

### DIFF
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -74,17 +74,10 @@ export class StudioApiImpl implements StudioAPI {
     // Do not wait for the pull application, run it separately
     void podmanDesktopApi.window.withProgress<void>(
       { location: podmanDesktopApi.ProgressLocation.TASK_WIDGET, title: `Pulling ${recipe.name}.` },
-      async progress => {
-        try {
-          await this.applicationManager.pullApplication(recipe, model);
-          // mark the task as completed
-          progress.report({ increment: -1 });
-        } catch (error) {
-          console.warn(error);
-          progress.report({ message: `Error: ${String(error)}` });
-        }
-      },
-    );
+      () => this.applicationManager.pullApplication(recipe, model),
+    ).catch(() => {
+      this.recipeStatusRegistry.setStatus(recipeId, {recipeId: recipeId, state: 'error', tasks: []});
+    });
   }
 
   async getLocalModels(): Promise<ModelInfo[]> {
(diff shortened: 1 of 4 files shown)


feloy:
Where are the progress/error expected to be visible? I cannot see anything on the UI or tray on mac

axel7083:
I had an error where the promise was not properly waited, I fixed it @feloy thanks for testing

Here is a demo

https://github.com/projectatomic/studio-extension/assets/42176370/081e4126-5fa5-4b92-b182-7b8c781cb729

feloy:
Can you see the error displayed, in case of error?

I didn't tested on this PR but on my PR about deleting models, using similar code, and I was not able to see the error message appearing anywhere after calling `progress.report({ message: `Error: ${String(error)}` });`

axel7083:
> Can you see the error displayed, in case of error?
> 
> I didn't tested on this PR but on my PR about deleting models, using similar code, and I was not able to see the error message appearing anywhere after calling `progress.report({ message: `Error: ${String(error)}` });`

Me neither, that's weird

axel7083:
Okey @feloy I investigate, the bug is coming from the podman-core. I will open an issue, I already have a fix

packages/backend/src/studio-api-impl.ts
```diff
@@ -73,9 +72,14 @@ export class StudioApiImpl implements StudioAPI {
     const model = await this.getModelById(modelId);
 
     // Do not wait for the pull application, run it separately
-    this.applicationManager.pullApplication(recipe, model).catch((error: unknown) => console.warn(error));
-
-    return Promise.resolve(undefined);
+    void podmanDesktopApi.window
```
lstocchi, line 75: can we remove the `void` here?

lstocchi (approved):
Tested and works fine. 🚀  I'll just remove that void before the function call.

packages/backend/src/studio-api-impl.ts
axel7083: No the reason is avoid linter to complain that we have "dangling promises"

lstocchi: for me it runs fine without it 🤔 
Do you mean the `yarn lint:check`?

benoitf: I agree here with @lstocchi we shouldn't use that void to fake the compiler

also you're doing a catch so linter should not report any problem

you can also remove the `<void>` after `.withProgress<void>` as default is void

benoitf: the rule is that you should use:

```ts
await promise // valid
promise.....catch(error...) //valid
```

and not
```ts
promise // invalid (and do not add void before to disable the linter as then at runtime you may have unhandled exception
void promise // no
```

PR #157, opened 2024-01-26
fix: add release workflow

### LINKED ISSUE #156: Add release workflow

jeffmaury:
Fixes #156


benoitf:
IMHO you need to make a PR to change https://github.com/projectatomic/studio-extension/blob/main/.github/workflows/build-next.yaml#L61 first

so each build after CI won't use `latest` rolling tag but another one `ci`, `dev`, whatever option you feel is better

benoitf (approved):

lstocchi (approved):

PR #160, opened 2024-01-26
fix: delete containers outside the pod

### LINKED ISSUE #159: Delete containers created outside the pod
This issue is just for tracking the temp workaround we will add until we have https://github.com/projectatomic/studio-extension/issues/131

lstocchi:
this patch allows to delete the container that is outside the pod after the creation + copy into the pod is complete.

it resolves #159


jeffmaury (approved):
LGTM

PR #164, opened 2024-01-26
chore: adding summarizer app

lstocchi:
This PR updates the catalog to show the summarizer sample and removes the two models that were in but not actually working


feloy (approved):

jeffmaury (changes requested):
LGTM. Tested on Win11 Pro

Can we update the readme so that is reflects the README from summarizer instead of the one on the root repo

lstocchi:
> Can we update the readme so that is reflects the README from summarizer instead of the one on the root repo

I just copied the part that makes sense.
I omitted the build/run image, ok?

image

jeffmaury (approved):

PR #167, opened 2024-01-29
Open Model Folder

feloy:
Fixes partially #50


packages/frontend/src/lib/table/model/ModelColumnActions.svelte
```diff
@@ -17,8 +18,20 @@ async function goDeleteModel() {
   await studioClient.deleteLocalModel(object.id);
   deleteConfirmVisible= false;
 }
+
+function openModelFolder() {
+  console.log(object.file);
```
benoitf, line 23:
```suggestion

```

packages/frontend/src/lib/table/model/ModelColumnActions.svelte
```diff
@@ -17,8 +18,20 @@ async function goDeleteModel() {
...
+  if (object && object.file) {
+    studioClient.openURL('file://'+object.file.path);
```
benoitf, line 24: I would use uris everywhere

build the URI using
```
Uri.file(object.file.path)
```

and then not have in studioClient
```
  async openURL(url: string): Promise<boolean> {
    return await podmanDesktopApi.env.openExternal(podmanDesktopApi.Uri.parse(url));
  }
```
but just accept URI (to avoid double parse/unexpected behavior)

benoitf: I did not pay enough attention to backend/frontend usage so you may disregard my comment as basically frontend has no access to Podman Desktop API and the URI class.

so either you revert the commit or you have `openFile(filePath: string)` method on studio-client that will do `openExternal(URI.file(filePath))`

sorry for the noise

lstocchi (commented):
When opening the model folder i see this dialog but the title and text are wrong. I'm not opening any external website

image

jeffmaury (approved):
LGTM

axel7083:
I had https://github.com/projectatomic/ai-studio/pull/122 open, I will put it as draft waiting for this one to be merged and reuse your code @feloy

feloy:
> When opening the model folder i see this dialog but the title and text are wrong. I'm not opening any external website
> 
> image

I created an issue on Podman Desktop: https://github.com/containers/podman-desktop/issues/5732

lstocchi (commented):
I would fix the issue on desktop before releasing tbh but maybe @jeffmaury wants to have it for the release.

BTW looking at it again ... do we really need to show that dialog confirmation? We are just opening a local folder, nothing too serious. Can we just open the folder without asking to confirm?

jeffmaury:
> I would fix the issue on desktop before releasing tbh but maybe @jeffmaury wants to have it for the release.
> 
> BTW looking at it again ... do we really need to show that dialog confirmation? We are just opening a local folder, nothing too serious. Can we just open the folder without asking to confirm?

That maybe the cause of security issue, @benoitf WDYT ?

benoitf:
I created the commit and the branch but forgot to open the PR

here it is https://github.com/containers/podman-desktop/pull/5743

PR #168, opened 2024-01-29
fix: listen to correct host port when checking model service 

lstocchi:
This PR fixes an incorrect behavior when there were multiple sample applications running at the same time. 
Before the port used to check if the model service was running was taken by looking at the container port. 
This patch fixes it and take the correct host port


jeffmaury (approved):
LGTM

PR #169, opened 2024-01-29
Telemetry

### LINKED ISSUE #35: Add Telemetry usage events
As we would like to better understand how the extension is getting used, we should be adding Telemetry usage events.
That should also be documented in the repository, in a similar fashion than: https://github.com/containers/podman-desktop/blob/main/USAGE_DATA.md

Initial list of usage data to look at:
- The user is opening the studio extension, tabs opened
- The recipe which is opened
- Pull of the sample application
- Start/Stop/Usage of the playground environment 
- Events related to actions (download, management) of the models

## TODO
- [ ] send duration for playground.ask response

feloy:
Fixes partially #35

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -604,15 +604,14 @@ describe('buildImages', () => {
     {} as unknown as ModelsManager,
     telemetryLogger,
   );
-  test('setTaskState should be called with error and telemetry seent if context does not exist', async () => {
+  test('setTaskState should be called with error if context does not exist', async () => {
     vi.spyOn(fs, 'existsSync').mockReturnValue(false);
     mocks.listImagesMock.mockRejectedValue([]);
     await expect(manager.buildImages(containers, 'config', taskUtils)).rejects.toThrow(
       'Context configured does not exist.',
     );
-    expect(mocks.logErrorMock).toHaveBeenCalled();
   });
-  test('setTaskState should be called with error and telemetry sent if buildImage executon fails', async () => {
+  test('setTaskState should be called with error if buildImage executon fails', async () => {
     vi.spyOn(fs, 'existsSync').mockReturnValue(true);
     mocks.builImageMock.mockRejectedValue('error');
     mocks.listImagesMock.mockRejectedValue([]);
@@ -620,7 +619,6 @@ describe('buildImages', () => {
       'Something went wrong while building the image: error',
     );
     expect(setTaskStateMock).toBeCalledWith('container1', 'error');
-    expect(mocks.logErrorMock).toHaveBeenCalled();
   });
   test('setTaskState should be called with error if unable to find the image after built', async () => {
     vi.spyOn(fs, 'existsSync').mockReturnValue(true);
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -31,7 +31,7 @@ import type { AIConfig, ContainerConfig } from '../models/AIConfig';
 import * as portsUtils from '../utils/ports';
 import { goarch } from '../utils/arch';
 import * as utils from '../utils/utils';
-import type { Webview } from '@podman-desktop/api';
+import type { Webview, TelemetryLogger } from '@podman-desktop/api';
 import type { CatalogManager } from './catalogManager';
 
 const mocks = vi.hoisted(() => {
@@ -47,6 +47,8 @@ const mocks = vi.hoisted(() => {
     startPod: vi.fn(),
     deleteContainerMock: vi.fn(),
     inspectContainerMock: vi.fn(),
+    logUsageMock: vi.fn(),
+    logErrorMock: vi.fn(),
   };
 });
 vi.mock('../models/AIConfig', () => ({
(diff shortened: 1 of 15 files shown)


lstocchi (commented):
Not going to approve as it's better that Jeff or Florent or Stevan look at it.
LGTM, i'm just unsure about the telemetry event names. I remember that we needed to use the function name 🤔 So instead of using `recipe.pull` we should use the name of the function we are into, but i'll wait for the comment of some expert

packages/shared/src/StudioAPI.ts
```diff
@@ -25,4 +26,9 @@ export abstract class StudioAPI {
   abstract askPlayground(modelId: string, prompt: string): Promise<number>;
   abstract getPlaygroundQueriesState(): Promise<QueryState[]>;
   abstract getPlaygroundsState(): Promise<PlaygroundState[]>;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  abstract telemetryLogUsage(eventName: string, data?: Record<string, any | TelemetryTrustedValue>): Promise<void>;
```
axel7083, line 31: Can't we use `unknown` ?

packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -823,6 +893,10 @@ describe('createAndAddContainersToPod', () => {
     {} as unknown as GitManager,
     {} as unknown as RecipeStatusRegistry,
     {} as unknown as ModelsManager,
+    {
+      logUsage: mocks.logUsageMock,
+      logError: mocks.logErrorMock,
+    } as unknown as TelemetryLogger,
```
jeffmaury, line 899: Can't we have a shared variable to represent the telemetry logger ?

packages/backend/src/managers/modelsManager.ts
```diff
@@ -173,6 +175,7 @@ export class ModelsManager {
       });
 
       try {
+        this.telemetry.logUsage('model.download', { 'model.id': model.id });
```
jeffmaury, line 178: Would be good to have a single event: here we will have 2 events in case of error

packages/backend/src/managers/modelsManager.ts
```diff
@@ -184,6 +187,7 @@ export class ModelsManager {
             'model-pulling': model.id,
           },
         });
+        this.telemetry.logError('model.download', { message: 'error downloading model', error: e });
```
jeffmaury, line 190: model.id is missing.

packages/backend/src/managers/modelsManager.ts
jeffmaury, line 178: Also would be good to have execution time for the action (download)

packages/backend/src/managers/playground.ts
```diff
@@ -107,6 +109,7 @@ export class PlayGroundManager {
     const connection = findFirstProvider();
     if (!connection) {
       this.setPlaygroundStatus(modelId, 'error');
+      this.telemetry.logError('playground.start', { message: 'unable to find an engine to start playground' });
```
jeffmaury, line 112: add model.id

packages/backend/src/studio-api-impl.ts
```diff
@@ -71,6 +72,8 @@ export class StudioApiImpl implements StudioAPI {
     const recipe = this.catalogManager.getRecipes().find(recipe => recipe.id === recipeId);
     if (!recipe) throw new Error('Not found');
 
+    await this.telemetryLogUsage('recipe.pull', { 'recipe.id': recipe.id, 'recipe.name': recipe.name });
```
jeffmaury, line 75: Not sure the telemetry event should be located here which is a gateway or we need to better handle errors

jeffmaury (commented):
Also would be good to have timing info for the pull operation

packages/backend/src/managers/applicationManager.ts
```diff
@@ -157,6 +161,7 @@ export class ApplicationManager {
     await timeout(5000);
     await this.restartContainerWhenModelServiceIsUp(engineId, modelServiceEndpoint, container).catch(
       (error: unknown) => {
+        this.telemetry.logError('recipe.pull', { message: 'error monitoring endpoint', error: error });
```
jeffmaury, line 164: Event is not uniform here recipe id, recipe name and duration are not sent. I'm wondering if we should have a single point for sending this event (pullApplication) and handle exceptions there

packages/backend/src/managers/modelsManager.ts
```diff
@@ -184,6 +196,11 @@ export class ModelsManager {
             'model-pulling': model.id,
           },
         });
+        this.telemetry.logError('model.download', {
+          'model.id': model.id,
+          message: 'error downloading model',
+          error: e,
```
jeffmaury, line 202: Execution time is missing

packages/backend/src/managers/playground.ts
```diff
@@ -237,6 +247,7 @@ export class PlayGroundManager {
       modelId,
     });
 
+    this.telemetry.logUsage('playground.start', { 'model.id': modelId });
```
jeffmaury, line 250: Would be good to have the time to start playground as well

packages/backend/src/managers/playground.ts
```diff
@@ -294,7 +312,7 @@ export class PlayGroundManager {
         this.sendQueriesState();
       }
     })().catch((err: unknown) => console.warn(`Error while reading streamed response for model ${modelInfo.id}`, err));
-
+    this.telemetry.logUsage('playground.ask', { 'model.id': modelInfo.id });
```
jeffmaury, line 315: Maybe we should also record the execution time for a request but that could be a follow up PR

jeffmaury (commented):
Execution time is missing for stopPlayground

PR #187, opened 2024-01-30
feat: playground error propagation

axel7083:
# Description

Propagate the http request error to the frontend so the user can have a "sense" of what is happening

https://github.com/projectatomic/ai-studio/assets/42176370/ab3251b1-92bd-422b-99b7-bea50e9a3e2e

> The `inProgress` of the demo has been removed, it is not part of the PR as it was a test.
# Related PR

Fixes https://github.com/projectatomic/ai-studio/issues/186

# Testing

- [x] Write unit tests

### DIFF
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -224,7 +224,7 @@ export class PlayGroundManager {
     post_req.write(post_data);
     post_req.end();
     post_req.on('error', error => {
-      console.error('connection on error.');
+      console.error('connection on error.', error);
       const q = this.queries.get(query.id);
       q.error = `Something went wrong while trying to request model.${String(error)}`;
       void this.sendQueriesState();
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -227,7 +227,7 @@ export class PlayGroundManager {
       console.error('connection on error.', error);
       const q = this.queries.get(query.id);
       q.error = `Something went wrong while trying to request model.${String(error)}`;
-      void this.sendQueriesState();
+      this.sendQueriesState();
     });
 
     this.queries.set(query.id, query);
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -178,11 +178,11 @@ export class PlayGroundManager {
       throw new Error('model is not running');
     }
 
-    const query = {
+    const query: QueryState = {
       id: this.getNextQueryId(),
       modelId: modelInfo.id,
       prompt: prompt,
-    } as QueryState;
+    };
 
     const post_data = JSON.stringify({
       model: modelInfo.file,
@@ -213,6 +213,7 @@ export class PlayGroundManager {
           if (!q) {
             throw new Error('query not found in state');
           }
+          q.error = undefined;
           q.response = result as ModelResponse;
           this.queries.set(query.id, q);
           this.sendQueriesState();
@@ -222,6 +223,12 @@ export class PlayGroundManager {
     // post the data
     post_req.write(post_data);
     post_req.end();
+    post_req.on('error', (error) => {
+      console.error('connection on error.');
+      const q = this.queries.get(query.id);
+      q.error = `Something went wrong while trying to request model.${String(error)}`;
+      void this.sendQueriesState();
+    });
 
     this.queries.set(query.id, query);
     this.sendQueriesState();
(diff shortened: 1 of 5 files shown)


axel7083:
cc @slemeur

feloy (approved):

packages/backend/src/managers/playground.ts
```diff
@@ -213,17 +213,22 @@ export class PlayGroundManager {
           if (!q) {
             throw new Error('query not found in state');
           }
+          q.error = undefined;
           q.response = result as ModelResponse;
           this.queries.set(query.id, q);
-          this.sendQueriesState().catch((err: unknown) => {
-            console.error('playground: unable to send the response to the frontend', err);
-          });
+          void this.sendQueriesState();
```
lstocchi, line 219: Remove the void and add the catch

packages/backend/src/managers/playground.ts
```diff
@@ -213,17 +213,22 @@ export class PlayGroundManager {
...
       });
     });
     // post the data
     post_req.write(post_data);
     post_req.end();
+    post_req.on('error', error => {
+      console.error('connection on error.');
+      const q = this.queries.get(query.id);
+      q.error = `Something went wrong while trying to request model.${String(error)}`;
+      void this.sendQueriesState();
```
lstocchi, line 230: Remove the void and add the catch

packages/backend/src/managers/playground.ts
```diff
@@ -213,17 +213,22 @@ export class PlayGroundManager {
...
-          });
+          void this.sendQueriesState();
         }
       });
     });
     // post the data
     post_req.write(post_data);
     post_req.end();
+    post_req.on('error', error => {
+      console.error('connection on error.');
```
lstocchi, line 227: Maybe you can add the error here

lstocchi (changes requested):

packages/backend/src/managers/playground.ts
axel7083: May I ask why ? the void is to avoid waiting on it, so we do not block the thread without being forced to add a catch clause

What do you want to do in the catch block ? The method `sendQueriesState` is a calling the PostMessage api method, it _cannot_ fails as the api does not say error may occurs, so I don't see the risk of having `unhandledRejection` raised. 
The console.error in the catch block would in my opinion just add lines that are not useful nor relevant. We might want the advice of @benoitf on the matter

I think, that the sendQueriesState should not be `async`, and the implemention inside it should be responsible of `catching` unlikely issue

lstocchi: adding the void makes the code less readable and potentially confusing for me and it does not add any value as it is just there to fake the compiler. You're calling the api so you cannot know if it returns or not an error. It may change.
IMO the goal is not to write less lines but to make them more understandable for everyone so we're less prone to errors when we work on it.

benoitf: to me, the risk to have floating promise is much more higher by silently ignoring possible errors
so I would advise to never add `void` keyword

Either the method handle by itself the asynchronous execution (we did that in the past for telemetry so it's no longer an async method) so you can call `this.sendQueriesState()` and linter will be happy.

Or if it's a promise you need to wait for the execution or catch the error.

Probably we need to tune the no-floating-promises to ensure it's not added by mistake.

```
"@typescript-eslint/no-floating-promises": [ "error", {"ignoreVoid": false } ]
```

axel7083: Related issue https://github.com/projectatomic/ai-studio/pull/197 I will rebase this PR when merged

packages/backend/src/managers/playground.ts
axel7083: Yes, I forgot apparently, thanks for noticing !

axel7083:
Rebased after https://github.com/projectatomic/ai-studio/pull/197 merged

packages/backend/src/managers/playground.ts
```diff
@@ -222,6 +223,12 @@ export class PlayGroundManager {
     // post the data
     post_req.write(post_data);
     post_req.end();
+    post_req.on('error', error => {
+      console.error('connection on error.', error);
+      const q = this.queries.get(query.id);
+      q.error = `Something went wrong while trying to request model.${String(error)}`;
+      void this.sendQueriesState();
```
lstocchi, line 230:
```suggestion
      this.sendQueriesState();
```

lstocchi (commented):
Code looks good. Giving a quick run

packages/backend/src/managers/playground.ts
axel7083: Thanks !! Almost forgot it!

lstocchi (approved):
LGTM. Thanks @axel7083 !! 🚀

PR #189, opened 2024-01-30
fix: button must have a spinner when progress true

axel7083:
# Description 

Fixe Button svelte component to have proper spinner when progress true

# Screenshots

https://github.com/projectatomic/ai-studio/assets/42176370/b3d4adca-bea6-4fa7-a192-70e370b0a4d4

# Related issues

Fixes https://github.com/projectatomic/ai-studio/issues/188

# Tests

### DIFF
--- a/packages/frontend/src/lib/button/Button.spec.ts
+++ b/packages/frontend/src/lib/button/Button.spec.ts
@@ -1,0 +1,20 @@
+import '@testing-library/jest-dom/vitest';
+import { test, expect } from 'vitest';
+import { render, screen } from '@testing-library/svelte';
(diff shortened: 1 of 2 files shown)


axel7083:
cc @slemeur

benoitf:
I think that when modifying code duplicated from the core of podman desktop you should first discuss in the upstream before applying what I call 'downstream/ai-studio' library because your manual change will be easily dropped/forgotten when for example ai-studio' would use the podman desktop button component

benoitf:
here for example you could use an icon for the 'send request' button as well
When in progress, icon will be replaced by the spinner

jeffmaury (approved):
LGTM

PR #197, opened 2024-01-30
fix: sendQueriesState not async

axel7083:
Fixes https://github.com/projectatomic/ai-studio/issues/195

### DIFF
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -239,7 +239,7 @@ export class PlayGroundManager {
     return Array.from(this.playgrounds.values());
   }
 
-  sendQueriesState() {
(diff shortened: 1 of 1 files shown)


packages/backend/src/managers/playground.ts
```diff
@@ -241,10 +239,12 @@ export class PlayGroundManager {
     return Array.from(this.playgrounds.values());
   }
 
-  async sendQueriesState() {
-    await this.webview.postMessage({
+  sendQueriesState() {
```
benoitf, line 242: that said, I think methods should still have return type

probably we're missing another linting rule

axel7083: Resolved

lstocchi (approved):
LGTM

PR #200, opened 2024-01-30
adopt playground containers

### LINKED ISSUE #57: Detect running playground containers based on label
It would be helpful to detect that playground containers are already running when we (re-)start Podman Desktop+AI Studio, so AI Studio can adopt them.

feloy:
This PR adds a new class `PodmanConnection`, whose instance listens for new/updated 'podman' container providers, so other managers can get informed of new containers and adopt some of them depending on their labels.

It is used as part of this PR for adopting playground containers.

It is expected to be used to detect and adopt recipe pods in a future PR.

Fixes #57

### DIFF
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -16,14 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import {
-  provider,
-  containerEngine,
-  type Webview,
-  type ProviderContainerConnection,
-  type ImageInfo,
-  type RegisterContainerConnectionEvent,
-} from '@podman-desktop/api';
+import { containerEngine, type Webview, type ImageInfo } from '@podman-desktop/api';
 import type { LocalModelInfo } from '@shared/src/models/ILocalModelInfo';
 import type { ModelResponse } from '@shared/src/models/IModelResponse';
 
@@ -34,20 +27,13 @@ import type { QueryState } from '@shared/src/models/IPlaygroundQueryState';
 import { MSG_NEW_PLAYGROUND_QUERIES_STATE, MSG_PLAYGROUNDS_STATE_UPDATE } from '@shared/Messages';
 import type { PlaygroundState, PlaygroundStatus } from '@shared/src/models/IPlaygroundState';
 import type { ContainerRegistry } from '../registries/ContainerRegistry';
+import type { PodmanConnection } from './podmanConnection';
 
 const LABEL_MODEL_ID = 'ai-studio-model-id';
 const LABEL_MODEL_PORT = 'ai-studio-model-port';
 
 // TODO: this should not be hardcoded
 const PLAYGROUND_IMAGE = 'quay.io/bootsy/playground:v0';
-
-function findFirstProvider(): ProviderContainerConnection | undefined {
-  const engines = provider
-    .getContainerConnections()
-    .filter(connection => connection.connection.type === 'podman')
-    .filter(connection => connection.connection.status() === 'started');
-  return engines.length > 0 ? engines[0] : undefined;
-}
 
 export class PlayGroundManager {
   private queryIdCounter = 0;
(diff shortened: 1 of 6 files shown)


packages/backend/src/managers/playground.ts
```diff
@@ -56,6 +59,33 @@ export class PlayGroundManager {
     this.queries = new Map<number, QueryState>();
   }
 
+  async adoptRunningPlaygrounds() {
+    const containers = await containerEngine.listContainers();
```
feloy, line 63: @benoitf this call returns an empty list when the extension is loaded at startup (and the correct list of containers when the extension is reloaded, after podman desktop has finished initialization).

What would be the right way to wait for the container engine to be up and running?

packages/backend/src/managers/playground.ts
```diff
@@ -139,7 +169,8 @@ export class PlayGroundManager {
         },
       },
       Labels: {
-        'ia-studio-model': modelId,
+        [LABEL_MODEL_ID]: modelId,
+        [LABEL_MODEL_PORT]: '' + freePort,
```
axel7083, line 173:
```suggestion
        [LABEL_MODEL_PORT]: `${freePort}`,
```

packages/backend/src/managers/playground.ts
feloy: Another solution would be to use `containerEngine.onEvent`, but it seems that we don't get any event when the containerEngine is started. @benoitf WDYT?

jeffmaury: We need to materialize the connection from AI studio to podman machine at a central point I'm pretty sure there is some code that try to get this connection

feloy: I'm finally using `provider.onDidRegisterContainerConnection` to get informed.

packages/backend/src/managers/playground.ts
```diff
@@ -60,6 +64,43 @@ export class PlayGroundManager {
     this.queries = new Map<number, QueryState>();
   }
 
+  async adoptRunningPlaygrounds() {
+    provider.onDidRegisterContainerConnection(async (e: RegisterContainerConnectionEvent) => {
```
lstocchi, line 68: Isn't the register method only called when you create a new connection? (e.g when you create a new podman machine?) Could we use the `onDidUpdateContainerConnection` so it also listens to machines that were stopped and now get started?

jeffmaury: I think what we need is a ConnectionProvider within AI Studio. As startup, it should get the first available Podman connection and then listen for connection changes. Every other piece of AI Studio should use this connection provider to get a Podman connection.

feloy: Please look at my latest commit, it should do what you describe

packages/backend/src/managers/podmanConnection.ts
```diff
@@ -0,0 +1,73 @@
...
+export class PodmanConnection {
+  #firstFound = false;
+  #connection: ContainerProviderConnection | undefined = undefined;
+
+  #toExecuteAtStartup: startupHandle[] = [];
+
+  init(): void {
+    // In case the extension has not yet registered, we listen for new registrations
+    // and retain the first started podman provider
+    const disposable = provider.onDidRegisterContainerConnection((e: RegisterContainerConnectionEvent) => {
```
jeffmaury, line 32: Why are we listening on connection registration andnot connection start/stop events ?

feloy: For this feature, I only want to list containers started before Podman Desktop is started, and I want to be informed when the podman extension is registered (or the `listContainers` would return an empty list if we try to retrieve the list before). 

(note that there is a `registries/ContainerRegistry` which is watching at events on a specific container, which we may merge with this class at some point)

lstocchi: I didn't test it so maybe i'm wrong but looking at the code it looks to me that it initializes the connection when the studio is activated if there is atleast one podman machine that is started and then it starts listening to if a machine gets registered.
So if i start my desktop having my podman machine stopped, the connection is not set as there is no podman machine running. If i start my podman machine in a second moment it does not trigger its initialization bc it only listen to machine registration. 
So if i start the playground or any other feature that requires to retrieve the active connection it fails, no? But i have my podman machine running fine, i just started it after my desktop was activated

feloy: Yes, you are probably right, I'll look at supporting this scenario too. Thanks

feloy: Finally, I prefer for now to remove this exposure and use of `connection` from the `PodmcnConnection` class, as it is not part of what is needed to fix the targeted issue, which is only related to getting a connection to the provider **at startup**.

I'll work at this refactoring as part of another PR, if necessary.

lstocchi: I still think that this PR is not complete. You're just adopting the containers if the podman machine is running when you activate desktop or if a new machine is created. It does not cover the other use cases.

feloy: I have made a few tests on Windows, and I cannot find another scenario where we could adopt a container other than the one already covered: the playground container being created with the AUTOREMOVE flag (equivalent to --rm), when the podman machine is stopped, the container is stopped and deleted, and when the machine is restarted, the container is not present anymore. So, there is no way we can find a playground container when starting a podman machine.

Do you have other scenarios in mind?

lstocchi: I wanted to test it but i'm not sure what i have to do 😓 So when starting ai-studio with a running machine + playground 
 i should go to the playground page and find it running? Bc i still see the play action and if i click it generates a new container🤔 

My other scenario is, you are using the playground, turn off the machine bc you switch to another or whatever. ai-studio should be able to remove the adopted container as it is stopped now, then you turn the machine on again, the container is there stopped, yo start it, ai-studio should be able to adopt it again.

BTW maybe my assumption was wrong as i expected it to work differently. How should i test it?

feloy: Here is a demo on Linux to see what you can test:

https://github.com/projectatomic/ai-studio/assets/9973512/e290e3c2-4fb4-427b-a333-c4a196ccdf66

> My other scenario is, you are using the playground, turn off the machine bc you switch to another or whatever. ai-studio should be able to remove the adopted container as it is stopped now, then you turn the machine on again, the container is there stopped, yo start it, ai-studio should be able to adopt it again.

For now, we are not checking when a machine is stopped, to mark playground as stopped. It would be an extension of the PR https://github.com/projectatomic/ai-studio/pull/194 (and not particularly part of this PR as it is a different logic).

As said in my previous message, when you stop a playground container, it is also deleted (because marked as AUTOREMOVE). So you will never have a playground container (not even stopped) when you start a podman machine.

lstocchi: If you stop the machine the container will be there, not deleted and you can restart it once you restart the machine.

image

lstocchi (approved):
Tested and works fine for this specific use case

packages/backend/src/managers/podmanConnection.ts
feloy: AFAIK, it is not the expected behaviour for a container marked as auto-removable. In Inspect of this container, is it marked as this?
```
  "HostConfig": {
    [...]
    "AutoRemove": true,
```

feloy: Not that I don't want to support these stopped containers, but if we consider having stopped playground containers, this adds a third state for the playground container in the UI (paused in addition to not-running and running), which we would also have to manage in the UI (set the playground in pause, etc). That would be a bigger change than just detecting a stopped container is started again.

lstocchi: Sorry for the slow response. Done it again now and i confirm that the Autoremove flag is there

PR #204, opened 2024-01-31
fix: upload model on podman machine on WSL to speed loading

lstocchi:
So, it should be ready.
When you are on WSL the model gets uploaded to the podman machine. This makes us loses like 20sec during the start of application but it makes us gain minutes when the model is actually loaded by the model service.

So to test it,
1) run an application
2) when it reaches the "application is started" state, if you see the model service logs, it should be already running with the model loaded
3) if you check inside the podman machine you should see the model 
`podman machine ssh ls`

If the user system is no Windows, it works as before. Qemu will be added on another PR

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -133,10 +133,17 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
       const configAndFilteredContainers = this.getConfigAndFilterContainers(recipe.config, localFolder);
 
       // get model by downloading it or retrieving locally
-      const modelPath = await this.modelsManager.requestDownloadModel(model, {
+      let modelPath = await this.modelsManager.requestDownloadModel(model, {
         'recipe-id': recipe.id,
         'model-id': model.id,
       });
+
+      // upload model to podman machine if user system is supported
+      modelPath = await this.modelsManager.uploadModelToPodmanMachine(model, modelPath, {
+        'recipe-id': recipe.id,
+        'model-id': model.id,
+      });
+
 
       // build all images, one per container (for a basic sample we should have 2 containers = sample app + model service)
       const images = await this.buildImages(
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -144,7 +144,6 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
         'model-id': model.id,
       });
 
-
       // build all images, one per container (for a basic sample we should have 2 containers = sample app + model service)
       const images = await this.buildImages(
         recipe,
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -24,9 +24,12 @@ import { Messages } from '@shared/Messages';
 import type { CatalogManager } from './catalogManager';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import * as podmanDesktopApi from '@podman-desktop/api';
-import { Downloader, type DownloadEvent, isCompletionEvent, isProgressEvent } from '../utils/downloader';
+import { Downloader } from '../utils/downloader';
 import type { TaskRegistry } from '../registries/TaskRegistry';
 import type { Task } from '@shared/src/models/ITask';
+import type { ProgressiveEvent } from '../utils/progressiveEvent';
+import { isCompletionProgressiveEvent, isProgressProgressiveEvent } from '../utils/progressiveEvent';
+import { Uploader } from '../models/uploader';
 
 export class ModelsManager implements Disposable {
   #modelsDir: string;
@@ -201,7 +204,7 @@ export class ModelsManager implements Disposable {
     // If we have an existing downloader running we subscribe on its events
     return new Promise((resolve, reject) => {
       const disposable = existingDownloader.onEvent(event => {
-        if (!isCompletionEvent(event)) return;
+        if (!isCompletionProgressiveEvent(event)) return;
 
         switch (event.status) {
           case 'completed':
@@ -215,7 +218,7 @@ export class ModelsManager implements Disposable {
     });
   }
 
-  private onDownloadEvent(event: DownloadEvent): void {
+  private onDownloadEvent(event: ProgressiveEvent): void {
     // Always use the task registry as source of truth for tasks
     const tasks = this.taskRegistry.getTasksByLabels({ 'model-pulling': event.id });
     if (tasks.length === 0) {
@@ -225,10 +228,10 @@ export class ModelsManager implements Disposable {
     }
 
     tasks.forEach(task => {
-      if (isProgressEvent(event)) {
+      if (isProgressProgressiveEvent(event)) {
         task.state = 'loading';
         task.progress = event.value;
-      } else if (isCompletionEvent(event)) {
+      } else if (isCompletionProgressiveEvent(event)) {
         // status error or canceled
         if (event.status === 'error' || event.status === 'canceled') {
           task.state = 'error';
@@ -299,5 +302,46 @@ export class ModelsManager implements Disposable {
     // perform download
     await downloader.perform(model.id);
     return downloader.getTarget();
+  }
+
+  async uploadModelToPodmanMachine(model: ModelInfo, localModelPath: string, labels?: { [key: string]: string }): Promise<string> {
+    const task: Task = this.taskRegistry.createTask(`Uploading model ${model.name}`, 'loading', {
+      ...labels,
+      'model-uploading': model.id,
+    });
+
+    const uploader = new Uploader(localModelPath);
+    uploader.onEvent((event: ProgressiveEvent) => {
+      if (isProgressProgressiveEvent(event)) {
+        task.state = 'loading';
+        task.progress = event.value;
+      } else if (isCompletionProgressiveEvent(event)) {
+        // status error or canceled
+        if (event.status === 'error' || event.status === 'canceled') {
+          task.state = 'error';
+          task.progress = undefined;
+          task.error = event.message;
+
+          // telemetry usage
+          this.telemetry.logError('model.upload', {
+            'model.id': model.id,
+            message: 'error uploading model',
+            error: event.message,
+            durationSeconds: event.duration,
+          });
+        } else {
+          task.state = 'success';
+          task.progress = 100;
+
+          // telemetry usage
+          this.telemetry.logUsage('model.upload', { 'model.id': model.id, durationSeconds: event.duration });
+        }
+      }
+
+      this.taskRegistry.updateTask(task); // update task
+    });
+
+    // perform download
+    return await uploader.perform(model.id);
   }
 }
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -27,9 +27,9 @@ import * as podmanDesktopApi from '@podman-desktop/api';
 import { Downloader } from '../utils/downloader';
 import type { TaskRegistry } from '../registries/TaskRegistry';
 import type { Task } from '@shared/src/models/ITask';
-import type { ProgressiveEvent } from '../utils/progressiveEvent';
-import { isCompletionProgressiveEvent, isProgressProgressiveEvent } from '../utils/progressiveEvent';
-import { Uploader } from '../models/uploader';
+import type { ProgressiveEvent } from '../models/progressiveEvent';
+import { isCompletionProgressiveEvent, isExecutingProgressiveEvent } from '../models/progressiveEvent';
+import { Uploader } from '../utils/uploader';
 
 export class ModelsManager implements Disposable {
   #modelsDir: string;
@@ -228,7 +228,7 @@ export class ModelsManager implements Disposable {
     }
 
     tasks.forEach(task => {
-      if (isProgressProgressiveEvent(event)) {
+      if (isExecutingProgressiveEvent(event)) {
         task.state = 'loading';
         task.progress = event.value;
       } else if (isCompletionProgressiveEvent(event)) {
@@ -316,7 +316,7 @@ export class ModelsManager implements Disposable {
 
     const uploader = new Uploader(localModelPath);
     uploader.onEvent((event: ProgressiveEvent) => {
-      if (isProgressProgressiveEvent(event)) {
+      if (isExecutingProgressiveEvent(event)) {
         task.state = 'loading';
         task.progress = event.value;
       } else if (isCompletionProgressiveEvent(event)) {
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -304,7 +304,11 @@ export class ModelsManager implements Disposable {
     return downloader.getTarget();
   }
 
-  async uploadModelToPodmanMachine(model: ModelInfo, localModelPath: string, labels?: { [key: string]: string }): Promise<string> {
+  async uploadModelToPodmanMachine(
+    model: ModelInfo,
+    localModelPath: string,
+    labels?: { [key: string]: string },
+  ): Promise<string> {
     const task: Task = this.taskRegistry.createTask(`Uploading model ${model.name}`, 'loading', {
       ...labels,
       'model-uploading': model.id,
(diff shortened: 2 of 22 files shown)


jeffmaury (commented):
The only concern I have is that it uses the default Podman machine which may be stopped and deployments are performed against the first Podman machine that is running

lstocchi:
@jeffmaury updated

packages/backend/src/utils/progressiveEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface ProgressiveEvent {
```
axel7083, line 19: I don't understand why you renamed from `ProgressEvent` to `ProgressiveEvent` ?

packages/backend/src/utils/progressiveEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+  status: 'error' | 'completed' | 'progress' | 'canceled';
+  message?: string;
+}
+
+export interface CompletionProgressiveEvent extends ProgressiveEvent {
+  status: 'completed' | 'error' | 'canceled';
+  duration: number;
+}
+
+export interface ProgressProgressiveEvent extends ProgressiveEvent {
```
axel7083, line 30: ProgressProgressiveEvent ? Why such wording ?

packages/backend/src/managers/modelsManager.ts
```diff
@@ -225,10 +228,10 @@ export class ModelsManager implements Disposable {
     }
 
     tasks.forEach(task => {
-      if (isProgressEvent(event)) {
+      if (isProgressProgressiveEvent(event)) {
```
axel7083, line 231: Is the rename really necessary ?

packages/backend/src/models/uploader.ts
```diff
@@ -0,0 +1,85 @@
+/**********************************************************************
```
axel7083, line 1: This should not be in the `models` directory

packages/backend/src/utils/progressiveEvent.ts
```diff
@@ -0,0 +1,50 @@
+/**********************************************************************
```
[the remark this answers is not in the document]
lstocchi: Why in models? They are generics events utility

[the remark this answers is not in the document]
axel7083: The `Model` directory does not mean is it related to machine learning model, but is made for entity definition (aka interfaces)

packages/backend/src/utils/progressiveEvent.ts
lstocchi: This way it's self-explanatory if you read it in some other code. If you just read ProgressEvent, you need to check what it extends. ProgressEvent looks to me too generic but i'm open to suggestions

packages/backend/src/managers/modelsManager.ts
lstocchi: To keep in line with the naming used. It can be updated with the above if you have some suggestion

packages/backend/src/utils/progressiveEvent.ts
axel7083: I feel like `Progress` and `Progressive` are too close and could lead to some confusion, we might want to find something different to improve readability.

`BaseEvent` instead of `DownloadEvent` and keep `ProgressEvent` to talk about the real progress.

or

`TransferEvent` if we want to keep some _action/wording_ of what is happening 

> I have a preference for BaseEvent, as we could create its own `IBaseEvent` class, and keep it separate from the logic of the Downloader/Uploader

packages/backend/src/managers/modelsManager.ts
axel7083: I did in https://github.com/projectatomic/ai-studio/pull/204#discussion_r1519602557 ! Tell me what you think about it

packages/backend/src/utils/progressiveEvent.ts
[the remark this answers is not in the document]
lstocchi: I wonder when we decided it. I am a bit puzzled as we don't stick to this organization at all. 
Just by looking at AIConfig i see that it contains functions not related to what you're saying.

packages/backend/src/utils/progressiveEvent.ts
lstocchi: progressing = happening or developing gradually or in stages. Look to me that it expresses well what it does and it can be reused. 
By using TrasferEvent we are binding it only to transferring actions which may not be true in future, no?
Maybe `ExecutingProgressiveEvent` instead of `ProgressProgressiveEvent`?

axel7083: here is the `Download` event

```
export interface DownloadEvent {
  id: string;
  status: 'error' | 'completed' | 'progress' | 'cancelled';
  message?: string;
}
```

This seems very _basic_ and generic, I still think something _short_ and easy to read like `BaseEvent` would be more suitable than `ExecutingProgressiveEvent`...

lstocchi:
@jeffmaury rebased

packages/backend/src/models/progressiveEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export interface ProgressiveEvent {
```
axel7083, line 19: _I still think this should be named `BaseEvent` to avoid meaningless interface `ExecutingProgressiveEvent` mixing too many different wording._

lstocchi: renamed

packages/backend/src/utils/WSLUploader.ts
```diff
@@ -0,0 +1,62 @@
...
+    const driveLetter = localPath.charAt(0);
+    const convertToMntPath = localPath
+      .replace(`${driveLetter}:\\`, `/mnt/${driveLetter.toLowerCase()}/`)
+      .replace(/\\/g, '/');
+    const remotePath = `/home/user/${path.basename(convertToMntPath)}`;
+    const machine = await getFirstRunningMachine();
+    // check if model already loaded on the podman machine
+    let existsRemote = true;
+    try {
+      await podmanDesktopApi.process.exec(getPodmanCli(), ['machine', 'ssh', machine.Name, 'stat', remotePath]);
```
axel7083, line 43: _Not related directly to this PR_

@benoitf I am curious about your opinion, should we simplify this process for extensions (in a far future) ? Like offering an api to create some kind of ssh session ?

packages/backend/src/managers/applicationManager.ts
```diff
@@ -133,7 +133,13 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
...
 
       // get model by downloading it or retrieving locally
-      const modelPath = await this.modelsManager.requestDownloadModel(model, {
+      let modelPath = await this.modelsManager.requestDownloadModel(model, {
+        'recipe-id': recipe.id,
+        'model-id': model.id,
+      });
+
+      // upload model to podman machine if user system is supported
+      modelPath = await this.modelsManager.uploadModelToPodmanMachine(model, modelPath, {
```
axel7083, line 142: We might want a fallback in case it failed ? Like using the default mounting ? let's say for XYZ reasons, it failed ?

axel7083:
While starting an Application I got the following 

| AI Studio | Pod details |
| --- | --- |
| image | image |

Is this expected ?

packages/backend/src/managers/applicationManager.ts
lstocchi: when i talked with Jeff about adding an option to allow users decide to upload or not the model, he said that he didn't want it, so the upload should be the default behavior, so i thought that if the upload fails the whole workflow fails. Like if the download step fails. But it can be changed, yes

lstocchi:
> Is this expected ?

No, have you done something in particular? I'll try now

axel7083:
> > Is this expected ?
> 
> No, have you done something in particular? I'll try now

(1) checkout 
(2) build
(3) yarn watch ...
(4) go to ai studio
(5) click on chatbot recipe
(6) click on start (I already have the model downloaded)

lstocchi:
@axel7083 i was not able to replicate it but i fixed an error that prevented the workflow to fail. So, atleast you should now see the error icon and the error description. If you could copy/paste your error i can try to figure out what's the problem

axel7083:
> @axel7083 i was not able to replicate it but i fixed an error that prevented the workflow to fail. So, atleast you should now see the error icon and the error description. If you could copy/paste your error i can try to figure out what's the problem

https://github.com/projectatomic/ai-studio/assets/42176370/017331ce-2c0d-4be9-888a-6cb036faa44b

Here is a video, I re-did the step I said above

axel7083:
> @axel7083 i was not able to replicate it but i fixed an error that prevented the workflow to fail. So, atleast you should now see the error icon and the error description. If you could copy/paste your error i can try to figure out what's the problem

https://github.com/projectatomic/ai-studio/assets/42176370/55c269b8-b5e2-4380-9923-bc089c11ea8a

Did you push your commit ?

lstocchi:
> Did you push your commit ?

🤦 no, sorry. Pushed now

axel7083:
> > Did you push your commit ?
> 
> 🤦 no, sorry. Pushed now


https://github.com/projectatomic/ai-studio/assets/42176370/fe499227-5fd8-4a35-8af7-2e3967e8213b

lstocchi:
> Maybe my repo is too old ? (will retry after I ate)

You can try, the error seems related to the podman machine though 🤔 
What if you run `podman machine list --format json`? Does it work fine? Do you have any value in the podman cli path setting?

axel7083:
> > Maybe my repo is too old ? (will retry after I ate)
> 
> You can try, the error seems related to the podman machine though 🤔 What if you run `podman machine list --format json`? Does it work fine? Do you have any value in the podman cli path setting?
> 
> I'll add a check to prevent this kind of error

```
PS C:\Users\axels> podman machine list --format json
[
    {
        "Name": "podman-machine-default",
        "Default": false,
        "Created": "2023-09-23T19:18:48.434351+02:00",
        "Running": true,
        "Starting": false,
        "LastUp": "2024-03-12T12:55:57.0740947+01:00",
        "Stream": "35",
        "VMType": "wsl",
        "CPUs": 12,
        "Memory": "689012736",
        "DiskSize": "72994521088",
        "Port": 64772,
        "RemoteUsername": "user",
        "IdentityPath": "C:\\Users\\axels\\.ssh\\podman-machine-default",
        "UserModeNetworking": false
    }
]
```

lstocchi:
> > > Maybe my repo is too old ? (will retry after I ate)
> > 
> > 
> > You can try, the error seems related to the podman machine though 🤔 What if you run `podman machine list --format json`? Does it work fine? Do you have any value in the podman cli path setting?
> > I'll add a check to prevent this kind of error
> 
> ```
> PS C:\Users\axels> podman machine list --format json
> [
>     {
>         "Name": "podman-machine-default",
>         "Default": false,
>         "Created": "2023-09-23T19:18:48.434351+02:00",
>         "Running": true,
>         "Starting": false,
>         "LastUp": "2024-03-12T12:55:57.0740947+01:00",
>         "Stream": "35",
>         "VMType": "wsl",
>         "CPUs": 12,
>         "Memory": "689012736",
>         "DiskSize": "72994521088",
>         "Port": 64772,
>         "RemoteUsername": "user",
>         "IdentityPath": "C:\\Users\\axels\\.ssh\\podman-machine-default",
>         "UserModeNetworking": false
>     }
> ]
> ```

Only that machine? I wonder why you don't have it set to default. Podman desktop should ask you to set the active machine as the default one.

axel7083:
> > > > Maybe my repo is too old ? (will retry after I ate)
> > > 
> > > 
> > > You can try, the error seems related to the podman machine though 🤔 What if you run `podman machine list --format json`? Does it work fine? Do you have any value in the podman cli path setting?
> > > I'll add a check to prevent this kind of error
> > 
> > 
> > ```
> > PS C:\Users\axels> podman machine list --format json
> > [
> >     {
> >         "Name": "podman-machine-default",
> >         "Default": false,
> >         "Created": "2023-09-23T19:18:48.434351+02:00",
> >         "Running": true,
> >         "Starting": false,
> >         "LastUp": "2024-03-12T12:55:57.0740947+01:00",
> >         "Stream": "35",
> >         "VMType": "wsl",
> >         "CPUs": 12,
> >         "Memory": "689012736",
> >         "DiskSize": "72994521088",
> >         "Port": 64772,
> >         "RemoteUsername": "user",
> >         "IdentityPath": "C:\\Users\\axels\\.ssh\\podman-machine-default",
> >         "UserModeNetworking": false
> >     }
> > ]
> > ```
> 
> Only that machine? I wonder why you don't have it set to default. Podman desktop should ask you to set the active machine as the default one.

Apparently it is not, but this case should be handle I guess

lstocchi:
> Apparently it is not, but this case should be handle I guess

I was able to replicate your case by using rootful machine.
So now i use the running connection name to recreate the podman machine, it should work.
Whenever you have time, please give it another try. Thanks

packages/backend/src/utils/WSLUploader.ts
```diff
@@ -0,0 +1,66 @@
...
+      .replace(`${driveLetter}:\\`, `/mnt/${driveLetter.toLowerCase()}/`)
+      .replace(/\\/g, '/');
+    const remotePath = `/home/user/${path.basename(convertToMntPath)}`;
+    const machineName = getFirstRunningMachineName();
+
+    if (!machineName) {
+      throw new Error('No podman machine is running');
+    }
+    // check if model already loaded on the podman machine
+    let existsRemote = true;
```
axel7083, line 45: In another PR, that would be better to have this as a callable method, so we could allow from the model list to upload it to the podman machine manually by the user. We would also probably want to be able to delete them

lstocchi: yes, I agree 👍

packages/backend/src/utils/WSLUploader.ts
```diff
@@ -0,0 +1,66 @@
...
+  async upload(localPath: string): Promise<string> {
+    if (!localPath) {
+      throw new Error('invalid local path');
+    }
+
+    const driveLetter = localPath.charAt(0);
+    const convertToMntPath = localPath
+      .replace(`${driveLetter}:\\`, `/mnt/${driveLetter.toLowerCase()}/`)
+      .replace(/\\/g, '/');
+    const remotePath = `/home/user/${path.basename(convertToMntPath)}`;
```
axel7083, line 38: Can we choose a better path ? I think it is a bit intrusive to place it inside `/home/user/`

axel7083:
> > Apparently it is not, but this case should be handle I guess
> 
> I was able to replicate your case by using rootful machine. So now i use the running connection name to recreate the podman machine, it should work. Whenever you have time, please give it another try. Thanks

 Once uploaded, the model is soo fast to start !! Love it.

jeffmaury (approved):
LGTM. Great improvment

packages/backend/src/utils/WSLUploader.ts
lstocchi: Any preference? or just create a new folder under user?

jeffmaury: ai-studio folder under $HOME ?

axel7083: Why not something like `/home/user/ai-studio/models/` ? maybe too much nested ? I would rather have it clear enough

lstocchi: To build under $HOME one must be root, i'm going to create it under user as @axel7083 suggested

axel7083 (approved):
Just tested and works fine ! Great job 👏 !!

PR #210, opened 2024-02-01
feat: using podman-desktop modals instead of custom one

axel7083:
### What does this PR do?

Uses the Podman-Desktop provided modals instead of custom one.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/4e51ed2d-cb17-4ead-a856-76e4c14c7bdc

Fixes https://github.com/projectatomic/ai-studio/issues/203

### DIFF
--- a/packages/frontend/src/lib/table/model/ModelColumnActions.svelte
+++ b/packages/frontend/src/lib/table/model/ModelColumnActions.svelte
@@ -6,7 +6,7 @@ import ListItemButtonIcon from "../../button/ListItemButtonIcon.svelte";
 import { studioClient } from "/@/utils/client";
 export let object: ModelInfo;
 
-const deleteModel = () => {
+function deleteModel() {
   studioClient.requestRemoveLocalModel(object.id).catch((err) => {
     console.error(`Something went wrong while trying to delete model ${String(err)}.`);
   })
--- a/packages/frontend/src/lib/table/model/ModelColumnActions.svelte
+++ b/packages/frontend/src/lib/table/model/ModelColumnActions.svelte
@@ -4,19 +4,12 @@ import { faTrash } from "@fortawesome/free-solid-svg-icons";
 import { faFolderOpen } from "@fortawesome/free-solid-svg-icons";
 import ListItemButtonIcon from "../../button/ListItemButtonIcon.svelte";
 import { studioClient } from "/@/utils/client";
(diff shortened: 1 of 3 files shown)


axel7083:
Not related directly to this PR, but the following https://github.com/containers/podman-desktop/issues/5744 would be a nice touch !

packages/frontend/src/lib/table/model/ModelColumnActions.svelte
```diff
@@ -4,19 +4,12 @@ import { faTrash } from "@fortawesome/free-solid-svg-icons";
...
-let deleteConfirmVisible: boolean = false;
-
-function deleteModel() {
-  deleteConfirmVisible= true;
-}
-
-async function goDeleteModel() {
-  await studioClient.deleteLocalModel(object.id);
-  deleteConfirmVisible= false;
+const deleteModel = () => {
```
jeffmaury, line 9: What is the rationale for using this pattern and not function ?

axel7083: There are pros and cons, but in the end, just a matter of preferences

lstocchi (commented):
It seems that if there is some error while deleting we do not show anything to the user.
Maybe this can be done in a follow-up PR as you're just replacing the modal but, thinking at this future change, wouldn't it be better to await the showWarningMessage and have the `deleteLocalModel` outside the then?

packages/frontend/src/lib/table/model/ModelColumnActions.svelte
benoitf: I would say it's about consistency with existing code ?

I mean, if for example someone love tabs and another spaces, or one prefer 2 spaces rather than 4 spaces, etc if everyone use the matter of preferences then it could produce an overall inconsistent code.

axel7083:
> It seems that if there is some error while deleting we do not show anything to the user. Maybe this can be done in a follow-up PR as you're just replacing the modal but, thinking at this future change, wouldn't it be better to await the showWarningMessage and have the `deleteLocalModel` outside the then?

Yes, we could also use ShowError api inside the backend and try catch the deletion

lstocchi (commented):
If you could change the function "style" as commented by Jeff/Florent i'm good with it

lstocchi (approved):
LGTM

Issue #173, opened 2024-01-29
Import model


jeffmaury:
A user should be able to import a model into the model registry

vrothberg:
How does the workflow look like?  Will there be an import button on the models page?

jeffmaury:
> How does the workflow look like? Will there be an import button on the models page?

I'm not sure we have a mock up for it but that's the idea

lstocchi:
What do we mean by importing? Can i download any model from huggingface (or anywhere else) and import it into desktop or i have the model (one present in the catalog, like mistral) in another location of my pc and i want desktop to find it?

lstocchi:
@slemeur have answered it during the call. So i can download a model from huggingface and import it

slemeur:
Correct, consider that the "import model" will be allowing a user to load a local model file and start the model serving with that model file.

lstocchi:
@slemeur you mean just by clicking on import, it start the model serving? So it is just one shot? 

I mean if i close ai lab and open it the day after, do i see the imported model in the model list or i need to execute import again?

slemeur:
We should see that as extending the catalog of model. 
- So on the list of models, we would have a button "Import".
- User will be able to select the model from the local disk
- Once imported, the model is indicated with the list as a "user's model"
- Then, the user is able to start a model serving from that model, or start a playground for it

PR #212, opened 2024-02-01
feat: display models directory in models list

axel7083:
### What does this PR do?

Adding a banner, displaying the absoute directory used where the models are stored, plus a button to open the directory, using the api function added by @feloy 

### Screenshot / video of UI

image

Fixes https://github.com/projectatomic/ai-studio/issues/211

### How to test this PR?

- [x] Unit tests have been added

### DIFF
--- a/packages/frontend/src/pages/Models.spec.ts
+++ b/packages/frontend/src/pages/Models.spec.ts
@@ -1,0 +1,68 @@
+import '@testing-library/jest-dom/vitest';
+import { vi, test, expect } from 'vitest';
+import { screen, fireEvent, render, waitFor } from '@testing-library/svelte';
+import Models from '/@/pages/Models.svelte';
+
+const mocks = vi.hoisted(() => {
+  return {
+    getModelsDirectoryMock: vi.fn(),
+    openFileMock: vi.fn(),
+    localModelsMock: {
+      subscribe: (_f: (msg: any) => void) => {
+        return () => {};
+      },
+    },
+    modelsPullingSubscribeMock: vi.fn(),
+    modelsPullingMock: {
+      subscribe: (f: (msg: any) => void) => {
+        f(mocks.modelsPullingSubscribeMock());
+        return () => {};
+      },
+    },
+  };
+});
+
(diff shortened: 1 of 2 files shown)


feloy:
Do we want the button to be the primary one? Or would secondary be more appropriate?

axel7083:
> Do we want the button to be the primary one? Or would secondary be more appropriate?

I don't mind, if we think the secondary is better we can use it

lstocchi (commented):
I wonder why we need this. There is already the action on each model row. 
Which is the reasoning to have the location of the models list? Maybe we could add this in the settings and allows the user to customize the location?

axel7083:
> I wonder why we need this. There is already the action on each model row. Which is the reasoning to have the location of the models list? Maybe we could add this in the settings and allows the user to customize the location?

The reason is, if you have no models, you want to put yours so IA Studio can use it.

For now the directory is not configurable, but we could image a (?) button or gears icon to change it. For now we do not have preferences, so I think it is important to display it somewhere

lstocchi:
> The reason is, if you have no models, you want to put yours so IA Studio can use it.
> 
> For now the directory is not configurable, but we could image a (?) button or gears icon to change it. For now we do not have preferences, so I think it is important to display it somewhere

The problem is that you should also explain to the user that they do not need to drop the file in that folder but they need to create a subfolder where to put their model in otherwise i don't think it will be detected.
For me this will be removed by https://github.com/projectatomic/ai-studio/issues/173 .
So the user select the model to import and we take care of creating the subfolder and copying/moving the model in the correct location.

axel7083:
> > The reason is, if you have no models, you want to put yours so IA Studio can use it.
> > For now the directory is not configurable, but we could image a (?) button or gears icon to change it. For now we do not have preferences, so I think it is important to display it somewhere
> 
> The problem is that you should also explain to the user that they do not need to drop the file in that folder but they need to create a subfolder where to put their model in otherwise i don't think it will be detected. For me this will be implemented by #173 . So the user select the model to import and we take care of creating the subfolder and copying/moving the model in the correct location.

Agree

PR #217, opened 2024-02-02
fix: properly set error status on messages when catching an api error

axel7083:
### What does this PR do?

This PR is fixing the error propagation between the front and back. The errors was not properly sent between the processes, and was raison `Received incompatible message.`.

Fixes https://github.com/projectatomic/ai-studio/issues/216

### How to test this PR?

### DIFF
--- a/packages/shared/src/messages/MessageProxy.ts
+++ b/packages/shared/src/messages/MessageProxy.ts
@@ -80,9 +80,14 @@ export class RpcExtension {
         } as IMessageResponse);
       } catch (err: unknown) {
         let errorMessage: string;
-        if (err instanceof Error) errorMessage = err.message;
-        else if (typeof err === 'string') errorMessage = err;
-        else errorMessage = String(err);
+        // Depending on the object throw we try to extract the error message
+        if (err instanceof Error) {
+          errorMessage = err.message;
+        } else if (typeof err === 'string') {
+          errorMessage = err;
+        } else {
+          errorMessage = String(err);
+        }
 
         await this.webview.postMessage({
           id: message.id,
--- a/packages/shared/src/messages/MessageProxy.ts
+++ b/packages/shared/src/messages/MessageProxy.ts
@@ -79,13 +79,17 @@ export class RpcExtension {
           status: 'success',
         } as IMessageResponse);
       } catch (err: unknown) {
-        console.error(`Something went wrong on channel ${message.channel}`, err);
+        let errorMessage: string;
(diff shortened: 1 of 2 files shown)


packages/shared/src/messages/MessageProxy.ts
```diff
@@ -78,12 +78,14 @@ export class RpcExtension {
           body: result,
           status: 'success',
         } as IMessageResponse);
-      } catch (e) {
+      } catch (err: unknown) {
+        console.error(`Something went wrong on channel ${message.channel}`, err);
```
jeffmaury, line 82: Not sure we should log as the code is generic and having an error may be on the normal path

jeffmaury (changes requested):
Fix lint errors

packages/shared/src/messages/MessageProxy.ts
axel7083: Yes, I removed it

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -78,12 +78,18 @@ export class RpcExtension {
           body: result,
           status: 'success',
         } as IMessageResponse);
-      } catch (e) {
+      } catch (err: unknown) {
+        let errorMessage: string;
+        if (err instanceof Error) errorMessage = err.message;
+        else if (typeof err === 'string') errorMessage = err;
+        else errorMessage = String(err);
```
lstocchi, line 85: please add curly brackets so it's in line with the rest of the code

axel7083: Done

lstocchi (approved):
LGTM

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -78,12 +78,23 @@ export class RpcExtension {
...
+          errorMessage = String(err);
+        }
+
         await this.webview.postMessage({
           id: message.id,
           channel: message.channel,
           body: undefined,
-          error: `Something went wrong on channel ${message.channel}: ${String(e)}`,
+          status: 'error',
+          error: errorMessage,
```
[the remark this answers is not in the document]
axel7083: image

No, because `Error` object in JS are not serializable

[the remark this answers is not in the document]
axel7083: And the traces are not relevant for the front. In an API communication, the message is important, so we can ignore the problems that occurs in the server

jeffmaury (approved):
LGTM

PR #222, opened 2024-02-02
fix: wait until playground is initialized

### LINKED ISSUE #147: Detect the playground container is ready
The playground container can get some time to start, and the user can send a request before the container is ready to reply.

It would be helpful to implement a readiness probe, so AI studio knows when the playground container is ready to reply.

jeffmaury:
Fixes #147

### What does this PR do?

Wait until the playground container is properly initialized

### Screenshot / video of UI

N/A

Fixes #147 

### How to test this PR?

Start a playground

### DIFF
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -215,11 +215,10 @@ export class PlayGroundManager {
       this.playgrounds.get(modelId).status === 'starting'
     ) {
       try {
-        const response = await fetch(`http://localhost:${freePort}`);
-        contacted = response.ok;
-        await timeout(1000);
+        await fetch(`http://localhost:${freePort}`);
+        contacted = true;
       } catch (err: unknown) {
-        /* empty */
+        await timeout(1000);
       }
     }
 
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -40,6 +40,8 @@ const LABEL_MODEL_PORT = 'ai-studio-model-port';
 // TODO: this should not be hardcoded
 const PLAYGROUND_IMAGE = 'quay.io/bootsy/playground:v0';
 
+const STARTING_TIME_MAX = 3600 * 1000;
+
 function findFirstProvider(): ProviderContainerConnection | undefined {
   const engines = provider
     .getContainerConnections()
@@ -203,6 +205,26 @@ export class PlayGroundManager {
           break;
       }
     });
+
+    let contacted = false;
+    const start = Date.now();
+    while (
+      Date.now() - start < STARTING_TIME_MAX &&
+      !contacted &&
+      this.playgrounds.get(modelId).status === 'starting'
+    ) {
+      try {
+        await fetch(`http://localhost:${freePort}`);
+        contacted = true;
+      } catch (err: unknown) {
+        /* empty */
+      }
+    }
+
+    if (!contacted) {
+      await containerEngine.stopContainer(image.engineId, result.id);
+      throw new Error(`Can't start playground for model ${modelId}`);
+    }
 
     this.updatePlaygroundState(modelId, {
       container: {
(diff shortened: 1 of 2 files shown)


benoitf (approved):
LGTM

packages/backend/src/managers/playground.ts
```diff
@@ -205,6 +207,26 @@ export class PlayGroundManager {
...
 
+    let contacted = false;
+    const start = Date.now();
+    while (
+      Date.now() - start < STARTING_TIME_MAX &&
+      !contacted &&
+      this.playgrounds.get(modelId).status === 'starting'
+    ) {
+      try {
+        await fetch(`http://localhost:${freePort}`);
```
dgolovin, line 218:
```suggestion
        const response  = await fetch(`http://localhost:${freePort}`);
```

packages/backend/src/managers/playground.ts
```diff
@@ -205,6 +207,26 @@ export class PlayGroundManager {
...
+        contacted = true;
```
dgolovin, line 219:
```suggestion
        contacted = response.ok;
```

dgolovin: nit: There are some use cases when response has error codes that does not trigger exceptions and host would be still unresponsive (network error in the 4xx–5xx range for example).

dgolovin (approved):

feloy:
The test are failing with heap memory limit. Is the `while` loop capped by time terminating during the tests?

packages/backend/src/managers/playground.ts
```diff
@@ -205,6 +207,26 @@ export class PlayGroundManager {
...
+    let contacted = false;
+    const start = Date.now();
+    while (
+      Date.now() - start < STARTING_TIME_MAX &&
+      !contacted &&
+      this.playgrounds.get(modelId).status === 'starting'
+    ) {
+      try {
+        const response = await fetch(`http://localhost:${freePort}`);
+        contacted = response.ok;
```
feloy, line 219: Some sleep between two tries would be less resource consuming

feloy:
I cannot make it work on mac. Let me add some sleep between requests to be able to check the logs, or I get by logs
```
INFO:     192.168.127.1:56827 - "GET / HTTP/1.1" 404 Not Found
```

jeffmaury:
> The test are failing with heap memory limit. Is the `while` loop capped by time terminating during the tests?

It is and seems related to another (flaky ?) test

packages/backend/src/managers/playground.ts
jeffmaury: Added

packages/backend/src/managers/playground.ts
```diff
@@ -217,6 +218,7 @@ export class PlayGroundManager {
       try {
         const response = await fetch(`http://localhost:${freePort}`);
         contacted = response.ok;
```
feloy, line 220: On my side (mac), I get a response with value `{}`. 
I would set `contacted = true` here, as the socket connection errors we want to detect go to the catch, so if we are at this point, socket connection is ok

jeffmaury: Should be fixed by 528e991

feloy (approved):
works on mac

packages/backend/src/managers/playground.ts
```diff
@@ -204,6 +207,26 @@ export class PlayGroundManager {
       }
     });
 
+    let contacted = false;
```
axel7083, line 210: Maybe we should create an utility function _somewhere_ that would manage the `retry` logics, we would reuse it later for other usage, with some timeout limits. See https://stackoverflow.com/a/45445291 for some retryable utility functions. I would prefer this approche as we could test each component individually (the retryable, the timeout, and finally the health state would be trivial)

Moreover, fetch is pretty bad at providing some custom timeout, we might not want wait for the maximum configured, see https://stackoverflow.com/questions/46946380/fetch-api-request-timeout

jeffmaury: Delayed to a follow up PR

Issue #179, opened 2024-01-29
Environment tab should be filled at startup


jeffmaury:
It should be filled from pods/containers that are associated with a local application. Mapping should be inspired from what compose does (using label)

feloy:
Fixed by #237

PR #237, opened 2024-02-06
feat: display environments list based on running pods

### LINKED ISSUE #178: Once application is started it should be visible in the Environments

feloy:
### What does this PR do?

Lists the running environments, based on the running pods with appropriate labels.

For now, only the id of the recipe is displayed, more to come...

Only the operations on podman machines from the GUI are detected (see issue https://github.com/containers/podman-desktop/issues/5842)

Fixes #178, #179

### How to test this PR?

- pull an application for a recipe
- check the recipe ID is listed in the Environments page
- Stop or delete the pod created for the app from the GUI or CLI
- check the recipe ID disappears from the environments list
- stop te podman engine from the GUI
- check the recipe ID disappears from the environments list
- restart the podman machine (from GUI) and restart the pod (from GUI or CLI)
- check the recipe ID is listed in the Environments page
- reload the AI studio extension
- check the recipe ID is still listed in the Environments page
- restart Podman Desktop
- check the recipe ID is still listed in the Environments page

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -38,6 +38,7 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
+    "@podman-desktop/api": "^0.0.202402061753-5c5d589",
     "openai": "^4.26.0",
     "simple-git": "^3.22.0"
   },
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -38,12 +38,11 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
-    "@podman-desktop/api": "^0.0.202402061753-5c5d589",
     "openai": "^4.26.0",
     "simple-git": "^3.22.0"
   },
   "devDependencies": {
-    "@podman-desktop/api": "1.7.0",
+    "@podman-desktop/api": "^0.0.202402061753-5c5d589",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^18",
     "vitest": "^1.1.0"
--- a/packages/backend/src/managers/environmentManager.ts
+++ b/packages/backend/src/managers/environmentManager.ts
@@ -34,6 +34,11 @@ export class EnvironmentManager {
 
   adoptRunningEnvironments() {
     this.podmanConnection.startupSubscribe(() => {
+      if (!containerEngine.listPods) {
+        // TODO(feloy) this check can be safely removed when podman desktop 1.8 is released
+        // and the extension minimal version is set to 1.8
+        return;
+      }
       containerEngine
         .listPods()
         .then(pods => {
--- a/packages/backend/src/managers/environmentManager.ts
+++ b/packages/backend/src/managers/environmentManager.ts
@@ -1,0 +1,129 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { type PodInfo, type Webview, containerEngine } from '@podman-desktop/api';
+import type { PodmanConnection } from './podmanConnection';
+import { LABEL_RECIPE_ID } from './applicationManager';
+import { MSG_ENVIRONMENTS_STATE_UPDATE } from '@shared/Messages';
+import type { EnvironmentState } from '@shared/src/models/IEnvironmentState';
+
+export class EnvironmentManager {
+  #environments: Map<string, EnvironmentState>;
+
+  constructor(
+    private webview: Webview,
+    private podmanConnection: PodmanConnection,
+  ) {
+    this.#environments = new Map();
+  }
+
+  adoptRunningEnvironments() {
+    this.podmanConnection.startupSubscribe(() => {
+      containerEngine
+        .listPods()
+        .then(pods => {
+          console.log('pods', pods);
+          const envsPods = pods.filter(pod => LABEL_RECIPE_ID in pod.Labels);
+          for (const podToAdopt of envsPods) {
+            this.adoptPod(podToAdopt);
+          }
+        })
+        .catch((err: unknown) => {
+          console.error('error during adoption of existing playground containers', err);
+        });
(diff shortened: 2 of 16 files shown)


packages/backend/package.json
```diff
@@ -38,6 +38,7 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
+    "@podman-desktop/api": "^0.0.202402061753-5c5d589",
```
axel7083, line 41: will the feature in `0.0.202402061753-5c5d589` be in 1.8 ?

feloy: @benoitf These 2 PRs will be needed by AI Studio in the next release:
- https://github.com/containers/podman-desktop/pull/5862
- https://github.com/containers/podman-desktop/pull/5864

Do I need to 'register' them somewhere, or will they be part of the next release automatically?

axel7083: @benoitf

benoitf: automatic as it's in the main branch and 1.8 will use the main branch

benoitf: but it means you can't cut new releases of AI studio using that API until 1.8 is out or you need to add some 'fallback' in the code

like if (api.engine.listPods) {...} else {....}

lstocchi: you should update the entry in devDependencies or it will be duplicated

packages/backend/src/managers/environmentManager.ts
```diff
@@ -0,0 +1,134 @@
...
+    private webview: Webview,
+    private podmanConnection: PodmanConnection,
+  ) {
+    this.#environments = new Map();
+  }
+
+  adoptRunningEnvironments() {
+    this.podmanConnection.startupSubscribe(() => {
+      if (!containerEngine.listPods) {
+        // TODO(feloy) this check can be safely removed when podman desktop 1.8 is released
```
lstocchi, line 38: can we open an issue to keep track of this and the other similar TODO?

lstocchi (commented):
I wonder why we don't show the pods when they are stopped. Shouldn't we be able to start an environment from this page later on?

packages/backend/src/managers/environmentManager.ts
feloy: right, created #242

feloy:
> I wonder why we don't show the pods when they are stopped. Shouldn't we be able to start an environment from this page later on?

I didn't want to complexify for now, by keeping only 2 states (running and non-running), and as restarting the pod is not in the list of features wwe want for now.

Also, as the start of the app container is managed by the extension (to restart it when the model is up and running), the user cannot simply restart the pod from podman desktop or the CLI, the extension would have to manage it, and that seems out of scope to me

packages/backend/package.json
```diff
@@ -38,11 +38,12 @@
...
```
benoitf, line 41: FYI you should not have the API as a direct dependency (it's a dev dependency)

lstocchi:
> Also, as the start of the app container is managed by the extension (to restart it when the model is up and running), the user cannot simply restart the pod from podman desktop or the CLI, the extension would have to manage it, and that seems out of scope to me

I expect i can start an environment from this page otherwise what's the difference from the pod list page? It's just a filter.

I would expect that the environment store would keep a list of all environments found (running and stopped) then the svelte page filters it based on its needs. Maybe another page that uses this store would need to filter non-running or something else.

feloy:
> > Also, as the start of the app container is managed by the extension (to restart it when the model is up and running), the user cannot simply restart the pod from podman desktop or the CLI, the extension would have to manage it, and that seems out of scope to me
> 
> I expect i can start an environment from this page otherwise what's the difference from the pod list page? It's just a filter.
> 
> I would expect that the environment store would keep a list of all environments found (running and stopped) then the svelte page filters it based on its needs. Maybe another page that uses this store would need to filter non-running or something else.

I have added some status in the latest PR, to try and support the "stopped" pods. But there is a gray zone between deleted pod and running pod, where some containers are running only (status = Degraded in the pod), but I don't think we are getting enough events to correctly handle this case.

Please test and tell me what you think. I have added the status in the Environments page.

lstocchi:
> I have added some status in the latest PR, to try and support the "stopped" pods. But there is a gray zone between deleted pod and running pod, where some containers are running only (status = Degraded in the pod), but I don't think we are getting enough events to correctly handle this case.

Got it. Inside the `watchPod -> containerEngine.onEvent` method do we only receive the remove, start and stop state? If we receive other events we could update the environments store if there is some change in a pod state. 

> Please test and tell me what you think. I have added the status in the Environments page.

Great, thanks! 
The problem i have is that when i start a recipe, the pod state is running at first (bc all containers start fine), then the sample app fails to connect to the model service and the pod gets degraded, but as we just listen to the start, stop and delete events, i see the env in running state even if it should be 'starting' as the pod is degraded. But idk if we have enough info (see comment above) to fix it. If not, or you want to handle it later, we can remove the status from displaying, and re-add it on a follow-up PR. I also have to use this for my PR https://github.com/projectatomic/ai-studio/pull/223 and i can look at it.

lstocchi:
> So, I'll remove the latest commit (and keep it locally in another branch in case we need it later)

I would have preferred to just filter it in svelte but i don't want to block this even further.

packages/frontend/src/lib/table/environment/ColumnName.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+  import type { EnvironmentState } from "@shared/src/models/IEnvironmentState";
+  import type { ModelInfo } from "@shared/src/models/IModelInfo";
+  export let object: EnvironmentState;
+</script>
+
+<div class="text-sm text-gray-700">
+  {object.recipeId}
```
jeffmaury, line 8: Can't we display the state as well ?

packages/backend/src/managers/environmentManager.ts
```diff
@@ -0,0 +1,134 @@
...
+          }
+        })
+        .catch((err: unknown) => {
+          console.error('error during adoption of existing playground containers', err);
+        });
+    });
+
+    this.podmanConnection.onMachineStop(() => {
+      // Podman Machine has been stopped, we consider all recipe pods are stopped
+      this.#environments.clear();
```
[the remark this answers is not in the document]
feloy: How can I have several machines? On Mac, I'm able to start only one Podman machine at a given time

packages/frontend/src/lib/table/environment/ColumnName.svelte
feloy: I'm waiting the design from Mo to work on the content here

packages/backend/src/managers/environmentManager.ts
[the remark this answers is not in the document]
jeffmaury: This is allowed on Windows

jeffmaury (approved):

packages/backend/src/managers/environmentManager.ts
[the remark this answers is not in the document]
feloy: I'll have a look on Windows and make a follow up PR if necessary. Thanks

PR #239, opened 2024-02-07
feat: create custom Readable to share stores logic

lstocchi:
### What does this PR do?

This PR adds the event-store class so we can share some common code between stores.

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

1. run tests or start studio and check that everything works as before. Only the recipe store has been updated for the moment

### DIFF
--- a/packages/frontend/src/stores/local-models.spec.ts
+++ b/packages/frontend/src/stores/local-models.spec.ts
@@ -1,0 +1,77 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+import { localModels } from './local-models';
+import { MSG_NEW_LOCAL_MODELS_STATE } from '@shared/Messages';
+import { rpcBrowser } from '../utils/client';
+import type { Unsubscriber } from 'svelte/store';
+
+const mocks = vi.hoisted(() => {
+  return {
+    getLocalModelsMock: vi.fn().mockResolvedValue([]),
+  };
+});
+
+vi.mock('../utils/client', async () => {
+  const subscriber = new Map();
+  const rpcBrowser = {
+    invoke: (msgId: string, _: unknown[]) => {
+      const f = subscriber.get(msgId);
+      f();
+    },
+    subscribe: (msgId: string, f: (msg: any) => void) => {
+      subscriber.set(msgId, f);
+      return {
+        unsubscribe: () => {
+          subscriber.clear();
+        },
+      };
+    },
+  };
+  return {
+    rpcBrowser,
+    studioClient: {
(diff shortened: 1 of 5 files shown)


feloy:
This complexifies a lot the code. What are the advantages of reusing this code from podman desktop, compared to the current code? 

What is the advantage of using a `Writable` instead of a `Readable`? I appreciate that the `Readable` guarantees that changes will come from the store only, and no other part can send new state to it.

lstocchi:
> What is the advantage of using a `Writable` instead of a `Readable`? I appreciate that the `Readable` guarantees that changes will come from the store only, and no other part can send new state to it.

For testing

jeffmaury (commented):
Seems not all stores are migrated, any reason ?

lstocchi:
> Seems not all stores are migrated, any reason ?

I wanted to see if there was any issue with it before updating all stores. I updated the recipe only because i was working on it and i noticed that it was complex to write a test that was simulating multiple state changes.

packages/frontend/src/stores/catalog.ts
```diff
@@ -10,15 +29,11 @@ const emptyCatalog = {
...
-  });
-  // Initialize the store manually
-  studioClient.getCatalog().then(state => {
-    set(state);
-  });
-  return () => {
-    sub.unsubscribe();
-  };
-});
+export const catalog: Writable<Catalog> = writable(emptyCatalog);
```
jeffmaury, line 32: Are those exports needed

packages/frontend/src/stores/catalog.ts
```diff
@@ -10,15 +29,11 @@ const emptyCatalog = {
...
+
+export const catalogEventStore = new EventStore<Catalog>(catalog, studioClient.getCatalog, MSG_NEW_CATALOG_STATE);
+const catalogEventStoreInfo = await catalogEventStore.setup();
+
+export const fetchCatalog = async () => {
```
jeffmaury, line 37: I don't see where this method is being used

lstocchi:
Still in progress.  Forgot to mark as draft, @feloy noticed a misbehavior on the localModel's page. I'm investigating as it seems the rpcbrowser is not actually working

feloy:
> Still in progress. Forgot to mark as draft, @feloy noticed a misbehavior on the localModel's page. I'm investigating as it seems the rpcbrowser is not actually working

For the problem I have raised (the name in the models page is not updated when we update the catalog file), I finally understand that the problem does not come from the stores, but from the Models page, as we are not subscribing to the catalog, so we cannot get the new name.

You probably says the rpcBrowser does not work because you get an error "Message not supported" when you update the catalog file. The reason is that because the Models page does not subscribe to the catalog store, the message is received but is not recognized by the rpcBrowser, because there is no subscriber: this is not the problem (we should hide the message in this case probably, or check that the message is registered, even if there is no subscriber)

lstocchi:
> > Still in progress. Forgot to mark as draft, @feloy noticed a misbehavior on the localModel's page. I'm investigating as it seems the rpcbrowser is not actually working
> 
> For the problem I have raised (the name in the models page is not updated when we update the catalog file), I finally understand that the problem does not come from the stores, but from the Models page, as we are not subscribing to the catalog, so we cannot get the new name.
> 
> You probably says the rpcBrowser does not work because you get an error "Message not supported" when you update the catalog file. The reason is that because the Models page does not subscribe to the catalog store, the message is received but is not recognized by the rpcBrowser, because there is no subscriber: this is not the problem (we should hide the message in this case probably, or check that the message is registered, even if there is no subscriber)

yes, correct. Already worked on a fix yesterday but i need to add some tests. I'll push it soon

lstocchi:
@feloy @jeffmaury updated it. I created a custom writable so when subscribing it subscribe the rcpBrowser to all events specified by the store. 
E.g the localModels depends on the localModelState and CatalogState - so if you update the catalog, the page is refreshed automatically without having to change page.

I would like not to cover the other stores in this PR as i also need to update the single pages that subscribe to them and if i tackle everything here it becomes a mess. Once this is merged i'll work on the other stores separately in small PRs so they are easily testable as well

lstocchi:
@feloy i saw your PR to mock writable when using readable for testing 👍 I updated this PR to use a Readable as your preference

feloy (approved):

PR #244, opened 2024-02-07
feat: checkout status check before starting application

axel7083:
### What does this PR do?

This is adding verification on the checkout tasks, which was simply skipping the cloning if the folder already exist.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/da8dbdc6-4fb8-48c3-9ad5-897fda847152

Fixes https://github.com/projectatomic/ai-studio/issues/243

### How to test this PR?

- [x] Unit tests has been added

**Manually**
(1) Go to the local repository and checkout an old commit
(2) Go to AI-Studio and click on `Run application`
(3) Should have a modal telling you the error

### DIFF
--- a/packages/backend/src/managers/gitManager.ts
+++ b/packages/backend/src/managers/gitManager.ts
@@ -46,7 +46,7 @@ export class GitManager {
     directory: string,
     origin: string,
     ref?: string,
-  ): Promise<{ ok?: boolean; error?: string; updatable?: boolean }> {
+  ): Promise<{ ok?: boolean; error?: string }> {
     const remotes: RemoteWithRefs[] = await this.getRepositoryRemotes(directory);
 
     if (!remotes.some(remote => remote.refs.fetch === origin)) {
@@ -64,20 +64,19 @@ export class GitManager {
     }
 
     if (status.modified.length > 0) {
-      return { error: 'The local repois' };
+      return { error: 'The local repository has modified files.' };
     }
 
     // If we are not in HEAD
     if (status.behind !== 0 || status.ahead !== 0) {
       return {
         error: `The local repository is not up to date ${status.behind} behind, ${status.ahead} ahead.`,
-        updatable: true,
       };
     }
 
     // Ensure the branch tracked is the one we want
     if (ref !== undefined && status.tracking !== ref) {
-      return { error: 'The local repository is not tracking the right branch.', updatable: true };
+      return { error: 'The local repository is not tracking the right branch.' };
     }
   }
 }
--- a/packages/backend/src/managers/gitManager.ts
+++ b/packages/backend/src/managers/gitManager.ts
@@ -42,30 +42,41 @@ export class GitManager {
     return simpleGit(directory).status();
   }
 
-  async isRepositoryUpToDate(directory: string, origin: string, ref?: string): Promise<{ ok?: boolean, error?: string, updatable?: boolean }> {
+  async isRepositoryUpToDate(
+    directory: string,
+    origin: string,
+    ref?: string,
+  ): Promise<{ ok?: boolean; error?: string; updatable?: boolean }> {
     const remotes: RemoteWithRefs[] = await this.getRepositoryRemotes(directory);
 
-    if(!remotes.some((remote) => remote.refs.fetch === origin)) {
-      return { error: `The local repository does not have remote ${origin} configured. Remotes: ${remotes.map((remote) => `${remote.name} ${remote.refs.fetch} (fetch)`).join(',')}`};
+    if (!remotes.some(remote => remote.refs.fetch === origin)) {
+      return {
+        error: `The local repository does not have remote ${origin} configured. Remotes: ${remotes
+          .map(remote => `${remote.name} ${remote.refs.fetch} (fetch)`)
+          .join(',')}`,
+      };
     }
 
     const status: StatusResult = await this.getRepositoryStatus(directory);
 
-    if(status.detached) {
+    if (status.detached) {
       return { error: 'The local repository is detached.' };
     }
 
-    if(status.modified.length > 0) {
-      return { error: 'The local repois'}
+    if (status.modified.length > 0) {
+      return { error: 'The local repois' };
     }
 
     // If we are not in HEAD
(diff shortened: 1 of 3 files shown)


jeffmaury (commented):
No simplegit knowledge but I have the impression that:

- does it allow to switch from one commit to another
- if there are some changes, we should allow the switch if git can handle it

axel7083:
> Looks to me weird as what would be the options to the user if the catalog is updated from 1 commit to another ?

That's an issue in all the cases, if there is a new commit, there is probably something which has changed, therefore need to update, we can in some case (commit behind) ask the user if he wants to pull updates ?

lstocchi (commented):
The detection works fine for me. My only doubt is the error message. Maybe we should provide an option that deletes and redownloads the repo automatically.
When i see the error "the repo is detached" is really too generic and idk what i should do. Maybe i don't recall i pull a newest commit. Or even if i want to fix it manually, idk which commit AI studio expects for that recipe. So i guess the only viable solution is to delete it myself, but it would be extremely helpful if studio offers me to do it.

axel7083:
> The detection works fine for me. My only doubt is the error message. Maybe we should provide an option that deletes and redownloads the repo automatically. When i see the error "the repo is detached" is really too generic and idk what i should do. Maybe i don't recall i pull a newest commit. Or even if i want to fix it manually, idk which commit AI studio expects for that recipe. So i guess the only viable solution is to delete it myself, but it would be extremely helpful if studio offers me to do it.

@lstocchi I am always very careful at deleting stuff that has been edited by the user. If the repository has commit AHEAD, it means he has purposefully make changes.

I think we could instead of having an option to delete the repository have inside the modal "Cancel" and "Continue"  option. This would let the user uses the modification without blocking the process.

lstocchi:
> @lstocchi I am always very careful at deleting stuff that has been edited by the user. If the repository has commit AHEAD, it means he has purposefully make changes.
> 
> I think we could instead of having an option to delete the repository have inside the modal "Cancel" and "Continue" option. This would let the user uses the modification without blocking the process.

Well, in this case the user is confirming that he agrees to delete and re-download the repo so i don't think there is any problem.

I could also agree with having the continue button but then we should somehow inform the user that the application may not work as expected. But i think we should still offers a way to reset the cloned repo.

lstocchi:
> > The modal api is **very limiting** I do not control the color of the buttons, nor the order.

Ok for me but we should add some text to explain it a bit better like `By continuing, the AI application may not run as expected. To reset its state, please delete the local folder`. Or something similar.

axel7083:
image

Thanks @lstocchi I added the sentence you mention.

> I would love that we can use markdown in the modal....

lstocchi (approved):
I still vote for having a `reset automatically` action but we can discuss it with Mo and work on it in a follow-up PR. Works fine for me. Thanks!

And i agree on the markdown

jeffmaury:
> > Looks to me weird as what would be the options to the user if the catalog is updated from 1 commit to another ?
> 
> I updated the behaviour, after _a lot of verifications_ if the repository is _clean_, and behind, the pull command is run.

That is not working for me, I have a dialog saying that the repo is detached and I have to delete the local folder

jeffmaury:
> > That is not working for me, I have a dialog saying that the repo is detached and I have to delete the local folder
> 
> @jeffmaury Behind and detached is not the same thing. Behind mean locally you have the latest commit checkout, but the remote has commit ahead. Detached is you have all the commit locally, but you purposefully checkout a previous commit, leading to a detached state

That's internal git stuff, as a user I should be able to switch from one commit to another, no ?

axel7083:
> > > That is not working for me, I have a dialog saying that the repo is detached and I have to delete the local folder
> > 
> > 
> > @jeffmaury Behind and detached is not the same thing. Behind mean locally you have the latest commit checkout, but the remote has commit ahead. Detached is you have all the commit locally, but you purposefully checkout a previous commit, leading to a detached state
> 
> That's internal git stuff, as a user I should be able to switch from one commit to another, no ?

Yes but when the repository is behind, we pull automatically, but it is detached, meaning the user has made __manual__ change, he is responsible of it.

jeffmaury:
> > > > That is not working for me, I have a dialog saying that the repo is detached and I have to delete the local folder
> > > 
> > > 
> > > @jeffmaury Behind and detached is not the same thing. Behind mean locally you have the latest commit checkout, but the remote has commit ahead. Detached is you have all the commit locally, but you purposefully checkout a previous commit, leading to a detached state
> > 
> > 
> > That's internal git stuff, as a user I should be able to switch from one commit to another, no ?
> 
> Yes but when the repository is behind, we pull automatically, but it is detached, meaning the user has made **manual** change, he is responsible of it.

This is where we diverge, you are deteched when your are on a specific commit or tag::

See the status on a repo cloned and checked out by AI Studio:

image

axel7083:
> See the status on a repo cloned and checked out by AI Studio:

I am not sure to get what you mean, have you configured the recipe to checkout a particular branch ?

jeffmaury:
> > See the status on a repo cloned and checked out by AI Studio:
> 
> @jeffmaury I am not sure to get what you mean, have you configured the recipe to checkout a particular branch ?

A specific commit (but should also be valid if checking a specific tag)

packages/backend/src/managers/gitManager.ts
```diff
@@ -33,4 +33,63 @@ export class GitManager {
...
+    }
+
+    if (status.ahead !== 0) {
+      return {
+        error: `The local repository has ${status.ahead} commit(s) ahead.`,
+      };
+    }
+
+    // Ensure the branch tracked is the one we want
+    if (ref !== undefined && status.tracking !== ref) {
```
axel7083, line 77: @jeffmaury here I am ensuring that the reference provided by the recipe catalog is matching the one expected

jeffmaury: As I see no changes, here are the scenarios that I think we should support:

- switch from one commit to another
- switch from one tag to another
- optionaly switch from one branch to a commit or tag (as we don't want to have recipes on branches as it's not stable

axel7083:
Closing in favour of https://github.com/projectatomic/ai-studio/pull/597

PR #245, opened 2024-02-08
feat: display all models in the models page

### LINKED ISSUE #43: UX: Differentiate the list of models between those available locally and those available remotely
In the list of models - at the moment we are displaying only those available locally.
It will be needed to have a design that allow to browse models from a remote registry. 

This issue is related to the UX needed to handle the display of those different models.

jeffmaury:
Fixes #43

### What does this PR do?

Display all models from the catalog in the models page. If the model is not downloaded, existing actions are disabled. Goal is to allow new actions (download,...) in the future

### Screenshot / video of UI

models

#43 

### How to test this PR?

Go to the Models page

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -167,7 +167,16 @@ describe('pullApplication', () => {
     mocks.createContainerMock.mockResolvedValue({
       id: 'id',
     });
-    modelsManager = new ModelsManager('appdir', {} as Webview, {} as CatalogManager, telemetryLogger);
+    modelsManager = new ModelsManager(
+      'appdir',
+      {} as Webview,
+      {
+        getModels(): ModelInfo[] {
+          return [];
+        },
+      } as CatalogManager,
+      telemetryLogger,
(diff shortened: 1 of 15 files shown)


feloy:
I wonder if we want to only disable the actions, or completely hide them (as they will be replaced by other actions in the future)

The title is not accurate anymore (Models on disk)

Probably not related to this PR, but the pointer should not change to a hand when a button is disabled

axel7083:
I would argue that they probably should be separated ? I don't really like having all my models mixed up. As a user I probably want to have some clear distinctions between what I have locally, and what I could have.

For now the catalog is kinda small, but in the future we might have a lot.

@mairin might have an opinion on the matter ?

feloy (approved):

mairin:
Let's drop the popularity column, @lstocchi and I talked about this before and we can't grab this data in the way we want so let's just drop for now.

@axel7083 What might be a natural way would be to have tabs like @feloy added to the containers tab, there they are All | Running | Stopped. Instead here it could be something like, All | Downloaded | Available to Download - what do you think?

Issue #238, opened 2024-02-06
recipe details page having application details partially hidden when opening IA-Studio


axel7083:
When opening Podman-Desktop with the latest version of AI-Studio I am getting the following windows

![image](https://github.com/projectatomic/ai-studio/assets/42176370/111f97c2-5836-4713-8e60-b16c11bf43bd)

> I did not resized my windows

cc @lstocchi

axel7083:
I started investigating, it comes from the <code> html element inside the markdown i will assign myself

axel7083:
Okey the problem is **annoying** the `<pre>` html element is using the `white-space: pre` which _magically_ set some kind of arbitrary width value, and the only way to prevent it is to set manually `max-width`  with a defined value but I can't make it works.

I tried making the with calculable of the markdown `max-width: calc(100% - 375px);` but without much success.

cc @feloy @lstocchi if you are motivated

lstocchi:
Just saw this on the MAC now, quite annoying and almost unusable as it doesn't scroll horizontally. I'll try to give it a look later if nobody do it earlier

lstocchi:
and thanks for the explanation @axel7083 👍

axel7083:
The only solution I thought are to make the application details position absolute, the markdown too. 

Then set the `right` css property of application details to `0px`.
For the markdown component set  `max-width: calc(100% - 375px)` however, it has a lot of issue, especially when changing tabs

Good luck @lstocchi

feloy:
Not sure if it is the same thing that Axel describes, but I would see the application details as a 'floating' window, which would overlap the markdown when it is opened, so the markdown would always have the complete width for being displayed.

axel7083:
> Not sure if it is the same thing that Axel describes, but I would see the application details as a 'floating' window, which would overlap the markdown when it is opened, so the markdown would always have the complete width for being displayed.

No I am not suggesting having a flotting island, as it would clearly hide the markdown and make it complicated to read. Imagine having a very very long first paragraph in the markdown, you could never read it, as bigger your windows will get, the paragraph will always fill all the space, and the `floating` island would always be hidden a part of it

axel7083:
I have a working solution waiting for https://github.com/projectatomic/ai-studio/pull/252 to be merged

PR #249, opened 2024-02-08
fix: adjust navPage size to make internal elements scrollable horizontally

lstocchi:
### What does this PR do?

This PR updates the width of the navPage so that inner elements are actually scrollable

### Screenshot / video of UI

with this patch
scrollable_div

before
before_scrollable_div

it is related to #238 , not sure if it actually fixes it but atleast it makes studio usable on small screen for the moment

### How to test this PR?

1. run studio and check all pages are visible as before and if you reduce its width, it is scrollable horizontally

### DIFF
--- a/packages/frontend/src/lib/NavPage.svelte
+++ b/packages/frontend/src/lib/NavPage.svelte
@@ -11,7 +11,7 @@ export let icon: IconDefinition | undefined = undefined;
 export let contentBackground = '';
 </script>
 
-<div class="flex flex-col w-full h-full shadow-pageheader">
+<div class="flex flex-col w-[calc(100%-theme(width.leftsidebar)-2px)] h-full shadow-pageheader">
   <div class="flex flex-col w-full h-full pt-4" role="region" aria-label="{title}">
     <div class="flex pb-2 px-5" role="region" aria-label="header">
         <div class="flex flex-col">


axel7083 (changes requested):
IMHO this is not what we want to display, I am not a big fan of having a horizontal scroll, I think we should find a solution to force the markdown to resize

I will check again today

axel7083:
> > IMHO this is not what we want to display, I am not a big fan of having a horizontal scroll, I think we should find a solution to force the markdown to resize
> > I will check again today
> 
> You will get at a point that you need to show a horizontal scrollbar, you cannot shrink the content forever. I expect it would have a minimun width supported.
> 
> This just fixes a problem on the navPage which does not take in consideration the left sidebar width. This is why you don't see the horizontal scrollbar in its full form (the right arrow on the bottom right is never visible).

The Podman-Desktop pages don't have horizontal scrollbars. When the content cannot get smaller, they just stop shirking it

lstocchi:
> The Podman-Desktop pages don't have horizontal scrollbars. When the content cannot get smaller, they just stop shirking it

E.g the containers page shows a scrollbar, the same the summary in the details page.

> We might want having a responsive design instead, and having the application details moving on top of the markdown

Maybe @mairin can give us guidance for this. The problem is that the markdown is not generated by us so it may have any structure

axel7083:
@lstocchi do you think we can close it ? https://github.com/projectatomic/ai-studio/pull/254 has been merged

lstocchi:
for me it still makes sense. Your PR didn't actually fix this issue.
If you shrink the page the scrollbar on the right is not visible anymore bc it goes over the width
no-scrollbar

axel7083:
I still think that a responsible design would be better, where we move on top the columns instead of scrolling.

I drafted a responsive design of _what could it be_

https://github.com/projectatomic/ai-studio/assets/42176370/f894c954-f8a2-4482-9437-a943615070ef

> I am very bad at responsive stuff we might ask @mairin for comments

lstocchi:
> I still think that a responsible design would be better, where we move on top the columns instead of scrolling.
> 
> I drafted a responsive design of _what could it be_
> 
>  responsible-recipe-page.mp4 
> > I am very bad at responsive stuff we might ask @mairin for comments

I agree with the responsive design👍it would also fix the behavior of unreadable text/image when shrinking
This PR just fixes the width of the div which is wrong in my opinion but it doesn't solve the other issues

axel7083:
Yes the responsive design is out of scope of this PR, but I would rather avoid having the horizontal scrollbar implemented, as it would open the door, to not try to make everything fit properly and just say "we can scroll"

mairin:
@axel7083 @lstocchi fwiw I agree with the approach axel mocked up with the column popping on the top - but if that has to be another PR that's cool.

PR #253, opened 2024-02-08
doc: ai-studio.yaml changes proposal

jeffmaury:
### What does this PR do?

Add a new proposal for a change

### Screenshot / video of UI

N/A

N/A

### How to test this PR?

N/A

### DIFF
--- a/docs/proposals/ai-studio.md
+++ b/docs/proposals/ai-studio.md
@@ -19,7 +19,7 @@ Define a dependency between containers: so in the definition of container, we wo
 that would define the ordering between the containers. This would be an optional field.
 
 Define a condition for the container to be properly started: this would be based on the HEALTHCHECK that can already
-be defined in a Dockerfile. In the first iteration, we would support only the ```test``` field. If ```heathcheck``` is defined,
+be defined in a Dockerfile. In the first iteration, we would support only the ```test``` field. If ```healthcheck``` is defined,
 then we would check for the healthcheck status field to be ```healthy```
 
 So the current chatbot file would be updated from:
@@ -65,14 +65,14 @@ application:
     - name: chatbot-inference-app
       contextdir: ai_applications
       containerfile: builds/Containerfile
-      depends:
-        - chatbot-model-service*
+      depends:                     # added
+        - chatbot-model-service*   # added
     - name: chatbot-model-service
       contextdir: model_services
       containerfile: base/Containerfile
       model-service: true
-      healthcheck:
-        test: curl -f localhost:7860 || exit 1
+      healthcheck:                               # added
+        test: curl -f localhost:7860 || exit 1   # added
       backend: 
         - llama
       arch:
@@ -82,8 +82,8 @@ application:
       contextdir: model_services
       containerfile: cuda/Containerfile
       model-service: true
-      healthcheck:
-        test: curl -f localhost:7860 || exit 1
+      healthcheck:                              # added
+        test: curl -f localhost:7860 || exit 1  # added
       backend: 
         - llama
       gpu-env:
--- a/docs/proposals/ai-studio.md
+++ b/docs/proposals/ai-studio.md
@@ -1,0 +1,97 @@
+# Motivation
+
+Today, there is no notion of ordering between the containers. But we know that we have a dependency between
+the client application and the container that is running the model.
+
+The second issue is that there is no concept of starting point for a container so today we rely only on the
+container being started by the container engine and we know that this is not adequate for the model service container
+
+So this is handle by a kind of dirty fix: the containers are all started in parallel but as the client application
+will fail because the model service is started (as it take a while), so we are trying to restart the client application
+until the model service is properly started.
+
+The purpose of this change is to propose an update to the ai-studio.yaml so that it is as much generic as it
+could be and inspired from the Compose specification.
+
+## Proposed changes
+
+Define a dependency between containers: so in the definition of container, we would add a new ```depends``` field
+that would define the ordering between the containers. This would be an optional field.
+
+Define a condition for the container to be properly started: this would be based on the HEALTHCHECK that can already
+be defined in a Dockerfile. In the first iteration, we would support only the ```test``` field. If ```heathcheck``` is defined,
+then we would check for the healthcheck status field to be ```healthy```
+
+So the current chatbot file would be updated from:
+
+```yaml
+application:
+  type: language
+  name: chatbot
+  description: This is a LLM chatbot application that can interact with a llamacpp model-service
+  containers:
+    - name: chatbot-inference-app
+      contextdir: ai_applications
+      containerfile: builds/Containerfile
+    - name: chatbot-model-service
+      contextdir: model_services
+      containerfile: base/Containerfile
+      model-service: true
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chatbot-model-servicecuda
+      contextdir: model_services
+      containerfile: cuda/Containerfile
+      model-service: true 
+      backend: 
+        - llama
+      gpu-env:
+        - cuda
+      arch: 
+        - amd64
+```
+
+to
+
+```yaml
+application:
+  type: language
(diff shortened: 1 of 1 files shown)


docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+      containerfile: base/Containerfile
+      model-service: true
+      healthcheck:
+        test: curl -f localhost:7860 || exit 1
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chatbot-model-service
```
feloy, line 81:
```suggestion
    - name: chatbot-model-servicecuda
```

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+application:
+  type: language
+  name: chatbot
+  description: This is a LLM chatbot application that can interact with a llamacpp model-service
+  containers:
+    - name: chatbot-inference-app
+      contextdir: ai_applications
+      containerfile: builds/Containerfile
+      depends:
+        - chatbot-model-service*
```
feloy, line 69:
```suggestion
      depends:                     # added
        - chatbot-model-service*   # added
```

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+    - name: chatbot-model-service
+      contextdir: model_services
+      containerfile: base/Containerfile
+      model-service: true
+      healthcheck:
+        test: curl -f localhost:7860 || exit 1
```
feloy, line 75:
```suggestion
      healthcheck:                               # added
        test: curl -f localhost:7860 || exit 1   # added
```

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chatbot-model-service
+      contextdir: model_services
+      containerfile: cuda/Containerfile
+      model-service: true
+      healthcheck:
+        test: curl -f localhost:7860 || exit 1
```
feloy, line 86:
```suggestion
      healthcheck:                              # added
        test: curl -f localhost:7860 || exit 1  # added
```

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+The purpose of this change is to propose an update to the ai-studio.yaml so that it is as much generic as it
+could be and inspired from the Compose specification.
+
+## Proposed changes
+
+Define a dependency between containers: so in the definition of container, we would add a new ```depends``` field
+that would define the ordering between the containers. This would be an optional field.
+
+Define a condition for the container to be properly started: this would be based on the HEALTHCHECK that can already
+be defined in a Dockerfile. In the first iteration, we would support only the ```test``` field. If ```heathcheck``` is defined,
```
feloy, line 22:
```suggestion
be defined in a Dockerfile. In the first iteration, we would support only the ```test``` field. If ```healthcheck``` is defined,
```

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+then we would check for the healthcheck status field to be ```healthy```
+
```
feloy, line 24: The ```test``` command is intended to be executed from inside the container in which it is defined.

docs/proposals/ai-studio.md
jeffmaury: I did it on purpose as they are logically the same container otherwise we should have 2 dependencies for the client application

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+will fail because the model service is started (as it take a while), so we are trying to restart the client application
+until the model service is properly started.
+
+The purpose of this change is to propose an update to the ai-studio.yaml so that it is as much generic as it
+could be and inspired from the Compose specification.
+
+## Proposed changes
+
+Define a dependency between containers: so in the definition of container, we would add a new ```depends``` field
+that would define the ordering between the containers. This would be an optional field.
```
axel7083, line 19: I do not think this is necessary, as Kubernetes works without it ?  If a container depends on a database, the container should handle a retry scenario waiting for the database to be available.

I think we should make this file closer to Kubernetes resource definition rather than inventing new concept

docs/proposals/ai-studio.md
axel7083: I also think we should simply use the official definition of Kubernetes for readiness and liveness, and use that in the pod for each contain.

See https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

jeffmaury: Yes that would be an option but we would support only the exec variant as it would be based on podman healthcheck implementation

docs/proposals/ai-studio.md
MichaelClifford: I don't mind adding the `dependency` flag here if it helps. But we could also enforce some of these checks in the application code itself (confirm there is an available model service before starting the client application). Sounds like that might be better solution?

jeffmaury: OK so the client application will wait until the model service is available and then starts its logic. So I keep only the healthcheck/readiness one

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+
+to
+
+```yaml
+application:
+  type: language
+  name: chatbot
+  description: This is a LLM chatbot application that can interact with a llamacpp model-service
+  containers:
+    - name: chatbot-inference-app
```
feloy, line 65: I would also add an healthcheck in the app container, to help AI studio check when the app is ready for the user to use it

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,97 @@
...
+      contextdir: ai_applications
+      containerfile: builds/Containerfile
+      depends:                     # added
+        - chatbot-model-service    # added
```
feloy, line 69:
```suggestion
      depends:                     # added
        - chatbot-model-service    # added
      healthcheck:                               # added
        test: curl -f localhost:8080 || exit 1   # added
```

docs/proposals/ai-studio.md
feloy: As discussed on slack, the healthcheck will also be helpful on the client container, to AI Studio knows when the complete app is ready for the user

docs/proposals/ai-studio.md
```diff
@@ -0,0 +1,98 @@
...
+      backend: 
+        - llama
+      arch:
+        - arm64
+        - amd64
+    - name: chatbot-model-service
+      contextdir: model_services
+      containerfile: cuda/Containerfile
+      model-service: true
+      healthcheck:                              # added
```
axel7083, line 86: readinessProbe?

docs/proposals/ai-studio.md
axel7083: readinessProbe instead of healthcheck

docs/proposals/ai-studio.md
MichaelClifford: Ok, so I can 1)  update the application code for the recipes to wait for a model service to become available. 2) Add a Healthcheck to the service and application Containerfiles? 

is that right? @feloy @jeffmaury ?

docs/proposals/ai-studio.md
jeffmaury: Fixed

docs/proposals/ai-studio.md
jeffmaury: 2) is to be delayed as this changes are not implemented yet. But 1) should be done now without disruption

feloy (approved):

docs/proposals/ai-studio.md
MichaelClifford: Here is the PR with the proposed change. If ok we can replicate to the other apps.  https://github.com/redhat-et/locallm/pull/41

PR #259, opened 2024-02-09
feat: recipe page responsive design

axel7083:
### What does this PR do?

Make the recipe page responsive. The Recipe/Application details will be placed on top on the `README.md` when the window size will reach a size of `1024px`. _(Ref https://tailwindcss.com/docs/responsive-design)_.

> The models tabs is a bit weird looking when the size is very small, the table itself is responsible of that behaviour, so we probably want **later** to improve the table component.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/dc2ecb57-7809-434d-83be-08f3fbaf81e4

Fixes https://github.com/projectatomic/ai-studio/issues/258

### How to test this PR?

- (1) Open any recipe
- (2) Make the window size vary

### DIFF
--- a/packages/frontend/src/lib/RecipeDetails.svelte
+++ b/packages/frontend/src/lib/RecipeDetails.svelte
@@ -44,7 +44,7 @@ const toggle = () => {
 </script>
 
 <div class="lg:my-5 max-lg:w-full max-lg:min-w-full" class:w-[375px]={open} class:min-w-[375px]={open}>
-  <div class:hidden={!open} class:block={open} class="h-fit lg:bg-charcoal-800 lg:rounded-l-md py-4 max-lg:block" aria-label="application details panel">
+  <div class:hidden={!open} class:block={open} class="h-fit lg:bg-charcoal-800 lg:rounded-l-md mt-4 py-4 max-lg:block" aria-label="application details panel">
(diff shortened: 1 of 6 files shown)


axel7083:
We might want to have @mairin opinion on the subject

lstocchi (commented):
Looks great.
I would reduce the size to start moving the items (maybe medium or small, @mairin can be more helpful) and when it's completely shrink there is a small horizontal scrollbar

responsive

axel7083:
@lstocchi  thanks for the tests! It was complicated to find, but the small horizontal bar was appearing because the text in the links in markdown where overflowing.

lstocchi (commented):
I don't see the scrollbar when switching to the models tab, even though i can actually scroll down
responsive_scrollbar

when i open/close the details, the open/close icon do not maintain the same position
open_details

axel7083:
@lstocchi I fixed the position of the toggle button. However, for the scrollbar, the issue is a bit complexe, the table has some hardcoded pixel values for the headers, and they are overflowing on the right, the scrollbar is hidden because of that.

The ideal fix would be to define `optional` column, that could be hidden when the size of the screen get too small.

For now, I cannot improve further the design nor fix the scroll bar without changing a lot the table component.

feloy (approved):
works great, thanks

Issue #177, opened 2024-01-29
Environment implementation


jeffmaury:
Tasks:

- [x] #178
- [x] #179
- [x] #180
- [x] #181
- [x] #162
- [x] #295
- [x] #286
- [x] #292
- [x] #285
- [x] #270
- [x] #328
- [x] #293
- [x] #235
- [x] #61
- [x] #265
- [x] #281
- [x] #284

slemeur:
Closing, all tasks completed

PR #261, opened 2024-02-09
[wip] Delete environment

feloy:


axel7083:
You can rebase, the linter issue has been fixed by https://github.com/projectatomic/ai-studio/pull/260

axel7083:
Would it be interesting to reuse the delete animation component used by podman desktop ? see https://github.com/containers/podman-desktop/pull/5717

packages/backend/src/managers/environmentManager.ts
```diff
@@ -131,4 +136,42 @@ export class EnvironmentManager {
...
       });
   }
+
+  async deleteEnvironment(recipeId: string) {
+    try {
+      this.setEnvironmentStatus(recipeId, 'stopping');
+      const envPod = await this.getEnvironmentPod(recipeId);
+      await containerEngine.stopPod(envPod.engineId, envPod.Id);
+      this.setEnvironmentStatus(recipeId, 'removing');
+      await containerEngine.removePod(envPod.engineId, envPod.Id);
```
jeffmaury, line 146: The object stored in #environments is not removed ?

packages/backend/src/studio-api-impl.ts
```diff
@@ -157,6 +157,34 @@ export class StudioApiImpl implements StudioAPI {
     return this.environmentManager.getEnvironmentsState();
   }
 
+  async requestRemoveEnvironment(recipeId: string): Promise<void> {
+    const recipe = this.catalogManager.getRecipeById(recipeId);
+    // Do not wait on the promise as the api would probably timeout before the user answer.
```
jeffmaury, line 162: Not sure the comment is relevant

packages/backend/src/managers/environmentManager.ts
feloy: No, it is removed when the pod is detected to be removed (on forgetPod). I prefer to centralize the state management by detecting the events on pod, and not touch it internally

feloy:
The deletion fails when the pod is not completely running. I'm working on it

packages/backend/src/studio-api-impl.ts
feloy: I think it is, to explain why we are not wait'ing for the Promise returned by showWarningMessage to be resolved, but are using `then` instead (or the request between front and back timeouts)

packages/backend/src/studio-api-impl.ts
```diff
@@ -149,6 +149,34 @@ export class StudioApiImpl implements StudioAPI {
...
+    podmanDesktopApi.window
+      .showWarningMessage(
+        `Delete the environment "${recipe.name}"? This will delete the containers running the application and model.`,
```
axel7083, line 157:
```suggestion
        `Delete the environment "${recipe.name}" ? This will delete the containers running the application and model.`,
```

feloy:
> Would it be interesting to reuse the delete animation component used by podman desktop ? see [containers/podman-desktop#5717](https://github.com/containers/podman-desktop/pull/5717)

I've added some animation by adding `inProgress` instead of `enabled`

packages/backend/src/studio-api-impl.ts
feloy: In English, no space is expected before question/exclam marks

axel7083:
When I delete the environment when the task `Starting application` is not finish

image

I am stuck in a loop of this never ending

feloy:
> When I delete the environment when the task `Starting application` is not finish
> 
> image
> 
> I am stuck in a loop of this never ending

I've added an issue in #177 (#270) to fix these inconsistent states on the recipe page

feloy:
> I restarted my podman machine after a reboot thus I had a pod and containers in a Created state but the env is Running. Is that expected ?

Yes, I created an issue to follow this: https://github.com/projectatomic/ai-studio/issues/265#issuecomment-1938405393

axel7083 (approved):
LGTM

PR #263, opened 2024-02-10
tests: mock for catalog store

feloy:
### What does this PR do?

Adds a mock for the catalog store and use it in the Recipe page

Sources: https://github.com/sveltejs/kit/discussions/9759


benoitf (commented):
hello, it looks like the mocking seems quite complex with the usage of a 3rd party class mocks/catalog

I would suggest to use simpler mocking
```
-const { mockCatalogStore } = await vi.hoisted(() => import('/@/stores/mocks/catalog'));
+import { readable, writable} from 'svelte/store';
+import * as catalogStore from'/@/stores/catalog';
```

```
 vi.mock('/@/stores/catalog', async () => {
   return {
+    catalog: vi.fn(),
   };
 });
```

and then when you want to change the catalog

you mocked the result to a readable
```
-  mockCatalogStore.mockSetSubscribeValue(initialCatalog);
+  vi.mocked(catalogStore).catalog = readable<Catalog>(initialCatalog);
```

when you need the update stuff, replace the mock by a writable mock
```
+  const customCatalog = writable<Catalog>(initialCatalog);
+  vi.mocked(catalogStore).catalog =customCatalog;
```

and then use the catalog set operation
```
-  mockCatalogStore.mockSetSubscribeValue(updatedCatalog);
+  customCatalog.set(updatedCatalog);
```

https://gist.github.com/benoitf/0b3a0c4642812c22b88f93e5e688fd25

feloy:
Effectively, it is much simpler like this, thanks

benoitf (approved):

deboer-tim (approved):

PR #274, opened 2024-02-12
feat: allowing column to be omitted when window size get too small

axel7083:
### What does this PR do?

Making column having an `optional` property, meaning they _can_ be hidden when the screen get too small.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/00f4a80d-78f4-4558-8865-58d6cab6bc95

Fixes https://github.com/projectatomic/ai-studio/issues/273

### How to test this PR?

- TODO: write unit tests


axel7083:
Closing as we still not have a single source of truth and the Table component is used by the core

Issue #265, opened 2024-02-12
Detect when "AI App" is running, by checking containers healthchecks


feloy:

axel7083:
Can we detect if the pod is stopped ?

https://github.com/projectatomic/ai-studio/assets/42176370/9c56056a-30fb-4180-bbea-75a3cd45be20

feloy:
Yes, you are raising two problems. 

First is that in the Environments page, we are displaying 'running' when the pod is started, but not completely running. As discussed in another issue, we cannot receive events when the state of the pod changes, and so we cannot get (reactively) the currenty state of the pod. That means we need to do some pulling to get the information. This issue proposes to pull the status of the application port, instead of pulling the status, as the application port is more precisely what the user is waiting for.

The second is that the status in the Recipe page is not correctly updated when the pod changes (deleted, ...). We need to work on this too

axel7083:
> Yes, you are raising two problems.
> 
> First is that in the Environments page, we are displaying 'running' when the pod is started, but not completely running. As discussed in another issue, we cannot receive events when the state of the pod changes, and so we cannot get (reactively) the current state of the pod. That means we need to do some pulling to get the information. This issue proposes to pull the status of the application port, instead of pulling the status of the pod, as the application port is more precisely what the user is waiting for.
> 
> The second is that the status in the Recipe page is not correctly updated when the pod changes (deleted, ...). We need to work on this too, using the `podmanConnection` class should help

Thanks for the details!

feloy:
@ekidneyrh This change will add a new state to the pod ued to run an AI App. 

Before, the state of the pod were either running or not running.

With this change, a pod can be running but not healthy (for example, the model container is started, but is not ready to serve connections yet). The frontend will have a new variable `healthy`, and we will want to materialise this value in the frontend.

We currently display this state in 2 places: 
- in the ai apps list, in the Status column:
  - either with `AI App is running` when the app has been started in the current Podman Desktop/AI Lab session
  - or with `Pod running` when the app has been started in a previous session of Podman Desktop/Ai Lab
- in the recipe page, in the "AI Apps Details" region:
  - the running Status of the pod is displayed (can be either RUNNING, DEGRADED, etc)

feloy:
fixed by #742

PR #276, opened 2024-02-12
[wip] Fix pulled application status

### LINKED ISSUE #270: fix pull application status on Recipe page when environment managed from Environments page

feloy:
### What does this PR do?

This PR merges the Environment manager into the Application manager. 

Fixes #270

### How to test this PR?

Play with environments (remove, restart) or pods (remove) and check if the status is corrected reflected in the recipe page.


packages/backend/src/studio-api-impl.ts
```diff
@@ -181,6 +181,32 @@ export class StudioApiImpl implements StudioAPI {
       });
   }
 
+  async requestRestartEnvironment(recipeId: string): Promise<void> {
```
axel7083, line 184: is it rebase ?

axel7083:
I tested it locally, it is working as expected

I tried 
- Stopping from the pod list ✅
- Stopping from CLI ✅

What is still _weird_ to me
- We have the `running` status in the environment page but still starting in recipe page

feloy:
> 
> What is still _weird_ to me
> 
> * We have the `running` status in the environment page but still starting in recipe page

Yes, I agree. This is because the env page does not (want to) use the state from the application manager (which is able to know the 'loading' status), which will disappear with #265. It should be fixed with #265

lstocchi:
I'm always a bit confused when i read WIP in the title as i don't get if it's ready to be tested or not. Could we use the draft option in gh to mark a PR as wip?
@feloy Is this ready to be reviewed?

feloy:
> I'm always a bit confused when i read WIP in the title as i don't get if it's ready to be tested or not. Could we use the draft option in gh to mark a PR as wip? @feloy Is this ready to be reviewed?

No, this one is not for review. I'll most probably replace it by another one with a different refactoring, based on what we discussed this morning

jeffmaury:
> > I'm always a bit confused when i read WIP in the title as i don't get if it's ready to be tested or not. Could we use the draft option in gh to mark a PR as wip? @feloy Is this ready to be reviewed?
> 
> No, this one is not for review. I'll most probably replace it by another one with a different refactoring, based on what we discussed this morning

Please use draft status in that case, I have the same concern

feloy:
closed in favor of #289

PR #280, opened 2024-02-12
feat: allow to swap model

### LINKED ISSUE #201: Add ability to select a model and run a sample application with it

lstocchi:
### What does this PR do?

This PR allows to swap models and start the application with the selected one.
Need to add some tests but it works if you want to test. On Windows the mistral model is faster than llama

### Screenshot / video of UI

swap_model

it resolves #201 

### How to test this PR?

1. select the chatbot recipe, swap model, run application and check it uses the selected model

### DIFF
--- a/packages/backend/src/studio-api-impl.spec.ts
+++ b/packages/backend/src/studio-api-impl.spec.ts
@@ -119,7 +119,7 @@ test('expect pull application to call the withProgress api method', async () => 
   mocks.withProgressMock.mockResolvedValue(undefined);
 
   await catalogManager.loadCatalog();
-  await studioApiImpl.pullApplication('recipe 1');
+  await studioApiImpl.pullApplication('recipe 1', 'model1');
   expect(mocks.withProgressMock).toHaveBeenCalledOnce();
 });
 
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -63,12 +63,10 @@ export class StudioApiImpl implements StudioAPI {
     return this.recipeStatusRegistry.getStatuses();
   }
 
-  async pullApplication(recipeId: string): Promise<void> {
+  async pullApplication(recipeId: string, modelId: string): Promise<void> {
     const recipe = this.catalogManager.getRecipes().find(recipe => recipe.id === recipeId);
     if (!recipe) throw new Error('Not found');
 
-    // the user should have selected one model, we use the first one for the moment
-    const modelId = recipe.models[0];
     const model = this.catalogManager.getModelById(modelId);
 
     // Do not wait for the pull application, run it separately
--- a/packages/frontend/src/lib/RecipeDetails.spec.ts
+++ b/packages/frontend/src/lib/RecipeDetails.spec.ts
@@ -19,27 +19,25 @@
 import '@testing-library/jest-dom/vitest';
 import { vi, test, expect, beforeEach } from 'vitest';
 import { screen, render } from '@testing-library/svelte';
-import Recipe from './Recipe.svelte';
+import userEvent from '@testing-library/user-event';
 import type { Catalog } from '@shared/src/models/ICatalog';
 import * as catalogStore from '/@/stores/catalog';
-import { readable, writable } from 'svelte/store';
+import { readable } from 'svelte/store';
+import RecipeDetails from './RecipeDetails.svelte';
+import { router } from 'tinro';
 
 const mocks = vi.hoisted(() => {
   return {
-    getCatalogMock: vi.fn(),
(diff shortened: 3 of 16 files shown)


lstocchi:
@mairin i was not sure about how to swap model. So, when the user clicks on the swap button, it opens the models tab and the swap button is hidden, then the user can click the icon in the row (model) to use and the item in the details panel is updated. WDYT? (an example in the gif above)

feloy (approved):
works great for me, on mac

axel7083:
I am not a huge fan of the big purple button which redirect to the models page. Some random ideas:

### **List icon**

Since the models is redirecting to the models list

image

### **Open icon**

The action is kinda goto open page that opens the page of models

image

mairin:
@lstocchi this is from the original model swap list mockup where the model swap was on the front page instead of a different tab:

image

I was imagining the swap model button being like that and being visible on hover. But in hindsight, I'm thinking that might not but fully accessible and maybe a bit awkward (you have to have a hover on the row then navigate to the button without leaving the row which is an awkward targeting task.) So now I'm thinking maybe we have a large "Swap Models" button same as in the mockup, but have it in the upper right corner of the Models tab (same as we do with buttons for say prune / create / kubeplay on the containers screen). And each row item in the list has a radio button? So you click the radio on the model you want to swap to and hit the "Swap Models" button to have it swap?  
 
Maybe the swap button is a little too much. I actually really like @axel7083 's suggestion to have the list all compatible models and make it a list, and keep the text message below the model explaining that you can swap for a different compatible model, but I would make "swap for a different compatible model" an underlined link to the models tab as well. I would also have a corresponding note in the same spot for non-default models eg "The default model for this recipe is $NAME. You can swap for $NAME or a different compatible model."

Handling model swapping while things are running - I guess this depends on the implementation because it's my understanding when we were using localAI this could be done on the fly, is that not the case anymore? I think depending on how we can handle it technically, that might determine the best way to handle UI-side right now.

So what do you think then about this:

## Models tab
- [ ] (1) In the model tab, a purple icon + "Swap Models" button will be in the upper right. Each row containing a model in the list below will have a radio button. Users can click on a radio button for the models that aren't currently running, and press the "Swap Models" button, and that will swap the model.
- [ ] (2) We should mark which model in the list of models is currently configured, which I missed in the design. (Whoops!) We do have the right hand sidebar which indicates this but it would be nice to reinforce in the list. Can we maybe add a "status" column (maybe replace the "actions" column with "status"?) and for the currently selected model have it say "Selected" maybe?
- [ ] (3) We should also probably mark, in the list, which model is the default for the recipe. The mockups do this via metadata on the screen like this but we can't rely on that being there for the recipe providing markdown nor be able to render it like this I think:
image So maybe in the "Status" column we can put, "Recommended" and in the case where the recommended is selected too, just put, "Selected, Recommended" or something in the Status column?

## Right Sidebar
- [ ] (4) For default models, have the "swap for a different compatible model" an underlined link to the models tab.
- [ ] (5) For nondefault models, have similar text underneath but it would read "The default model for this recipe is $NAME. You can swap for $NAME or a different compatible model." and have the "swap for..." part underlined and link to the models tab.
- [ ] (6) Drop the fat swap button. Replace as @axel7083 mocked up with the list all compatible models icon + tooltip: image

OK and then let's talk about any technical constraints / limitations around hot-swapping models on a running app?

lstocchi:
Thanks Mo, i love this. I'm going to add your suggestions.
Just one doubt about the first point

> ## Models tab
> * [ ]  (1) In the model tab, a purple icon + "Swap Models" button will be in the upper right. Each row containing a model in the list below will have a radio button. Users can click on a radio button for the models that aren't currently running, and press the "Swap Models" button, and that will swap the model.

Don't you think it is enough to just click the radio button? The button on the upper right makes sense if there is a multi-selection to do but in this case it's limited to one item. If one clicks on the radio button but see the "Selected" status on the old row, it would be weird no? Maybe we can update the row status at radio button selection and remove the button on the upper right. WDYT?

axel7083:
> image

I really like this design!! 🎉

lstocchi:
Current UI
model_table_2

axel7083 (approved):
LGTM! Nice work

mairin:
@lstocchi Looks great - I actually like the star on the left better than on the right (the label for recommended column is really long.) Can you swap back to that? I agree with all your other points, the insta radio selection sounds good to me if it doesn't cause any technical issues!

lstocchi:
> @lstocchi Looks great - I actually like the star on the left better than on the right (the label for recommended column is really long.) Can you swap back to that? I agree with all your other points, the insta radio selection sounds good to me if it doesn't cause any technical issues!

Done 👍 
image

PR #289, opened 2024-02-13
Duplicate environments actions on the recipe page

feloy:
### What does this PR do?

This is a first PR on the road to refactoring the status of the recipe/application/environment.

- It duplicates the actions buttons visible in the Environments page, so we can also restart or remove an environment/app from the recipe page
- the status in the recipe page is updated as much as possible at this point (the implementation will change in future PRs, so please don't look too much at edge cases)
- in status in the Environment page is not correct when the app is reloading (it is marked as `running` insteadof `loading`). This will also be worked on following PRs.
- the design is not well crafted for the buttons on the recipe page, this will be part of a following PR, when the different possible states are clearer


jeffmaury:
If the application is running and you click Delete, the button switches to Run application but I don"t think it should be enabled
Also if the application is running the label 'This will clone the repo' is displayed

jeffmaury:
Also while the application is Loading, I can click on Delete but after the delete has been processed, Loading is is displayed indefenitively. Maybe part of a follow up PR

feloy:
@jeffmaury  I tried to fix these inconsistencies, but I cannot find any good solution before refactoring the statuses.

lstocchi (commented):
IMO the delete action should be visible when the environment is stopped. The restart when it is actually started. Similar to what we do for podman machines. 
I would like to see the stop action tbh when it's loading. So i can stop it if i changed my mind.

feloy:
> IMO the delete action should be visible when the environment is stopped. The restart when it is actually started. Similar to what we do for podman machines. I would like to see the stop action tbh when it's loading. So i can stop it if i changed my mind.

I have created an issue to track these problems, starting with the mock made by Mairin: https://github.com/projectatomic/ai-studio/issues/293

At this stage, I cannot do anything about this, before having a better status handling

lstocchi:
So, in my opinion if the actions need to be shown enabled/disabled/hidden/whatever when the environment has a specific state I would prefer to have the status handling done first. I find it confusing otherwise to see those actions in the recipe page. If you need to merge this PR to progress, we can just remove the code that shows the buttons to the user and work on them in further PR when we have a better status handling. My 2cent

feloy:
> So, in my opinion if the actions need to be shown enabled/disabled/hidden/whatever when the environment has a specific state I would prefer to have the status handling done first. I find it confusing otherwise to see those actions in the recipe page. If you need to merge this PR to progress, we can just remove the code that shows the buttons to the user and work on them in further PR when we have a better status handling. My 2cent

OK, I'll continue working on this PR. Moving it to draft

axel7083:
Why not keeping the old `Start application` button when there is not pod ? instead of having` (no pod running) delete (disabled) restart (disabled)`

packages/backend/src/managers/applicationManager.ts
```diff
@@ -64,19 +74,28 @@ export interface ImageInfo {
 }
 
 export class ApplicationManager {
+  #environments: Map<string, EnvironmentState>;
```
axel7083, line 77:
```suggestion
  // Map recipeId => EnvironmentState
  #environments: Map<string, EnvironmentState>;
```

packages/frontend/src/lib/RecipeDetails.svelte
```diff
@@ -53,27 +49,55 @@ const toggle = () => {
...
+          <div class="shrink-0">
+            <EnvironmentControls recipeId={recipeId} object={envState} tasks={recipeStatus?.tasks} />
+          </div>  
         </div>
         {#if recipeStatus !== undefined && recipeStatus.tasks.length > 0}
           <div class="mt-4 text-sm font-normal py-2">
             <TasksProgress tasks="{recipeStatus.tasks}" />
           </div>
         {/if}
+        <!--div class="flex flex-row">
```
axel7083, line 67: remove comments ?

packages/frontend/src/lib/table/environment/ColumnStatus.svelte
```diff
@@ -1,8 +1,9 @@
...
-import type { EnvironmentState } from '@shared/src/models/IEnvironmentState';
-export let object: EnvironmentState;
+import TasksProgress from '../../progress/TasksProgress.svelte';
+import type { EnvironmentCell } from '/@/pages/environments';
+export let object: EnvironmentCell;
 </script>
 
 <div class="text-sm text-gray-700">
-  {object.status}
+  <TasksProgress mini={true} tasks="{object.tasks}" />
```
axel7083, line 8: I would rather have a separate component, responsible of taking the last tasks rather than having a lot of properties to the TasksProgress.

I would rather you only use a TaskElement inside the `Column status` rather than making the TasksProgress too complex

packages/shared/src/models/IEnvironmentState.ts
```diff
@@ -18,7 +18,7 @@
 
 import type { PodInfo } from '@podman-desktop/api';
 
-export type EnvironmentStatus = 'none' | 'running' | 'starting' | 'stopping' | 'removing' | 'error' | 'unknown';
+export type EnvironmentStatus = 'none' | 'running' | 'starting' | 'error' | 'unknown';
```
axel7083, line 21: Why removing the stopping / removing but keeping the starting ? Wouldn't it be better to have instead a `loading` state ? or `transitioning` or else

axel7083 (changes requested):

packages/frontend/src/lib/progress/TaskElement.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
```
axel7083, line 1: This could have a been a separate PR

feloy:
> (1) Why not keeping the old `Start application` button when there is not pod ? instead of having` (no pod running) delete (disabled) restart (disabled)`
> 

I'm using this mockup: https://github.com/projectatomic/ai-studio/issues/293, and have something similar to the podman machine management.

> (2) When restarting the environment from the Environment page, I have the `There is not environment` for a few seconds before it re-appear, is it because the pod is removed and re-created ?

Yes, right

> 
> But everything is working very nicely, I really like the improvements on state management. Well done 🎉 !!!

Thanks 🙏

As said in the video, I still need to manage edge cases (when pod is deleted manually, extension restarted, etc)

packages/frontend/src/lib/RecipeDetails.svelte
feloy: yes, I'll do some cleanup when I'm sure I don't need anything from the commented code

packages/frontend/src/lib/table/environment/ColumnStatus.svelte
feloy: Yes, I was wondering what would be the best solution. I'll change to have a separate component

axel7083:
> > (1) Why not keeping the old `Start application` button when there is not pod ? instead of having` (no pod running) delete (disabled) restart (disabled)`
> 
> I'm using this mockup: #293, and have something similar to the podman machine management.

Yes but I think it would be a nice touch keeping the purple button when nothing as never been started, as it catch the eyes

packages/shared/src/models/IEnvironmentState.ts
feloy: I still need to make some cleanup, I think the 'starting' will also be removed.

This is the state of the pod, I'm not sure yet what to do with it, and it is really necessary. I'll need to continue working on it

feloy:
> > > (1) Why not keeping the old `Start application` button when there is not pod ? instead of having` (no pod running) delete (disabled) restart (disabled)`
> > 
> > 
> > I'm using this mockup: #293, and have something similar to the podman machine management.
> 
> Yes but I think it would be a nice touch keeping the purple button when nothing as never been started, as it catch the eyes
> 
> Do you also plan on adding the pod icon on the right ? (maybe in another PR ?)

Yes, I'm focusing first on this PR on the state management, I'll finalize the UX when the state management is stable

packages/frontend/src/lib/progress/TaskElement.svelte
feloy: I have created a PR for this: https://github.com/projectatomic/ai-studio/pull/296

packages/frontend/src/lib/table/environment/ColumnStatus.svelte
```diff
@@ -1,8 +1,19 @@
...
+  }
+}
 </script>
 
-<div class="text-sm text-gray-700">
-  {object.status}
-</div>
+{#if task}
+  <div class="text-sm text-gray-700">
+    <TaskItem task="{task}" />
```
axel7083, line 17: Nice!

packages/frontend/src/lib/table/environment/ColumnActions.svelte
```diff
@@ -1,28 +1,7 @@
...
 </script>
 
-{#if object.status === 'stopping' || object.status === 'removing'}
-  <div class="pr-4"><Spinner size="1.4em" /></div>
-{:else}
-  <ListItemButtonIcon icon="{faTrash}" onClick="{() => deleteEnvironment()}" title="Delete Environment" />
-
-  <ListItemButtonIcon icon="{faRotateForward}" onClick="{() => restartEnvironment()}" title="Restart Environment" />
-{/if}
+<EnvironmentControls recipeId="{object.recipeId}" object="{object.envState}" tasks="{object.tasks}" />
```
axel7083, line 7: Why controls ? why not using the wording `action` ?

packages/backend/src/managers/applicationManager.ts
```diff
@@ -550,4 +569,190 @@ export class ApplicationManager {
     // Update task
     taskUtil.setTask(checkoutTask);
   }
+
+  /*
+   *
+   * Pod status
```
axel7083, line 575: ?

axel7083 (approved):
Amazing job 👏 ! 

The restart action is working well either on the application details than on the environment side. 
Deletion action is also working fine on both side. I like having the option to delete the pods before it is started!

packages/backend/src/managers/applicationManager.ts
feloy: This was to separate the old content of applicationManager and the content of environmentManager which has been added, but it will probably not be true in a few changes to the file, I will remove it

packages/frontend/src/lib/table/environment/ColumnActions.svelte
feloy: right, I'll change to EnvironmentActions

Issue #150, opened 2024-01-26
Extension not working at all if there is a problem with a model


lstocchi:
i started desktop, when to ai-studio and saw the catalog completely empty. I checked the logs and i saw 
```
Trying to call on an unknown channel getCatalog.
Error: channel does not exists
Trying to call on an unknown channel getPullingStatuses.
Error: channel does not exists
....
```

I went to the extensions page to disable the extension and saw 
```
Extension error: EPERM: operation not permitted, stat 'C:\Users\baldr\podman-desktop\ai-studio\models\llama-2-7b-chat.Q5_K_S\llama-2-7b-chat.Q5_K_S.gguf'
Stack trace
Error: EPERM: operation not permitted, stat 'C:\Users\baldr\podman-desktop\ai-studio\models\llama-2-7b-chat.Q5_K_S\llama-2-7b-chat.Q5_K_S.gguf' at statSync (node:fs:1690:3) at Object.statSync (node:electron/js2c/asar_bundle:2:4807) at ModelsManager.getLocalModelsFromDisk (C:\Users\baldr\Work\github.com\containers\studio-extension\packages\backend\dist\extension.js:8850:23) at reloadLocalModels (C:\Users\baldr\Work\github.com\containers\studio-extension\packages\backend\dist\extension.js:8820:12) at ModelsManager.loadLocalModels (C:\Users\baldr\Work\github.com\containers\studio-extension\packages\backend\dist\extension.js:8831:11) at Studio.activate (C:\Users\baldr\Work\github.com\containers\studio-extension\packages\backend\dist\extension.js:8945:30) at async activate (C:\Users\baldr\Work\github.com\containers\studio-extension\packages\backend\dist\extension.js:8964:3) at async ExtensionLoader.activateExtension (C:\Users\baldr\Work\github.com\containers\podman-desktop\packages\main\dist\index.cjs:35260:9) at async ExtensionLoader.loadExtension (C:\Users\baldr\Work\github.com\containers\podman-desktop\packages\main\dist\index.cjs:34723:7) at async Promise.all (index 8)
```

axel7083:
Fixed by https://github.com/projectatomic/ai-studio/pull/304

PR #304, opened 2024-02-14
fix: loadLocalModels issue when models being used

axel7083:
### What does this PR do?

Wrapping the `fs.statSync` in a try catch to avoid the `Error: EPERM: operation not permitted` issue appearing when the models is being used (mounted).

This method was used to check the size and the creation date of the file, those field are made optional by this PR, as we still want to show the modes is on disk, but with less information, as we cannot get those.

Fixes https://github.com/projectatomic/ai-studio/issues/303

### How to test this PR?


lstocchi (approved):
LGTM
this should also fix #150

PR #333, opened 2024-02-14
feat: renaming route paths

axel7083:
### What does this PR do?

Tinro is a bit annoying with certain scenario. When working on https://github.com/projectatomic/ai-studio/issues/332 I had the issue of having multiple route resolved when I was not suppose too.

To avoid certain issue in the future, I renamed the following routes 
- model details from `/models/:id/*` to `/model/:id/*`
- recipe details from `/recipes/:id/*` to `/recipe/:id/*`

### How to test this PR?

- Explore the application, click on tables to ensure no 404

### DIFF
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -47,11 +47,6 @@ onMount(() => {
         <Environments />
       </Route>
 
-      <!-- Models -->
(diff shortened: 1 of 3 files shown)


jeffmaury (approved):
Does not sound logical to me but if it solve issues

PR #335, opened 2024-02-15
support extension restart and manual deletion of pod

### LINKED ISSUE #328: support deleting pod manually and restarting extension for application status

feloy:
### What does this PR do?

- application being deployed and extension restarted: because AI Studio manages the restart of the app container, the pod is not ok, it needs to be deleted by user => This deletion is correctly supported and the Play button is enabled again. 
- application deployed and correctly running, and extension restarted => the status "Pod running" is displayed in the Environments page and the pod name is displayed in the "Application Details" block. 
- application being deployed or correctly running, and the pod is deleted manually => tasks cleared and replaced by a single 'Pod deleted manually', visible in "Application Details" (not in Environments, as the line has been deleted)

- replaced delete with stop button
- matching pod is deleted before to start a new one
- handle pod stopped manually
- handle machine stopped manually

Fixes #328

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -231,6 +231,7 @@ describe('pullApplication', () => {
     mockForPullApplication({
       recipeFolderExists: false,
     });
+    mocks.listPodsMock.mockResolvedValue([]);
     vi.spyOn(modelsManager, 'isModelOnDisk').mockReturnValue(false);
     doDownloadModelWrapperSpy.mockResolvedValue('path');
     const recipe: Recipe = {
@@ -297,6 +298,7 @@ describe('pullApplication', () => {
     mockForPullApplication({
       recipeFolderExists: true,
     });
+    mocks.listPodsMock.mockResolvedValue([]);
     vi.spyOn(modelsManager, 'isModelOnDisk').mockReturnValue(false);
     vi.spyOn(modelsManager, 'doDownloadModelWrapper').mockResolvedValue('path');
     const recipe: Recipe = {
@@ -325,6 +327,7 @@ describe('pullApplication', () => {
     mockForPullApplication({
       recipeFolderExists: true,
     });
+    mocks.listPodsMock.mockResolvedValue([]);
     vi.spyOn(modelsManager, 'isModelOnDisk').mockReturnValue(true);
     vi.spyOn(modelsManager, 'getLocalModelPath').mockReturnValue('path');
     const recipe: Recipe = {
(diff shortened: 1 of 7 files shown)


lstocchi (commented):
I'm still a bit puzzled about the buttons and their behaviors TBH (i would expect the same behaviors we have for podman machines - the stop and restart button to be enabled when the env is running, or the delete when it's stopped and so on.. ).

@mairin It would be great to know how this should work bc otherwise it's hard to make an objective review and it's just personal taste

My thoughts for the moment as i don't know how it should effectively work:

1) i have a recipe pod created yesterday, i open my ai-studio and i see this. I would expect to being able to click the start button and have the restart disabled.

image


2) I click on restart. Everything is built again correctly but eventually i see. Shouldn't the different states/run application summary be hidden?

image

3) I stop the pod manually and i see. I would expect to see the name of the pod without the old states/summary 

image

4) i delete the pod and i see. Shouldn't we show again the Run application button?

image

lstocchi (commented):
When moving the mouse over a disabled action the cursor remain the same 
cursor_same

feloy:
> I'm still a bit puzzled about the buttons and their behaviors TBH (i would expect the same behaviors we have for podman machines - the stop and restart button to be enabled when the env is running, or the delete when it's stopped and so on.. ).
> 
> @mairin It would be great to know how this should work bc otherwise it's hard to make an objective review and it's just personal taste
> 
> My thoughts for the moment as i don't know how it should effectively work:
> 
> 1. i have a recipe pod created yesterday (it is stopped now), i open my ai-studio and i see this. I would expect to being able to click the start button and have the restart disabled.
> 
> image
> 

For now, because the extension is managing the state of the containers in the pod (it restarts the app container when the model is running), it is not interesting to just restart a stopped pod, because it will never restart correctly. The only way for now is to delete and recreate the pod (with the Restart button)

Bu I agree that when the pod is stopped, the display should reflect it better

> 2. I click on restart. Everything is built again correctly but eventually i see. Shouldn't the different states/run application summary be hidden?
> 
> image
> 

> 3. I stop the pod manually and i see. I would expect to see the name of the pod without the old states/summary
> 
> image
> 

Right, I need to work on the Stop status of a pod

> 4. i delete the pod and i see. Shouldn't we show again the Run application button?
> 
> image

That's an open question we started discuss on https://github.com/projectatomic/ai-studio/pull/289#issuecomment-1943346809

lstocchi (approved):
LGTM
You should also update the message when stopping the env. It says `Delete the environment "${recipe.name}"?`. It should say `Stop ...`.

feloy:
> LGTM You should also update the message when stopping the env. It says `Delete the environment "${recipe.name}"?`. It should say `Stop ...`.

Good catch, thanks

axel7083 (approved):
LGTM 👏 !!

tested and works fine when 
- still starting and deleting from cli
- still starting and deleting from podman

PR #342, opened 2024-02-15
feat: dispose extension resources on deactivate 

axel7083:
### What does this PR do?

Making some Managers implements Disposable so they can be registered to call `dispose()` when the extension is disabled

Fixes https://github.com/projectatomic/ai-studio/issues/341

### How to test this PR?


lstocchi (approved):
It would be good to add some tests

PR #344, opened 2024-02-15
feature: adding tasks store for models page

axel7083:
### What does this PR do?

Instead of gettings the tasks from the RecipeState stores, we fetch them directly from a new Tasks store. 

Part of https://github.com/projectatomic/ai-studio/issues/339

> This is necessary as we want to be able to download models outside of a recipe (for the playground, from the tables etc.)

Fixes https://github.com/projectatomic/ai-studio/issues/343

### How to test this PR?

- Previous tests ensure no regression

### DIFF
--- a/packages/frontend/src/pages/Models.svelte
+++ b/packages/frontend/src/pages/Models.svelte
@@ -50,7 +50,6 @@ function filterModels(): void {
 onMount(() => {
   // Subscribe to the tasks store
   const tasksUnsubscribe = tasks.subscribe(value => {
-    console.log('tasks store', value);
     pullingTasks = value.filter(task => task.state === 'loading');
     loading = false;
     filterModels();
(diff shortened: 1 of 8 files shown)


packages/frontend/src/pages/Models.svelte
```diff
@@ -48,10 +48,10 @@ function filterModels(): void {
 }
 
 onMount(() => {
-  // Pulling update
-  const modelsPullingUnsubscribe = modelsPulling.subscribe(runningTasks => {
-    // Only display error | loading tasks.
-    tasks = runningTasks.filter(task => task.state !== 'success');
+  // Subscribe to the tasks store
+  const tasksUnsubscribe = tasks.subscribe(value => {
+    console.log('tasks store', value);
```
lstocchi, line 53: to be removed?

lstocchi (approved):
LGTM

packages/frontend/src/pages/Models.svelte
axel7083: yes nice catch!

PR #347, opened 2024-02-15
feat: adding vscode button

axel7083:
### What does this PR do?

Adding a `VS Code Desktop` button.

> See https://code.visualstudio.com/brand#brand-legal for the name of button used

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/7b63d7b8-8167-4d60-bbb3-990f3e05af9c

Fixes https://github.com/projectatomic/ai-studio/issues/27

### How to test this PR?

- [x] Unit test has been added

**manually**
(1) Open a recipe
(2) Start application
(3) See the `Open in VS Code Desktop` button

### DIFF
--- a/packages/frontend/src/lib/RecipeDetails.svelte
+++ b/packages/frontend/src/lib/RecipeDetails.svelte
@@ -12,6 +12,8 @@ import { environmentStates } from '../stores/environment-states';
 import type { EnvironmentState } from '@shared/src/models/IEnvironmentState';
 import EnvironmentActions from './EnvironmentActions.svelte';
 import Button from './button/Button.svelte';
+import VSCodeIcon from '/@/lib/images/VSCodeIcon.svelte';
+import { localRepositories } from '/@/stores/localRepositories';
 
 export let recipeId: string;
 export let modelId: string;
@@ -20,6 +22,9 @@ $: envState = $environmentStates.find((env: EnvironmentState) => env.recipeId ==
 $: recipe = $catalog.recipes.find(r => r.id === recipeId);
 $: recipeStatus = $recipes.get(recipeId);
 $: model = $catalog.models.find(m => m.id === modelId);
+$: localPath = $localRepositories.find(
+  local => !!local.labels && 'recipe-id' in local.labels && local.labels['recipe-id'] === recipeId,
+);
 
 let open: boolean = true;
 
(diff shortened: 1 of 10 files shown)


axel7083:
cc @slemeur

lstocchi (commented):
Looks great, GJ!! 

just one doubt. Shouldn't we show the button if we detect the cloned repo?

I run the chatbot, i clicked on vscode and it opened the source finely. But my podman crashed bc i was running `yarn watch` and when clicking the vscode button it replaced the vscode window with yarn watch and the desktop process was killed.... No problem as this won't happen in prod, but when reopening ai-studio and the recipe page i didn't see the vscode button again. Then my question, should we always show it when we detect the cloned repo? It may happen that i edit the code on day 1, then i come back after 1 week and i don't see the button and it's weird.

axel7083:
> Looks great, GJ!!
> 
> just one doubt. Shouldn't we show the button if we detect the cloned repo?
> 
> I run the chatbot, i clicked on vscode and it opened the source finely. But my podman crashed bc i was running `yarn watch` and when clicking the vscode button it replaced the vscode window with yarn watch and the desktop process was killed.... No problem as this won't happen in prod, but when reopening ai-studio and the recipe page i didn't see the vscode button again. Then my question, should we always show it when we detect the cloned repo? It may happen that i edit the code on day 1, then i come back after 1 week and i don't see the button and it's weird.

For now, this PR only is plugged to the LocalRepositoryRegistry, which is for now not persistent, we need to make it persistent as we want to support repository provided by the user, so when we will have this persistence system, the button will automatically be shown.

> The persistence system would only be for the external repository, we could today already list the repositories in the ia-studio folder

packages/frontend/src/lib/RecipeDetails.svelte
```diff
@@ -20,6 +22,9 @@ $: envState = $environmentStates.find((env: EnvironmentState) => env.recipeId ==
 $: recipe = $catalog.recipes.find(r => r.id === recipeId);
 $: recipeStatus = $recipes.get(recipeId);
 $: model = $catalog.models.find(m => m.id === modelId);
+$: localPath = $localRepositories.find(
+  local => !!local.labels && 'recipe-id' in local.labels && local.labels['recipe-id'] === recipeId,
```
jeffmaury, line 26: Such logic should be in the store or utility to be reused

axel7083: Good idea, I will create an utility class

axel7083:
> > The persistence system would only be for the external repository, we could today already list the repositories in the ia-studio folder
> 
> Agree, but i didn't get if you want to show the button when we see the cloned repo in the ai-studio folder or just when the user runs it

when the repo is cloned we should see the button

axel7083:
rebased

jeffmaury (approved):
LGTM

PR #363, opened 2024-02-19
Environments depends on recipe and model

### LINKED ISSUE #281: environmentManager should consider environment as (recipe, model), not only (recipe)

feloy:
### What does this PR do?

Index the environments (applications) by (recipeId, modelId) instead of recipeId only.

Fixes #281 

### How to test this PR?

Start, stop, restart applications from environments page and recipe page. Should work as before

### DIFF
--- a/packages/shared/src/models/IRecipeModelIndex.ts
+++ b/packages/shared/src/models/IRecipeModelIndex.ts
@@ -1,3 +1,21 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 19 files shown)


packages/shared/src/models/IRecipeModelIndex.ts
```diff
@@ -0,0 +1,4 @@
+export interface RecipeModelIndex {
```
axel7083, line 1: What is the purpose of this interface ?

axel7083, line 1: copyrights ?

axel7083 (approved):
LGTM

packages/shared/src/models/IRecipeModelIndex.ts
feloy: Represents the type of the key of the `ApplicationRegistry` class:

```
export class ApplicationRegistry<T extends RecipeModelIndex> {
```

Issue #262, opened 2024-02-09
add & update sample applications from locallm repository


sallyom:
* [This chatbot application](https://github.com/redhat-et/locallm/blob/main/chatbot-langchain/ai-studio.yaml) should replace the current chatbot sample
  * eventually the [older chatbot application](https://github.com/redhat-et/locallm/blob/main/chatbot/ai-studio.yaml) will be removed from locallm 
* [sample RAG application](https://github.com/redhat-et/locallm/blob/main/rag-langchain/ai-studio.yaml)

slemeur:
Part of this issue has been addressed previously - following on the RAG sample here: https://github.com/projectatomic/ai-studio/issues/474

PR #370, opened 2024-02-19
fix: consider app running if containers are running

lstocchi:
### What does this PR do?

This patch changes how the app is considered running. 
Bc of the proposals made by Jeff, we are moving to cloud native built applications that should restart themself if the model service is not ready https://github.com/redhat-et/locallm/pull/41 and we are goingto use the readiness probe to check if a container is running succesfully.
So, if we do not find any probe in the ai-studio (as it is now, we still have to add the health check), if all containers in a pod are running, the app is considered ready.

### Screenshot / video of UI

N/A

it is part of the catalog update to use the new streamlit apps #262 

### How to test this PR?

1. run the chatbot app, you should see the app is running when both containers are actually running. If you open the browser, you will see the sample app ready. If you ask a question it will fail as the model service is not ready. It will be fixed after https://github.com/redhat-et/locallm/pull/41 is merged and the catalog updated

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -186,14 +186,16 @@ export class ApplicationManager {
   }
 
   async waitContainerIsRunning(engineId: string, container: ContainerAttachedInfo): Promise<void> {
-    const sampleAppContainerInspectInfo = await containerEngine.inspectContainer(engineId, container.name);
-    if (sampleAppContainerInspectInfo.State.Running) {
-      return;
+    const TIME_FRAME_MS = 5000;
+    const MAX_ATTEMPTS = 60 * (60000 / TIME_FRAME_MS); // try for 1 hour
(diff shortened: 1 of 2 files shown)


packages/backend/src/managers/applicationManager.ts
```diff
@@ -202,25 +185,15 @@ export class ApplicationManager {
...
     await timeout(5000);
-    await this.restartContainerWhenModelServiceIsUp(engineId, modelServiceEndpoint, container).catch(
-      (error: unknown) => {
-        console.error('Error monitoring endpoint', error);
-      },
-    );
+    await this.waitContainerIsRunning(engineId, container).catch((error: unknown) => {
+      console.error('Error monitoring container', error);
+    });
   }
```
feloy, line 197: If the app takes time to start, the recurse call it will create a high call stack. A for-loop would be more appropriate IMO

feloy (approved):

PR #372, opened 2024-02-19
feat: improving recipe details

axel7083:
### What does this PR do?

Implement the design for the actions on the Application Details page. It adds the Pod icon which is used to redirect the user to the pod page when clicked on.

> The play, delete and restart are fully hidden when not usable, we have a fairly limited width, so instead of disabling them we hide them completely to make some space.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/dc3e2b96-69a6-4437-924e-28334fc069d3

> There is a bug when using the navigation api for pods, where we are not redirected to a tab, see https://github.com/containers/podman-desktop/issues/6071

Fixes https://github.com/projectatomic/ai-studio/issues/293

### How to test this PR?

- [x] Write unit tests


feloy (approved):
great, thanks

Issue #61, opened 2024-01-16
Add ability to open the sample application once started


slemeur:

axel7083:
@slemeur can we have more details on how the extension will know _What is the application ?_

I think we need to add this insight inside the `ia-studio.yaml`. We add a tag/labels explicitly stating it the application (frontend) run on this container and expose this/those port.

If we agree, then this story depends on https://github.com/projectatomic/studio-extension/issues/59 which should include those informations (cc @lstocchi )

slemeur:
linking to #74

slemeur:
Agree, that's maybe a metadata that would be nice to be provided with `ai-studio.yaml` - or the open app just list the expose ports and the user clicks to open it.

lstocchi:
Can't we rely on the Containerfile expose directive? 🤔

feloy:
blocked by #356

PR #381, opened 2024-02-20
feat: add labels with model and app ports

feloy:
### What does this PR do?

Adds labels to the application's pod giving the ports of the model and app, so the Environments manager can know them.
 

As part of #61 

### How to test this PR?

Start an application with ai-studio.yaml containing ports information for model and/or app, and check the labels exist in the pod

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -37,7 +37,7 @@ import { getPortsInfo } from '../utils/ports';
 import { goarch } from '../utils/arch';
 import { getDurationSecondsSince, timeout } from '../utils/utils';
 import type { LocalRepositoryRegistry } from '../registries/LocalRepositoryRegistry';
-import { LABEL_MODEL_ID } from './playground';
+import { LABEL_MODEL_ID, LABEL_MODEL_PORT } from './playground';
 import type { EnvironmentState } from '@shared/src/models/IEnvironmentState';
 import type { PodmanConnection } from './podmanConnection';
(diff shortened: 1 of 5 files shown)


packages/backend/src/managers/applicationManager.ts
```diff
@@ -345,13 +346,22 @@ export class ApplicationManager {
     }
 
     // create new pod
+    const labels = {
+      [LABEL_RECIPE_ID]: recipe.id,
+      [LABEL_MODEL_ID]: model.id,
+    };
+    const modelPorts = images.filter(img => img.modelService).flatMap(img => img.ports);
```
lstocchi, line 353: Any reason why you also add the model port? As we do not ping the model anymore to see if it's running we should not expose it in the pod, no?

feloy: In the design made by Mo (https://github.com/projectatomic/ai-studio/issues/235), the port of the model is displayed. I can imagine the developer may want to know the model's port, to make some manual requests on it, for example

lstocchi: maybe let's ask @jeffmaury . BTW looks good to me for the rest 👍

lstocchi (approved):
Tested, works fine!!

jeffmaury:
2 remarks:

- if there are several ports we keep only the first one, we should store all of them and them ask the user to choose
- what is the added value of having them from labels rather that info from pod ?

jeffmaury (approved):
LGTM

PR #388, opened 2024-02-20
feat: reuse existing model downloading tasks

axel7083:
### What does this PR do?

When we start a model download in the models Table, we have an issue when starting the application, where it detects the models already being on disk (in reality it is **partially** on disk) and it skip the step, when it should not.

Now, when requesting to download a model, the system will check for existing downloader and tasks. Avoiding potential errors.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/4d068def-8323-40bb-a905-dc52a5d03e3a

Fixes https://github.com/projectatomic/ai-studio/issues/384

### How to test this PR?

(1) Go to models table 
(2) Download a model
(3) Go to any recipe that can use this model
(4) Click on "Start application"

### DIFF
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -24,12 +24,7 @@ import { MSG_NEW_MODELS_STATE } from '@shared/Messages';
 import type { CatalogManager } from './catalogManager';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import * as podmanDesktopApi from '@podman-desktop/api';
-import {
-  Downloader,
-  type DownloadEvent,
-  isCompletionEvent,
-  isProgressEvent,
-} from '../utils/downloader';
+import { Downloader, type DownloadEvent, isCompletionEvent, isProgressEvent } from '../utils/downloader';
 import type { TaskRegistry } from '../registries/TaskRegistry';
 import type { Task } from '@shared/src/models/ITask';
 
@@ -38,7 +33,7 @@ export class ModelsManager implements Disposable {
   #models: Map<string, ModelInfo>;
   #watcher?: podmanDesktopApi.FileSystemWatcher;
 
-  #downloaders: Map<string, Downloader> = new Map<string, Downloader>;
+  #downloaders: Map<string, Downloader> = new Map<string, Downloader>();
 
   constructor(
     private appUserDirectory: string,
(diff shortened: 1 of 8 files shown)


packages/backend/src/managers/modelsManager.ts
```diff
@@ -171,7 +173,101 @@ export class ModelsManager implements Disposable {
...
+        // telemetry usage
+        this.telemetry.logUsage('model.download', { 'model.id': event.id, durationSeconds: event.duration });
+      }
+    }
+
+    this.taskRegistry.updateTask(task); // update task
+  }
+
+  private createDownloader(model: ModelInfo): Downloader {
+    console.log('Creating a downloader.');
```
lstocchi, line 253: to be removed

packages/backend/src/managers/modelsManager.ts
```diff
@@ -171,7 +173,101 @@ export class ModelsManager implements Disposable {
     }
   }
 
-  async downloadModel(model: ModelInfo, labels?: { [key: string]: string }): Promise<string> {
+  async requestDownloadModel(model: ModelInfo, labels?: { [key: string]: string }): Promise<string> {
+    // Check there is no existing downloader running
+    if (!this.#downloaders.has(model.id)) {
+      console.debug('no downloader has been found.');
```
lstocchi, line 179: remove

lstocchi (commented):
Code LGTM, a couple of console log to be removed. Gonna give it a try

axel7083:
> If i try to start 2 recipes (summarizer and code gen) with the same model, the first works fine i see the model downloading progressing (and also in the models page), on the second recipe i just see the checkout repository status and stop. Like it is stuck
> 
> image
> 
> when the model is downloaded (i see it in the models page) an error popup image if i go to the containers page i see the pod has been created but if i start it it says libpod error

That's the **very very** annoying case. I think we should not allow starting simultaneously multiple recipe at _the same time_.

jeffmaury:
I don't understand why we can start multiple recipe using the same model. When a task is found for the model, instead of updating it for the recipe, can't we add a new task for recipe_id, model_id ?

axel7083:
> I don't understand why we can start multiple recipe using the same model. When a task is found for the model, instead of updating it for the recipe, can't we add a new task for recipe_id, model_id ?

We can, i will work on that this afternoon

axel7083:
@lstocchi @jeffmaury thanks for your feedback, I updated the implementation to properly manage multiple AI-App being downloaded at the same time. I updated the video demonstration in the pull request.

PR #397, opened 2024-02-21
fix: remove previous tasks when starting application

### LINKED ISSUE #396: Starting an app should delete previous tasks
When the stops an application, then starts it later, the tasks for the previous deletion are still displayed. Only the tasks for the current operation should be displayed

https://github.com/projectatomic/ai-studio/assets/9973512/2a5b37d1-08e5-492f-a6a9-b2c2f6de0808

feloy:
### Screenshot / video of UI

Before

https://github.com/projectatomic/ai-studio/assets/9973512/9d42dbb4-7b0e-4894-8ed2-7c108a4ca022

After

https://github.com/projectatomic/ai-studio/assets/9973512/d80e14cf-1511-49a4-bfb5-912742da9060

Fixes #396 

### How to test this PR?

Stop then start an application

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -93,6 +93,15 @@ export class ApplicationManager {
   }
 
   async pullApplication(recipe: Recipe, model: ModelInfo) {
+    // clear any existing status / tasks related to the pair recipeId-modelId.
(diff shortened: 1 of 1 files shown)


axel7083 (approved):
Since I changed the tasks system we need to cleanup our work! Thanks for the fix, LGTM

PR #410, opened 2024-02-21
fix: change model names to include Huggingface organization

### LINKED ISSUE #362: Model names should include huggingface organization name
In the models tab, when models are pulled from huggingface they should include the organization name. For example `TheBloke/Llama-7B-Chat-GGUF`. This will make the source of the model more clear.

jeffmaury:
Fixes #362

### What does this PR do?

Change model names

### Screenshot / video of UI

image

#362 

### How to test this PR?

Go to the models page


lstocchi:
It's happening something weird to me. I went to the models page of a recipe and it's empty and also the right sidebar is different like this is based on an old version but it seems to correctly rebased.

image

lstocchi (commented):
You also need to update the model id in each recipe

lstocchi (approved):
LGTM

feloy (approved):

PR #422, opened 2024-02-26
build: add workflow for playground image

### LINKED ISSUE #375: Provide workflow for playground image
Also a tag like playground-version should be created

jeffmaury:
Fixes projectatomic/ai-studio-playground-images#2

### What does this PR do?

Add workflow for publishing playground image(s)

### Screenshot / video of UI

N/A

projectatomic/ai-studio-playground-images#2 

### How to test this PR?

Check workflows

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,9 @@ updates:
     schedule:
       interval: daily
     open-pull-requests-limit: 10
+
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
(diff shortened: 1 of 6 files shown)


benoitf (commented):
it's worrying me that playground is in `packages`

the packages/* folder are yarn workspaces 
https://github.com/projectatomic/ai-studio/blob/main/package.json#L50

so I would probably move the playground to another directory

but then, as it's not related to the same lifecycle I would move out to its own repository as anyway it's publishing its own package `ai-studio-playground-chat` I would expect to have then `ai-studio-playground-chat` repository with this Containerfile

PR #443, opened 2024-03-05
feat(ContainerRegistry): start container event listener

axel7083:
### What does this PR do?

Adding a startup event for the _future_ InferenceManager (https://github.com/projectatomic/ai-studio/issues/434). Currently when a container is manually stop (and not removed) and start, we do not have access to this event.

> This is because today, in the Playground when creating a container we use `AutoRemove: true`. We need to move out of this option.

Related to https://github.com/projectatomic/ai-studio/issues/434

### How to test this PR?

### DIFF
--- a/packages/backend/src/registries/ContainerRegistry.spec.ts
+++ b/packages/backend/src/registries/ContainerRegistry.spec.ts
@@ -154,6 +154,8 @@ test('ContainerRegistry should fire ContainerStart when container start', () => 
   // Generate a remove event
   callback({ status: 'remove', id: 'random', type: 'container' });
 
+  expect(startListenerMock).not.toHaveBeenCalled();
+
   // Call it a second time
   callback({ status: 'start', id: 'random', type: 'container' });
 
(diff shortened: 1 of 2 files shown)


packages/backend/src/registries/ContainerRegistry.spec.ts
```diff
@@ -121,3 +136,27 @@ test('ContainerRegistry subscriber disposed should not be called', () => {
...
+  // Create the ContainerRegistry and init
+  const registry = new ContainerRegistry();
+  registry.init();
+
+  const startListenerMock = vi.fn();
+  registry.onStartContainerEvent(startListenerMock);
+
+  // Generate a remove event
+  callback({ status: 'remove', id: 'random', type: 'container' });
+
```
feloy, line 156: I would add a `not.toHaveBeenCalled()` here, to be sure that startListenerMock has been called during the second call to `callback`, not the first one

axel7083: Yes good suggestion ! Thanks and added ✅  !

feloy (approved):

PR #444, opened 2024-03-05
feat: adding InferenceManager

axel7083:
### What does this PR do?

Adding a fully autonomous Inference Manager. It replaces most of the logic that was made inside the PlaygroundManager. This is the firsts step for having a `Service` page, specific for the models, and independent of the playground. 

Requires 
- [x] https://github.com/projectatomic/ai-studio/pull/443
- [x] https://github.com/containers/podman-desktop/pull/6257

Fixes https://github.com/projectatomic/ai-studio/issues/434

### How to test this PR?


packages/backend/src/utils/inferenceUtils.ts
```diff
@@ -0,0 +1,147 @@
...
+    },
+    HealthCheck: {
+      // must be the port INSIDE the container not the exposed one
+      Test: ['CMD-SHELL', `curl -sSf localhost:8000/docs > /dev/null`],
+      Interval: 1_000_000_000 * 15, // 15s
+      Retries: 4 * 5, // 20 * 15s = 300s = 5 minutes
+    },
+    Labels: {
+      ...config.labels,
+      [LABEL_INFERENCE_SERVER]: JSON.stringify(config.modelsInfo.map(model => model.id)),
```
jeffmaury, line 142: The server is started only for the first model, why labels are about all models

axel7083: For now we only support one model. But in the future the Inference server can support several models see https://github.com/redhat-et/locallm/pull/72. So using something like `MODEL_ID` does not make sense as we want to be able to link more than one model to a single Inference server;

packages/backend/src/managers/inference/inferenceManager.ts
```diff
@@ -0,0 +1,362 @@
...
+   */
+  async createInferenceServer(config: InferenceServerConfig): Promise<void> {
+    if (!this.isInitialize()) throw new Error('Cannot start the inference server: not initialized.');
+
+    // Fetch a provider container connection
+    const provider = getProviderContainerConnection(config.providerId);
+
+    // Get the image inspect info
+    const imageInfo: ImageInfo = await getImageInfo(provider.connection, config.image, (event: PullEvent) => {
+      console.debug('pull image event', event);
```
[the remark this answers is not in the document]
axel7083: Done

packages/backend/src/utils/inferenceUtils.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  return imageInfo;
+}
+
+/**
+ * Given an {@link InferenceServerConfig} and an {@link ImageInfo} generate a container creation options object
+ * @param config the config to use
+ * @param imageInfo the image to use
+ */
+export function GenerateContainerCreateOptions(
```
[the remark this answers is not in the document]
axel7083: Yes you are right, nice catch

packages/backend/src/utils/inferenceUtils.ts
jeffmaury: So if you find such a container you will think it supports several models where it has been started with a single one

axel7083: https://github.com/projectatomic/ai-studio/blob/fa1fd271259eb9a94c9c357cc2b833afbfa43214/packages/backend/src/utils/inferenceUtils.ts#L101-L103

If you go just a little higher at line 103, you can see why I am only using the first at line 105, because currently we do not support more than one model. I still include all of them, because of guard at line 103, which prevent from having more than one.

In the future, when we will support than one, we will already have the proper logic to handle multiples, as we will simply remove the guard at line 103.

jeffmaury (approved):
LGTM

PR #457, opened 2024-03-07
fix: disable selinux label when mounting the volume for the playground

### LINKED ISSUE #456: Playground cannot start on Linux and MacOS applehv

jeffmaury:
Fixes #456

### What does this PR do?

Change options while starting the playground server

### Screenshot / video of UI

N/A

#456

### How to test this PR?

Start the playground for a model on Linux or MacOS applehv

### DIFF
--- a/packages/backend/src/managers/playground.ts
+++ b/packages/backend/src/managers/playground.ts
@@ -33,7 +33,7 @@ import type { PlaygroundState, PlaygroundStatus } from '@shared/src/models/IPlay
 import type { ContainerRegistry } from '../registries/ContainerRegistry';
 import type { PodmanConnection } from './podmanConnection';
 import OpenAI from 'openai';
-import { getDurationSecondsSince, timeout } from '../utils/utils';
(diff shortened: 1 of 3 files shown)


packages/backend/src/managers/playground.ts
```diff
@@ -212,6 +212,7 @@ export class PlayGroundManager {
             Type: 'bind',
           },
         ],
+        SecurityOpt: [DISABLE_SELINUX_LABEL_SECURITY_OPTION],
```
axel7083, line 215: What are the implication of this security option ? Why does it is required ? I cannot really understand the link with applehv from the documentation https://docs.podman.io/en/v4.6.1/markdown/options/security-opt.html

jeffmaury: The volumes are not readable if this option is not set on Linux and applehv. Inspired from https://github.com/containers/podman/issues/3683

axel7083: So we might want to include this option on windows or we don't care ?

jeffmaury: It is set for all platforms, I prefer to have a single general flow

PR #458, opened 2024-03-07
feat(models): download to `.tmp` file

axel7083:
### What does this PR do?

Download the model to `*.tmp` file in the targeted directory, then renaming when download finish to the name expected.

Fixes https://github.com/projectatomic/ai-studio/issues/389

### How to test this PR?

### DIFF
--- a/packages/backend/src/utils/downloader.ts
+++ b/packages/backend/src/utils/downloader.ts
@@ -131,12 +131,6 @@ export class Downloader {
         .catch((err: unknown) => {
           callback({ error: `Something went wrong while trying to rename downloaded file: ${String(err)}.` });
         })
-        .finally(() => {
-          // Finally delete the tmp file
-          promises.rm(tmpFile).catch((err: unknown) => {
-            console.error('Something went wrong while trying to delete the temporary file', err);
-          });
-        });
     });
     stream.on('error', e => {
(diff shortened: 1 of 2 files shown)


benoitf (approved):

packages/backend/src/utils/downloader.ts
```diff
@@ -117,7 +117,33 @@ export class Downloader {
...
+          promises.rm(tmpFile).catch((err: unknown) => {
+            console.error('Something went wrong while trying to delete the temporary file', err);
+          });
+        });
+    });
+    stream.on('error', e => {
+      callback({
+        error: e.message,
+      });
+    });
```
benoitf, line 145: I don't know why but it loose my comment, I was saying that it would have been more readable to me extracting the code to an async function
```
async renameFile(): Promise<void> {
try {

} catch {

} finally {

}
```
and doing in the stream.on('finish')
```ts
   renameFile(..).catch...
```

axel7083: I was almost scared when I did not see any comments and just an approval !!

axel7083: And I just noticed, I am renaming, not copy past, I do not need to delete....

PR #463, opened 2024-03-07
fix: remove categories from catalog interface

axel7083:
### What does this PR do?

We need to clear out some old dependencies/requirements (we have some hardcoded _recent_ category etc. which are not used), moreover categories where at the same level than Models and Recipe, removing for now. 

> We can re-introduce then later when we would have completed https://github.com/projectatomic/ai-studio/issues/447

Fixes https://github.com/projectatomic/ai-studio/issues/461

### How to test this PR?


jeffmaury:
I don't understand the rationale for this. This is a change from the mockup and I don't remember being addressed during UI Call.
@slemeur @ekidneyrh WDYT ?

axel7083:
> I don't understand the rationale for this. This is a change from the mockup and I don't remember being addressed during UI Call. @slemeur @ekidneyrh WDYT ?

If you want to split the models and the recipes. What do we do with the categories ?

jeffmaury:
> > I don't understand the rationale for this. This is a change from the mockup and I don't remember being addressed during UI Call. @slemeur @ekidneyrh WDYT ?
> 
> If you want to split the models and the recipes. What do we do with the categories ?

They are related to recipes (a way to group and display them)

axel7083:
> > > I don't understand the rationale for this. This is a change from the mockup and I don't remember being addressed during UI Call. @slemeur @ekidneyrh WDYT ?
> > 
> > 
> > If you want to split the models and the recipes. What do we do with the categories ?
> 
> They are related to recipes (a way to group and display them)

Okey got it. I will close this PR and include the change in https://github.com/projectatomic/ai-studio/pull/464

PR #472, opened 2024-03-08
refactor(git): make method static

axel7083:
### What does this PR do?

Remove the GitManager in favour of a simple static method.

Fixes https://github.com/projectatomic/ai-studio/issues/471
### How to test this PR?

- [x] Existing tests ensure no regression


benoitf (commented):
I would say I tend to prefer class over functions as they better fit for testing

benoitf:
> This changes does not make it harder to test, as we are able to keep the same test, with minimal change.

well, you need to add a top level vi.mock entry `vi.mock('../utils/gitUtils', () => ({` each time you are using that function
and no typesafety

vs you can just spy a given method to override the behaviour

To me it's a less maintainable approach using functions

PR #473, opened 2024-03-08
proposal(backend): create a ManagerRegistry

axel7083:
### What does this PR do?

See `docs/proposal/managers-registry.md`.


docs/proposals/managers-registry.md
```diff
@@ -0,0 +1,72 @@
...
+Inside the StudioAPI we can have an instance of the `ManagerRegistry` and calling the `getInstance` for each manager we require when the api is triggered.
+
+````ts
+import type { ModelsManager } from './modelsManager';
+
+... 
+
+async getModelsDirectory(): Promise <string> {
+  return this.managers.getInstance(ModelsManager).getModelsDirectory();
+}
```
feloy, line 24: I would prefer to use this instead:
```
init() {
  this.modelsManager = this.managers.getInstance(ModelsManager);
  [...]
}

async getModelsDirectory(): Promise <string> {
  return this.modelsManager.getModelsDirectory();
}
```

This way, we won't need to rewrite all the calls to methods, but only the init methods. We can develop later if the init methods of the dependencies are too much heavy.

axel7083: I would agree at first to simply the usage maybe. But this remove the lazy loading we would get from this proposal

axel7083: I also think it would simplify the StudioApiImpl class to not keep those many private properties

axel7083:
Closing for now

PR #482, opened 2024-03-11
feat: introduce snippet manager

axel7083:
### What does this PR do?

Using the library open source [postman-code-generators](https://github.com/postmanlabs/postman-code-generators) to generate code snippet.

Currently this only generate dummy `get` request snippet or a given URL. We can use this as a base to generate more complexe snippet.

Fixes https://github.com/projectatomic/ai-studio/issues/435

### How to test this PR?

- [x] unit tests has been provided.


packages/backend/src/managers/SnippetManager.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { beforeEach, expect, test, vi } from 'vitest';
+import { SnippetManager } from './SnippetManager';
+import { getLanguageList, convert } from 'postman-code-generators';
+import type { Webview } from '@podman-desktop/api';
+import { Request } from 'postman-collection';
+import { MSG_SUPPORTED_LANGUAGES_UPDATE } from '@shared/Messages';
+vi.mock('postman-code-generators', () => {
```
benoitf, line 25: it's nice to mock here for the manager tests, but I think we're missing a test to really test the postman-code-generators

For example, if there is an automated PR bumping the version of postman-code-generators, there won't be any test that will check the behavior of the library

I think then we should use nock to mock a request and see how postman-code-generators behave

jeffmaury (commented):
I have 2 interrogations:

- seems language variant is mandatory can't it be optional
- I see a lot of replacement from yarnpkg.com to npmjs.org that seems unrelated

axel7083:
> * seems language variant is mandatory can't it be optional

When using `listLanguages()` you get an array of `Language` which contains a list of variants. There will always have at least one.

> * I see a lot of replacement from yarnpkg.com to npmjs.org that seems unrelated

Weird indeed, I will take a look

packages/backend/src/managers/SnippetManager.spec.ts
axel7083: I adapted the tests

benoitf: thanks, we'll able to trust tests in case of version update 🎉

benoitf (approved):

types/postman-code-generators.d.ts
```diff
@@ -0,0 +1,28 @@
+declare module 'postman-code-generators' {
```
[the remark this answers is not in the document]
axel7083: Yes, but adding a `@types/` is time consuming, I have it somewhere inside my head for now this will do the trick

PR #484, opened 2024-03-11
refactor: using enum for message channel

axel7083:
### What does this PR do?

We are declaring manually `export const` which can lead to duplicates, and could create issue. Using an enum ensure we do not have duplicates in the channel used.

### How to test this PR?


jeffmaury (commented):
I would rename MESSAGES as Messages

axel7083:
> I would rename MESSAGES as Messages

Agreed ✅

jeffmaury (approved):
LGTM

PR #491, opened 2024-03-11
fix: upversion when doing release

### LINKED ISSUE #460: Version should be upversioned when doing release

jeffmaury:
Fixes #460

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -79,6 +79,35 @@ jobs:
           draft: true
           prerelease: false
 
+      - name: Create the PR to bump the version in the main branch (only if we're tagging from main branch)
+        if: ${{ github.event.inputs.branch == 'main' }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -79,6 +79,35 @@ jobs:
           draft: true
           prerelease: false
 
+      - name: Create the PR to bump the version in the main branch (only if we're tagging from main branch)
+        if: ${{ github.event.inputs.branch == 'main' }}
+        run: |
```
axel7083, line 84: Is this bash from podman-desktop?

jeffmaury: Not sure WDYM ?

benoitf (approved):

.github/workflows/release.yaml
jeffmaury: It was inspired by Podman Desktop and bootc extension

PR #503, opened 2024-03-12
fix: propagate security option to inference server

axel7083:
### What does this PR do?

Propagate https://github.com/projectatomic/ai-studio/pull/457 fix to inference server config


lstocchi (approved):
LGTM

PR #508, opened 2024-03-12
feat: adding Services page

axel7083:
### What does this PR do?

Adding a very basic _Services_ page to list the inference servers. The Status is updated depending on the container status `stopped` | `running` then based on the health check result for better details `starting` | `unhealthy` | `healthy`.

### Screenshot / video of UI

**demo**

https://github.com/projectatomic/ai-studio/assets/42176370/e0ab66c3-550f-42da-afe9-4a37ea4fccb0

> ⚠️ I am using a trick to get access to the `studioclient` from the devtools. To reproduce `(window as any).studioclient = studioClient;` inside the `packages/frontend/src/utils/client.ts` This is necessary as we do not provide yet the form to create the Inference Server.

**healthy**

image

**unhealthy**

image

**multiple service running**

image

Fixes https://github.com/projectatomic/ai-studio/issues/507

### How to test this PR?

- [x] unit test has been provided


packages/backend/src/utils/inferenceUtils.ts
```diff
@@ -131,7 +131,7 @@ export function generateContainerCreateOptions(
     },
     HealthCheck: {
       // must be the port INSIDE the container not the exposed one
-      Test: ['CMD-SHELL', `curl -sSf localhost:8000/docs > /dev/null`],
+      Test: ['CMD-SHELL', `curl -sSf localhost:9990/docs > /dev/null`],
```
axel7083, line 134:
```suggestion
      Test: ['CMD-SHELL', `curl -sSf localhost:8000/docs > /dev/null`],
```

packages/frontend/src/lib/table/service/ServiceAction.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+import type { InferenceServer } from '@shared/src/models/IInference';
+import { studioClient } from '/@/utils/client';
+import { faPlay, faStop, faTrash } from '@fortawesome/free-solid-svg-icons';
+import ListItemButtonIcon from '/@/lib/button/ListItemButtonIcon.svelte';
+export let object: InferenceServer;
+
+function stopInferenceServer() {
+  studioClient.stopInferenceServer(object.container.containerId);
```
lstocchi, line 9: Should we display the error if the stop fails?

packages/frontend/src/lib/table/service/ServiceAction.svelte
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function startInferenceServer() {
+  studioClient.startInferenceServer(object.container.containerId);
```
lstocchi, line 13: Should we display the error if the start fails?

packages/frontend/src/pages/InferenceServers.svelte
```diff
@@ -0,0 +1,36 @@
...
+];
+const row = new Row<InferenceServer>({});
+
+let data: InferenceServer[];
+$: data = $inferenceServers;
+</script>
+
+<NavPage title="Services" searchEnabled="{false}">
+  <svelte:fragment slot="content">
+    <div slot="content" class="flex flex-col min-w-full min-h-full">
```
lstocchi, line 24: Shouldn't these have different slot names?

packages/frontend/src/pages/InferenceServer.spec.ts
```diff
@@ -0,0 +1,75 @@
+/**********************************************************************
```
lstocchi, line 1: Missing an s in the filename -> `InferenceServers.spec.ts`

packages/frontend/src/pages/InferenceServers.svelte
axel7083: That's a copy past from Models.svelte so I don't know

packages/frontend/src/lib/table/service/ServiceAction.svelte
axel7083: We are not doing it inside the Environment page, so I do not think so ? Where would you want it to be displayed ?

packages/frontend/src/pages/InferenceServer.spec.ts
axel7083: Oh yes thanks

packages/frontend/src/lib/table/service/ServiceAction.svelte
axel7083: I added a catch block so at least for now we do not lost it

packages/frontend/src/pages/InferenceServers.svelte
lstocchi: Just verified and the inner div slot is never used.
So also on Models.svelte should be removed.
BTW as it does not give any issue it can be removed in another PR.

lstocchi (approved):
LGTM

PR #510, opened 2024-03-12
feat: adding @podman-desktop/ui-svelte

axel7083:
### What does this PR do?

Adding the `@podman-desktop/ui-svelte` library using `1.10.3`.

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1165

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -12,7 +12,7 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
-    "@podman-desktop/ui-svelte": "next",
+    "@podman-desktop/ui-svelte": "1.10.3",
     "tinro": "^0.6.12"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


benoitf:
I think there is a missing entry in tailwind.config.js to scan svelte and css files from the`@podman-desktop/ui-svelte`  in node_modules/

also maybe need to introduce it after 0.2.0 tag

benoitf (changes requested):
it's a good time to introduce it

but it needs two things IMHO
- [x] requires a bump in the version as there was an issue in previous builds
- [x] addition into the tailwind config file like https://github.com/containers/podman-desktop-extension-bootc/blob/main/packages/frontend/tailwind.config.cjs#L26

benoitf:
converting to draft as not updated

axel7083:
cc @benoitf

benoitf (approved):

Issue #479, opened 2024-03-11
Polish list of Models UI


slemeur:
- icons
- font styles
- table and pre-requisites

feloy:
@slemeur could you please point to the mockup you are refering to?

vrothberg:
@feloy do you have everything you need or are there open questions?

feloy:
> @feloy do you have everything you need or are there open questions?

I need more information about the **icon** and **table and pre-requisites**

- icon: for the moment, there is no icon field for the models definition in the catalog file `ai.json`. Do we want to add it?

- table and pre-requisites: which pre-requisites?

jeffmaury:
> > @feloy do you have everything you need or are there open questions?
> 
> I need more information about the **icon** and **table and pre-requisites**
> 
>     * icon: for the moment, there is no icon field for the models definition in the catalog file `ai.json`. Do we want to add it?
> 
>     * table and pre-requisites: which pre-requisites?

Icon could be based on registry

vrothberg:
@slemeur PTAL

vrothberg:
@jeffmaury WDYT?

jeffmaury:
> @jeffmaury WDYT?

Discussed with @slemeur yesterday

feloy:
As discussed during UX call (22/03):
- we will display a generic Model icon on each line (we cannot use the Hugging Face icon). The icon could be colorized to differentiate downloaded models from others
- the cpu/gpu prerequisite is already displayed (HW Compat column)
- we won't display the size of the file
- we will display the memory required (this information needs to be added first to the model's information in the catalog)

slemeur:
Also please use consistent font sizes with the list of containers

slemeur:
Please keep the size of the model.

feloy:
> * we won't display the size of the file when the model is not downloaded

> we won't display the size of the file **when the model is not downloaded**

This was probably not clear. It has beensaid to not display the size of the file **before the model is downloaded**, but we will keep the file size once it is on disk.

slemeur:
OK perfect! thanks for clarifying

feloy:
> Also please use consistent font sizes with the list of containers

I have created this issue https://github.com/projectatomic/ai-studio/issues/667. We need to be careful as this can have impact on many other places

PR #511, opened 2024-03-12
ui: set correct font for name column in models list

feloy:
### What does this PR do?

Change the font of the first column in the models list, to have the same as in other lists

Part of #479

### DIFF
--- a/packages/frontend/src/lib/table/model/ModelColumnName.svelte
+++ b/packages/frontend/src/lib/table/model/ModelColumnName.svelte
@@ -8,6 +8,8 @@ function openDetails() {
 }
 </script>
 
-<button class="text-sm text-gray-700" on:click="{() => openDetails()}">
-  {object.name}
+<button on:click="{() => openDetails()}">
+  <div class="text-sm text-gray-300 overflow-hidden text-ellipsis">
+    {object.name}
+  </div>
 </button>


axel7083 (approved):

packages/frontend/src/lib/table/model/ModelColumnName.svelte
```diff
@@ -8,6 +8,8 @@ function openDetails() {
 }
 </script>
 
-<button class="text-sm text-gray-700" on:click="{() => openDetails()}">
-  {object.name}
+<button on:click="{() => openDetails()}">
+  <div class="text-sm text-gray-300 overflow-hidden text-ellipsis">
```
lstocchi, line 12: Should this be white? 
By looking at the color here https://design.penpot.app/#/view/52961d58-0a92-80c2-8003-2a6919ad37d6?page-id=51af7f49-bd09-8090-8003-8c5efc23847f&section=interactions&index=12 it seems to me the name is white and the rest is gray-300

feloy: I find difficult to extract the color from the mockup.

I reused the one used in other tables (AI Apps, Pods, etc)

feloy: Not sure if it is intentional to have different colors in this specific table, compared to other tables.

Should I 100% respect the mockup (white for first column, gray-300 for other columns), or have a consistent design across tables (gray-300 for first column, gray-700 for others)? @ekidneyrh WDYT?

feloy: Do we want to make this change of colors across all pages in Podman Desktop too? Or only on AI Studio?

If only on AI Studio, I can do it as part of this PR. If it's across Podman Desktop, I would prefer to sync with PD as part of another PR

lstocchi (approved):
LGTM

PR #512, opened 2024-03-12
fix: fix model table header and image size

lstocchi:
### What does this PR do?

I didn't notice that https://github.com/projectatomic/ai-studio/pull/506 made the models table header have a different color compared to the background and also the images background were shrinked. This fixes both

### Screenshot / video of UI

fix_ui

N/A

### How to test this PR?

1. check icons and models table header


axel7083 (approved):
Nice improvement ! LGTM

Issue #451, opened 2024-03-06
Cannot start 2 recipes in parralel


jeffmaury:
If they use the same port, we assume the first one use the port because it is free but the second one may think the same if the build of image is long or starting the server is long.

We should use random ports in all cases

feloy:
Another solution is to server.listen(0) to get a random port, close the server, and reuse the port immediately.

PR #514, opened 2024-03-12
fix: set mapped ports on app labels instead of original ones

feloy:
### What does this PR do?

When several AI apps are started, with same ports defined in ai-studio.yaml, new ports will be affected, by detecting free ones. 

This PR sets the real affected ports, instead of the original ones, in the pods labels, so the ports displayed in the Models list are correct.

Part of #451 

### How to test this PR?

Start the ChatBot and the Summarizer AI Apps.

Go to the AI Apps list page, the ports should be different for theses two Apps. By clicking on the link to access the app, you should go to the good app.


feloy:
Not necessary, as we want to use random port (#451), we will create the pod with localPort set to 0 (to ask the kernel to assign random free port), and we will need to get the assigned port after the pod is created, and set the labels after. So this change will not be necessary

PR #521, opened 2024-03-13
remove support for port ranges

feloy:
### What does this PR do?

We were previously getting the ports from the EXPOSE directive of the Containerfile used to build the model's and recipe's images, which support port ranges.

We are now getting the ports from the ai-studio.yaml file, which is defined as an array of numbers (https://github.com/projectatomic/ai-studio/blob/main/packages/backend/src/managers/applicationManager.ts#L457).

```
 imageInfoList.push({
   id: image.Id,
   modelService: container.modelService,
   ports: container.ports?.map(p => `${p}`) ?? [], <-- p is of type number
   appName: container.name,
});
```

For this reason, we don't need to support anymore that a port is defined as a range

Part of #451

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -329,7 +329,6 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
     }
 
     const portmappings: PodCreatePortOptions[] = [];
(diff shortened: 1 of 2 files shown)


axel7083 (approved):
LGTM! Thanks for cleaning !!

benoitf (approved):

PR #524, opened 2024-03-13
fix: set mapped (random) ports on app labels instead of original ones

### LINKED ISSUE #451: Cannot start 2 recipes in parralel

feloy:
### What does this PR do?

This PR changes the way the port is selected for the host port of the containers used on a recipe's pod.

Before, the first free port following the internal port were used.

If two recipes were started in a short term, and the detection for the second recipe were done before the first pod was created (and the port effectively used), the same ports were used for both recipes.

With this technique, we are opening/closing a server with a port set to 0, for the kernel to select 'random' free ports for us, and we reuse this free port for the recipe's containers. This reduces the risks of collision.

Later (when https://github.com/containers/podman-desktop/issues/6399 ix fixed), to completely avoid collisions, we will be able to create the pod directly with setting the hostPort to 0 and getting back the allocated port with `getPodInspect`.

Fixes #451 

### How to test this PR?

Delete recipes images, and start several recipes in a row.

There should be no ports conflicts between AI apps.

- [x] tested on mac os arm
- [x] tested on Fedora Linux
- [ ] tested on Windows 11

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -857,7 +857,9 @@ describe('createPod', async () => {
   test('call createPod with sample app exposed port', async () => {
     const images = [imageInfo1, imageInfo2];
     vi.spyOn(manager, 'getRandomName').mockReturnValue('name');
-    vi.spyOn(portsUtils, 'getPortsInfo').mockResolvedValue('9000');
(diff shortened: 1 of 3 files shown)


axel7083 (approved):
image

Seems to be working fine on Windows 11

jeffmaury (approved):
LGTM

PR #529, opened 2024-03-13
feat: service creation form

axel7083:
### What does this PR do?

Adding a `rocket` icon to the Models action (for models that are downloaded). This action will open the `Creation Service` form allowing to create a service.

### Screenshot / video of UI

image

Fixes https://github.com/projectatomic/ai-studio/issues/528

### How to test this PR?

**manually**
(1) Go to models page
(2) Click on the rocket icon (need model to be downloaded)
(3) Click on `Create service`

### DIFF
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -13,6 +13,7 @@ import Recipe from '/@/pages/Recipe.svelte';
 import Model from './pages/Model.svelte';
 import { onMount } from 'svelte';
 import { getRouterState } from '/@/utils/client';
+import CreateService from '/@/pages/CreateService.svelte';
 import Services from '/@/pages/InferenceServers.svelte';
 import ServiceDetails from '/@/pages/InferenceServerDetails.svelte';
 
@@ -60,6 +61,7 @@ onMount(() => {
       <Route path="/models/*" breadcrumb="Models">
         <Models />
       </Route>
+
       <Route path="/model/:id/*" breadcrumb="Model Details" let:meta>
         <Model modelId="{meta.params.id}" />
       </Route>
@@ -69,7 +71,11 @@ onMount(() => {
       </Route>
 
       <Route path="/service/:id/*" breadcrumb="Service Details" let:meta>
-        <ServiceDetails containerId="{meta.params.id}" />
+        {#if meta.params.id === 'create'}
+          <CreateService />
+        {:else}
+          <ServiceDetails containerId="{meta.params.id}" />
+        {/if}
       </Route>
     </div>
   </main>
--- a/packages/frontend/src/lib/table/model/ModelColumnAction.spec.ts
+++ b/packages/frontend/src/lib/table/model/ModelColumnAction.spec.ts
@@ -21,6 +21,7 @@ import { test, expect, vi, beforeEach } from 'vitest';
 import { fireEvent, render, screen, waitFor } from '@testing-library/svelte';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import ModelColumnActions from '/@/lib/table/model/ModelColumnActions.svelte';
+import { router } from 'tinro';
 
 const mocks = vi.hoisted(() => ({
   requestRemoveLocalModel: vi.fn(),
@@ -71,6 +72,9 @@ test('Expect folder and delete button in document', async () => {
   const deleteBtn = screen.getByTitle('Delete Model');
   expect(deleteBtn).toBeInTheDocument();
 
+  const rocketBtn = screen.getByTitle('Create Model Service');
+  expect(rocketBtn).toBeInTheDocument();
+
   const downloadBtn = screen.queryByTitle('Download Model');
   expect(downloadBtn).toBeNull();
 });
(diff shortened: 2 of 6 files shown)


lstocchi (commented):
Works fine. 
Just two things. 
The first is about the recognition, if i start 2-3 services i don't really know which model they are bound to, the name is not really useful. Maybe adding a new column with the model name mounted?
The second is about the port, it would be good to update the port if we detect that 8888 is busy. If i start a service without pay attention to the port i gets redirected to the services page and i don't see anything new. Then if i go to the container page i see that it was not started because of the port already in use.

axel7083:
> Maybe adding a new column with the model name mounted? 

Yes, the name of the container is clearly not useful, we can open an issue solving that

> The second is about the port, it would be good to update the port if we detect that 8888 is busy.

We then need an api `getFreePort()` or something.

jeffmaury:
> > Maybe adding a new column with the model name mounted?
> 
> Yes, the name of the container is clearly not useful, we can open an issue solving that
> 
> > The second is about the port, it would be good to update the port if we detect that 8888 is busy.
> 
> We then need an api `getFreePort()` or something.

@feloy just added a method for that

axel7083:
@jeffmaury thanks for your comment!

> * Name is too cryptic we should generate modelid-1 then modelid-2

Name is hidden for now, but I think it was @slemeur would said we would require in the future to let the user choose a name for it ?

> * I think image name is confusing for users should be removed

From a developer point of vue, I think it make sense to show the image we will use, otherwise it feel like magic, this is an optional field with our image by default. We could add some text details under the title to have more details about it, but I think that it empower the developer to have a better understanding of what is happening and how you can reproduce what we provide you.

> * Port is not checked and may lead to errors we should use a dynamic port and if using an already bound port no error is reported

As mention earlier, we need some kind of `getFreePort()` method. (cc @feloy I think you worked on something similar ? could it be reuse ?)

> * The tooltip on the icon is RUNNING would be good to have RUNNING on port_number

I don't understand what/where you talking about

jeffmaury:
> @jeffmaury thanks for your comment!
> 
> > * Name is too cryptic we should generate modelid-1 then modelid-2
> 
> Name is hidden for now, but I think it was @slemeur would said we would require in the future to let the user choose a name for it ?
> 
> > * I think image name is confusing for users should be removed
> 
> From a developer point of vue, I think it make sense to show the image we will use, otherwise it feel like magic, this is an optional field with our image by default. We could add some text details under the title to have more details about it, but I think that it empower the developer to have a better understanding of what is happening and how you can reproduce what we provide you.
> 
> Moreover I think it would make some consistency within podman desktop, for example when creating a kind cluster
> 
> image
> 
> > * Port is not checked and may lead to errors we should use a dynamic port and if using an already bound port no error is reported
> 
> As mention earlier, we need some kind of `getFreePort()` method. (cc @feloy I think you worked on something similar ? could it be reuse ?)
> 
> > * The tooltip on the icon is RUNNING would be good to have RUNNING on port_number
> 
> I don't understand what/where you talking about
When a model is running there is a tooltip shown on the left icon and it says RUNNING would be good to have RUNNING on xxxx (being the port number)

jeffmaury:
> @jeffmaury thanks for your comment!
> 
> > * Name is too cryptic we should generate modelid-1 then modelid-2
> 
> Name is hidden for now, but I think it was @slemeur would said we would require in the future to let the user choose a name for it ?
> 
> > * I think image name is confusing for users should be removed
> 
> From a developer point of vue, I think it make sense to show the image we will use, otherwise it feel like magic, this is an optional field with our image by default. We could add some text details under the title to have more details about it, but I think that it empower the developer to have a better understanding of what is happening and how you can reproduce what we provide you.
> 
> Moreover I think it would make some consistency within podman desktop, for example when creating a kind cluster
> 
> image
> 
> > * Port is not checked and may lead to errors we should use a dynamic port and if using an already bound port no error is reported
> 
> As mention earlier, we need some kind of `getFreePort()` method. (cc @feloy I think you worked on something similar ? could it be reuse ?)
> 
> > * The tooltip on the icon is RUNNING would be good to have RUNNING on port_number
> 
> I don't understand what/where you talking about

On the playground list, when you hover on the left icon the tooltip is RUNNING would be better as RUNNING on xxxx (the port number)

axel7083:
> On the playground list, when you hover on the left icon the tooltip is RUNNING would be better as RUNNING on xxxx (the port number)

First this is not related to this PR, as this has been added in https://github.com/projectatomic/ai-studio/pull/508 and for the container icon, _sadly_ we are using the `StatusIcon.svelte` from podman-desktop and we cannot set a custom tooltip as we are limited to `RUNNING` | `USED` | `STARTING` | `DEGRADED` | `DELETING` :(

axel7083:
@lstocchi @jeffmaury rebased and updated to use the `getHostFreePort` api !

jeffmaury:
Error (port in use) is not reported but could be a follow up PR as port is now random.
Image field should be removed

axel7083:
> Error (port in use) is not reported but could be a follow up PR as port is now random. Image field should be removed

@jeffmaury I removed the code for the container image, and updated the design

jeffmaury (approved):
LGTM

PR #533, opened 2024-03-13
fix: Provide an button that allows the user to open the recipe app fr…

### LINKED ISSUE #468: Provide an button that allows the user to open the recipe app from the recipe screen
As a user, after starting a recipe, I have no way to understand what I can do next - which should be opening the application.

Screenshot 2024-03-08 at 11 46 16

We would like to add a clear button that indicates to the end-user they can open the application in a browser.
I'd suggest we add a button "Open AI App" as we have "Start AI App".

jeffmaury:
…om the recipe screen

Fixes #468

### What does this PR do?

Add a new Open action for a recipe or an AI app. If there is a single port, then no quickpick and browser is open without user interaction

### Screenshot / video of UI

open

#468

### How to test this PR?

1. Start a recipe
2. From the recipe screen click on the Open icon
3. From the AI apps page you should see the same icon

### DIFF
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -37,6 +37,10 @@ import type { InferenceServer } from '@shared/src/models/IInference';
 import type { InferenceServerConfig } from '@shared/src/models/InferenceServerConfig';
 import type { InferenceManager } from './managers/inference/inferenceManager';
 
+interface PortQuickPickItem extends podmanDesktopApi.QuickPickItem {
+  port: number;
+}
+
 export class StudioApiImpl implements StudioAPI {
   constructor(
     private applicationManager: ApplicationManager,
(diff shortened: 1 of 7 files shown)


packages/backend/src/studio-api-impl.ts
```diff
@@ -236,6 +240,50 @@ export class StudioApiImpl implements StudioAPI {
       });
   }
 
+  async requestOpenApplication(recipeId: string, modelId: string): Promise<void> {
+    const recipe = this.catalogManager.getRecipeById(recipeId);
```
axel7083, line 244: We might want to keep the implementation simple and move the logic outside ?

vrothberg (commented):
Lovely improvement, thanks!

slemeur:
👍

vrothberg (commented):
Is this PR ready to get in?

jeffmaury:
> Is this PR ready to get in?

Needs code refactoring but feature wise it is ready

Issue #525, opened 2024-03-13
Create Playground page


jeffmaury:
A playground page should be added in the navigator bar (left)

That page should list existing playground sessions/conversation with a model and its inteference server

It should also be possible to create a new playground session from there

- [x] #539
- [x] #570
- [x] #580 
- [x] #613 
- [x] #543
- [ ] ~~Code snippets (milestone 1.0)~~
- [x] Make submitPlaygroundMessage reply more quickly (call to `new OpenAI` is too long?)
- [x] #595
- [x] #615 
- [x] #617

axel7083:
depends on https://github.com/projectatomic/ai-studio/issues/509

feloy:
Mockup for playground page:

![model_details_stable_diffusion_playground](https://github.com/projectatomic/ai-studio/assets/9973512/43331230-3f18-4c6b-b221-763577010522)

feloy:
done

PR #539, opened 2024-03-14
feat: Playgrounds navigation

feloy:
### What does this PR do?

Adds a Playground page, and the possibility to create a new playground with an optional name, and a given model.

When a new playground is created and no model service is running for its model, the model server is created and started. If the model server exists but is stopped, it is started.

Out of scope for this PR:
- the playground details page only displays the playground id
- the PlaygroungManagerV2 API needs to be adapted to accept playgroundId as input, and not accept info already known by the playground (modelId)
- the user is not redirected yet to the playground page (as it is empty for now)

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/9973512/7748094d-d755-4ae7-8696-cda4cf213fa0

No models are downloaded yet:

no-model

More models could be downloaded:

some-models

All models are downloaded:

all-models

Part of #525 

### How to test this PR?

Go to the Playground page and use "New Playground" button to create new playgrounds.

Creating new playgrounds should start "Services" for the model chosen for the playground, if not already running

Depending on the models downloaded or not, different messages should appear (see screenhost above)


packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -23,21 +23,62 @@ import type { ModelOptions } from '@shared/src/models/IModelOptions';
...
-
-export class PlaygroundV2Manager implements Disposable {
+import type { PlaygroundV2 } from '@shared/src/models/IPlaygroundV2';
+import { Publisher } from '../utils/Publisher';
+import { Messages } from '@shared/Messages';
+import type { ModelInfo } from '@shared/src/models/IModelInfo';
+import { withDefaultConfiguration } from '../utils/inferenceUtils';
+
+export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Disposable {
+  #playgrounds: Map<string, PlaygroundV2>;
```
axel7083, line 33: We might prefer to add a `modelId` to the Conversation object rather than creating a new store ?

packages/shared/src/StudioAPI.ts
```diff
@@ -144,4 +145,8 @@ export abstract class StudioAPI {
    * Create a new conversation and return a conversationId
    */
   abstract createPlaygroundConversation(): Promise<string>;
+
+  abstract createPlayground(name: string, model: ModelInfo): Promise<void>;
+
+  abstract getPlaygroundsV2(): Promise<PlaygroundV2[]>;
```
axel7083, line 151: If we combine `PlaygroundV2` and `Conversation` we could remove this method

packages/backend/src/managers/playgroundV2Manager.ts
feloy: I have superposed our 2 logics without merging them for now, it will be part of the next PR

axel7083 (approved):
Accepting as it is since it will be simplified in a follow up PR

jeffmaury:
Small hint the when the form is presented for the first time no model is selected by default the first one should be. I don't know why but this does not happen after the first playground has been created.

feloy:
> Small hint the when the form is presented for the first time no model is selected by default the first one should be. I don't know why but this does not happen after the first playground has been created.

should be fixed now. Thanks

jeffmaury (approved):
LGTM

PR #545, opened 2024-03-14
fix: do not use Z flag on QEMU

### LINKED ISSUE #538: Stop using the Z flag on QEMU
@feloy noticed that the Z flag causes issues on QEMU

```
Error starting the application "ChatBot": starting container fed8637c0442e1bb56be148ebba4f34fbcdccbf86861fa60c95ceb17f52a3e26: lsetxattr /Users/phmartin/podman-desktop/ai-studio/models/hf.TheBloke.llama-2-7b-chat.Q5_K_S/llama-2-7b-chat.Q5_K_S.gguf: operation not supported
```

It works fine if it gets removed

lstocchi:
### What does this PR do?

it adds a check so that the Z flag is not used when working with qemu machine

### Screenshot / video of UI

N/A

it resolves #538 

### How to test this PR?

1. run a recipe on qemu

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -29,6 +29,7 @@ import type { AIConfig, ContainerConfig } from '../models/AIConfig';
 import * as portsUtils from '../utils/ports';
 import { goarch } from '../utils/arch';
 import * as utils from '../utils/utils';
+import * as podman from '../utils/podman';
 import type { Webview, TelemetryLogger, PodInfo } from '@podman-desktop/api';
 import type { CatalogManager } from './catalogManager';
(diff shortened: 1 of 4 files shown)


jeffmaury (commented):
As the process to detect if a machine is QEMU is MacOS specific I think we should ensure first we are on MacOS ?

jeffmaury:
> As the process to detect if a machine is QEMU is quite heavy and MacOS specific I think we should ensure first we are on MacOS ?

lstocchi:
> As the process to detect if a machine is QEMU is quite heavy and MacOS specific I think we should ensure first we are on MacOS ?

yes, good point. Updating

jeffmaury (approved):
LGTM. Tested on MacOS and Linux

PR #553, opened 2024-03-14
feat: making InferenceServerConfig creation having more optional options

axel7083:
### What does this PR do?

@feloy raise the point that the frontend should not be responsible of providing the default values for creating an InferenceServer.

This PR is simplifying the options for creating the inference servers. Now, only the ModelInfo is required.

Fixes https://github.com/projectatomic/ai-studio/issues/552

### How to test this PR?


feloy (approved):
thanks for the changes

PR #556, opened 2024-03-14
feat: using SnippetManager in the Service Details to create code snippet

axel7083:
### What does this PR do?

This PR is using the SnippetManager in the frontend to generate code snippet. 

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/1dfc04d1-d728-43ef-876d-8b0f7485ef10

Fixes https://github.com/projectatomic/ai-studio/issues/555 https://github.com/projectatomic/ai-studio/issues/403

### How to test this PR?

- [x] write unit tests


packages/frontend/src/pages/InferenceServerDetails.svelte
```diff
@@ -74,6 +128,49 @@ $: service = $inferenceServers.find(server => server.container.containerId === c
...
+                  on:change="{onLanguageChange}"
+                  id="languages"
+                  class="border text-sm rounded-lg focus:ring-purple-500 focus:border-purple-500 block p-1 bg-charcoal-900 border-charcoal-900 placeholder-gray-700 text-white"
+                  name="languages">
+                  {#each $snippetLanguages as language}
+                    <option class="my-1" value="{language.key}">{language.label}</option>
+                  {/each}
+                </select>
+                {#if selectedVariant !== undefined}
+                  <select
```
jeffmaury, line 151: The select should be disable if the language has only one variant

jeffmaury (approved):
LGTM

packages/frontend/src/pages/InferenceServerDetails.svelte
jeffmaury: Yes

feloy:
When I try with curl, the path seems incorrect, as I get a 404. The path is `/` in the snippet, it seems it should be `/v1/chat/completions`

packages/frontend/src/pages/InferenceServerDetails.svelte
axel7083: I would rather apply the suggestion of @jeffmaury as displaying `N/A` or `Not Applicable` could be weird, when we do have a name for the variant. Showing a hint, of the library used generally.

Here is why it look likes when it is disabled when only one variant.

image

image

image

axel7083:
> When I try with curl, the path seems incorrect, as I get a 404. The path is `/` in the snippet, it seems it should be `/v1/chat/completions`

Oh nice catch @feloy, thanks for noticing. I fixed it

image

Here is what is resulting when executing the curl command

image

feloy (approved):
great!

PR #561, opened 2024-03-15
fix: scrollbar design and overflow

axel7083:
### What does this PR do?

Fix the scrollbar used in AI-Studio to match the one in Podman-Desktop. It mostly remove css class that make it being displayed when it should not.

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |
| image | image |

#### Ensuring no issue when a lot of elements are shown

image

Fixes https://github.com/projectatomic/ai-studio/issues/560

### DIFF
--- a/packages/frontend/src/app.css
+++ b/packages/frontend/src/app.css
@@ -1,3 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+:root {
+  color-scheme: dark;
+}
--- a/packages/frontend/src/app.css
+++ b/packages/frontend/src/app.css
@@ -1,7 +1,3 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
-
-:root {
-  color-scheme: dark;
-}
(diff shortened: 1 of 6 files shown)


packages/frontend/src/app.css
```diff
@@ -1,3 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+:root {
+  color-scheme: dark;
+}
```
benoitf, line 7: shouldn't podman-desktop apply that automatically ?

I mean, it may not be the case but I guess it should be the role of Podman Desktop to do that

axel7083: Yes but it is not doing it.

We are absolutely not ready to support light theme so this is currently the best solution IMO to have a _decent_ ui.

I have noticed the following but I don't seems to see it working

https://github.com/containers/podman-desktop/blob/e4fcfe5ac143ef122b767581762867997193433f/packages/preload-webview/src/webview-preload.ts#L97

benoitf: we're not setting this value for the iframe

I'll add it so we don't need it there

axel7083: css changes removed

benoitf (approved):
👍 nice

packages/frontend/src/app.css
benoitf: created https://github.com/containers/podman-desktop/pull/6444 let me if it works for you

feloy (approved):

PR #565, opened 2024-03-15
fix: Provide a landing page for the studio

### LINKED ISSUE #44: Provide a landing page for the studio
As a user, when I first open the Studio, I'm landing on a dashboard. at the moment this dashboard is empty, but we should be providing explanations about what the Studio is for and how to use it.

Maybe, we could make it load an markdown file, so the content would be easy to load.

jeffmaury:
Fixes #44

### What does this PR do?

Fill the dashboard. Uses the repo README.md but could be a specific markdown file

### Screenshot / video of UI

dashboard

#44 

### How to test this PR?

Launch Podman Desktop and click on the AI Studio in the nav bar

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "preview": "vite preview",
-    "build": "(cp ../../README.md src || copy ../../README.md src) && vite build",
+    "build": "vite build",
     "test": "vitest run --coverage",
     "test:watch": "vitest watch --coverage",
     "watch": "vite --mode development build -w"
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "preview": "vite preview",
-    "build": "vite build",
+    "build": "(cp ../../README.md src || copy ../../README.md src) && vite build",
     "test": "vitest run --coverage",
     "test:watch": "vitest watch --coverage",
     "watch": "vite --mode development build -w"
--- a/packages/frontend/.gitignore
+++ b/packages/frontend/.gitignore
@@ -1,0 +1,1 @@
+src/README.md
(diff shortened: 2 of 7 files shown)


packages/frontend/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "preview": "vite preview",
-    "build": "vite build",
+    "build": "(cp ../../README.md src || copy ../../README.md src) && vite build",
```
benoitf, line 8: do we need to copy the README file ? we can't import the file from the parent folder ?
Also if you need to copy you should probably use an npmjs package for that rather than one for windows and one for mac/linux

vrothberg (commented):
Thanks!

I think using the README is a good starting point but we need a more specific text later on.  Users visiting the upstream project and those already having AI Lab installed are at different stages in their journey.  The latter may need more guidance on how to use AI Lab rather than installation instructions.

jeffmaury:
> Thanks!
> 
> I think using the README is a good starting point but we need a more specific text later on. Users visiting the upstream project and those already having AI Lab installed are at different stages in their journey. The latter may need more guidance on how to use AI Lab rather than installation instructions.

Fully aprove that would be a follow up PR

vrothberg (commented):
Maybe I am doing something wrong, but please double check since I don't see the text displayed:

Screenshot 2024-03-18 at 09 40 05

jeffmaury:
> Maybe I am doing something wrong, but please double check since I don't see the text displayed:
> 
> Screenshot 2024-03-18 at 09 40 05

You need to run yarn build in the AI Studio folder before running it in Podman Desktop. Just realized it is missing from the README.md file, will create an issue

vrothberg (commented):
That's it, thanks!

LGTM

benoitf (approved):
approving, padding can come in a follow up PR

PR #570, opened 2024-03-18
Playground page

feloy:
### What does this PR do?

- Adds a playground page, and the possibility to send prompts and get replies from the assistant. 
- Redirects to the playground's page after creation
- displays timer 
- auto-scroll when assistant is replying

Out of scope for this PR:
- parameters
- details of queries (code snippets)
- delete button

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/9973512/65e976fc-7dfc-4b40-b334-db141ae293bb

Part of #525 

### How to test this PR?

Start one or several playgrounds, and send prompts to the different playgrounds.

Note: The model can 'core dump' if your model doesn't have enough memory for replying to several prompts at the same time.


vrothberg (commented):
This looks amazing.  Thank you, @feloy !

vrothberg:
> Note: The model can 'core dump' if your model doesn't have enough memory for replying to several prompts at the same time.

Shall we limit the number of prompts for now to prevent that?

jeffmaury:
2 remarks:

- the timing report is not here anymore, would be good to restore it
- as the height of the window is limited, you have to scroll down as soon as you are sending a second request, would be good to auto scroll automatically:

playground

Great improvment

jeffmaury (commented):
I don't understand why we need a distinction between playground and conversation but that could be a follow up PR

jeffmaury:
> I don't understand why we need a distinction between playground and conversation but that could be a follow up PR

As the server is stateless, the memory requirement on the server side is related to the size of the context window (ie the number of chat messages in a single conversation)

jeffmaury:
I found it more natural to have the timer on the upper part (next to Assistant) ? (maybe @slemeur or @ekidneyrh can help ?)
Also when the request is finished, the timer was displayed where it disappear now

feloy:
> I found it more natural to have the timer on the upper part (next to Assistant) ? (maybe @slemeur or @ekidneyrh can help ?) Also when the request is finished, the timer was displayed where it disappear now

But the timer could disappear now we are scrolling automatically

jeffmaury (approved):
LGTM

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -81,40 +82,34 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
...
-    userInput: string,
-    options?: ModelOptions,
-  ): Promise<void> {
-    const server = this.inferenceManager.get(containerId);
+  async submit(containerId: string, userInput: string, options?: ModelOptions): Promise<void> {
+    const playground = this.#playgrounds.get(containerId);
+    if (playground === undefined) throw new Error('Playground not found.');
+
+    const servers = this.inferenceManager.getServers();
+    const server = servers.find(s => s.models.map(mi => mi.id).includes(playground.modelId));
```
axel7083, line 97:
```suggestion
    const server = this.inferenceManager.get(containerId);
    if(server === undefined) {
        throw new Errro('unknown inference server');
    }
    
    if(!server.models.some(model => model.id === playground.modelId) {
        throw new Error('The requested models is not on the inference server ')
    }
```

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -81,40 +82,34 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
...
-  async submit(
-    containerId: string,
-    modelId: string,
-    conversationId: string,
-    userInput: string,
-    options?: ModelOptions,
-  ): Promise<void> {
-    const server = this.inferenceManager.get(containerId);
+  async submit(containerId: string, userInput: string, options?: ModelOptions): Promise<void> {
+    const playground = this.#playgrounds.get(containerId);
```
axel7083, line 93: Why are the conversation and playground still two separate entities ?

axel7083 (changes requested):

packages/shared/src/models/IPlaygroundMessage.ts
```diff
@@ -28,6 +28,7 @@ export interface ChatMessage {
 export interface AssistantChat extends ChatMessage {
   role: 'assistant';
   completed: boolean;
+  endTimestamp?: number;
```
axel7083, line 31: Instead of having an `endTimestamp` we could simply change the `completed` boolean field to  a number, and having completed undefined when not completed, and the end timestamp value in it when completed.

Thisn would avoid having many fields when completed and simplify the implementation

packages/frontend/src/stores/conversations.ts
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { Readable } from 'svelte/store';
+import { readable } from 'svelte/store';
+import { Messages } from '@shared/Messages';
+import { rpcBrowser, studioClient } from '/@/utils/client';
+import type { Conversation } from '@shared/src/models/IPlaygroundMessage';
+
+export const conversations: Readable<Conversation[]> = readable<Conversation[]>([], set => {
```
axel7083, line 25: Is there a reason why you did not use `RPCReadable`?

```
export const conversations = RPCReadable<Conversation[]>(
[],
[Messages.MSG_CONVERSATIONS_UPDATE],
studioClient.getPlaygroundConversations,
)
```

packages/backend/src/managers/playgroundV2Manager.ts
axel7083: Will me merged in a later refactor PR

packages/shared/src/models/IPlaygroundMessage.ts
axel7083: @feloy what do you think about this one ?

feloy: Yes, ket me change this, and the store below

packages/frontend/src/stores/conversations.ts
feloy: because of the debouncer, using the RPCReadable here breaks the user experience. It will receive the tokens only every 10s or so.

axel7083: Oh I see, i've never look up the implementation, after reading it, it makes sense.

axel7083: Please add a comment to explain it in the code

feloy:
@axel7083 it is ready for review again

axel7083:
> @axel7083 it is ready for review again

When the inference server is not yet ready and in `starting` stage (meaning the health status is not healthy yet) I am not getting anything happening when clicking on submit.

https://github.com/projectatomic/ai-studio/assets/42176370/b550855a-3f32-4e15-84e1-af66741e380f

packages/frontend/src/pages/Playground.svelte
```diff
@@ -1,5 +1,139 @@
...
+    }
+  } else if (isUserChat(message)) {
+    const msg = message as UserChat;
+    return msg.content?.split('\n') ?? [];
+  }
+  return [];
+}
+
+function askPlayground() {
+  studioClient.submitPlaygroundMessage(playgroundId, prompt);
```
axel7083, line 63: This is an async call and should be catched

feloy: good point, thanks. I've changed to catch and display the error. I've added an entry in https://github.com/projectatomic/ai-studio/issues/525 to make the experience better for the user when the inference server is not running yet

packages/frontend/src/pages/Playground.svelte
```diff
@@ -1,5 +1,151 @@
...
+    }
+  } else if (isUserChat(message)) {
+    const msg = message as UserChat;
+    return msg.content?.split('\n') ?? [];
+  }
+  return [];
+}
+
+function askPlayground() {
+  studioClient
```
axel7083, line 64: When you click on the submit, it should be displayed or show a loading state. It has a `inProgress` properties.

Also the error state should be reset when clicking again on the button

packages/frontend/src/pages/Playground.svelte
```diff
@@ -1,5 +1,151 @@
...
+        <div class="flex flex-row flex-none w-full">
+          <textarea
+            aria-label="prompt"
+            bind:value="{prompt}"
+            rows="2"
+            class="w-full p-2 outline-none text-sm bg-charcoal-800 rounded-sm text-gray-700 placeholder-gray-700"
+            placeholder="Type your prompt here"></textarea>
+
+          <div class="flex-none text-right m-4">
+            <Button disabled="{!sendEnabled}" on:click="{() => askPlayground()}">Send prompt</Button>
```
axel7083, line 145: Please a title saying why it is disabled.
image

As a user I feel frustated when I cannot click on a button and I don't know why

packages/frontend/src/pages/Playground.svelte
```diff
@@ -1,5 +1,151 @@
...
+  if (isUserChat(latest) || (isAssistantChat(latest) && isPendingChat(latest))) {
+    scrollToBottom(scrollable);
+  }
+});
+
+async function scrollToBottom(element: Element) {
+  element.scroll?.({ top: element.scrollHeight, behavior: 'smooth' });
+}
+
+function elapsedTime(msg: AssistantChat): string {
```
axel7083, line 91: The way you compute the elapsed time is leading to incoherence when the inference server is stopped/restart and we return to the page.

https://github.com/projectatomic/ai-studio/assets/42176370/3ef933ee-2a60-4c13-8b62-a07f3a2bfcc4

IMO we should remove the code about the elapsed time as this PR is big enough. And make follow up PR

axel7083: The elapsed time should be computed by the backend, and not by the frontend

packages/frontend/src/pages/Playground.svelte
```diff
@@ -1,5 +1,151 @@
...
+  type AssistantChat,
+} from '@shared/src/models/IPlaygroundMessage';
+import NavPage from '../lib/NavPage.svelte';
+import { playgrounds } from '../stores/playgrounds-v2';
+import { catalog } from '../stores/catalog';
+import Button from '../lib/button/Button.svelte';
+import { afterUpdate } from 'svelte';
+
 export let playgroundId: string;
+let prompt: string;
```
axel7083, line 19: prompt should be reset to '' when the user click on submit

image

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -81,40 +82,32 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
...
    */
-  async submit(
-    containerId: string,
-    modelId: string,
-    conversationId: string,
-    userInput: string,
-    options?: ModelOptions,
-  ): Promise<void> {
-    const server = this.inferenceManager.get(containerId);
+  async submit(playgroundId: string, userInput: string, options?: ModelOptions): Promise<void> {
```
axel7083, line 90: submit should be changed to a sync method as we should not be waiting on its result. For example

```ts

submit(playgroundId: string, userInput: string, options?: ModelOptions): void {
// in a sync maner ensure conversation exist

// do not wait for the result
process(playgroundId, userInput, options).catch(err => ...)
}

private process(playgroundId: string, userInput: string, options?: ModelOptions): Promise<void> {
// do usual logic
}
```

axel7083: Otherwise we get the timeout error while we are getting the conversations responses
image

packages/frontend/src/pages/Playground.svelte
```diff
@@ -1,5 +1,151 @@
...
+    });
+}
+
+afterUpdate(() => {
+  if (!conversation?.messages.length) {
+    return;
+  }
+  const latest = conversation.messages[conversation.messages.length - 1];
+  if (isUserChat(latest) || (isAssistantChat(latest) && isPendingChat(latest))) {
+    scrollToBottom(scrollable);
```
axel7083, line 83: I feel weird about the scroll bottom, as we force the user to stay at the bottom every time we receive a token

https://github.com/projectatomic/ai-studio/assets/42176370/26449316-eab0-4d55-8c98-a4b60431756c

packages/frontend/src/pages/Playground.svelte
jeffmaury: That could be a follow up PR

packages/frontend/src/pages/Playground.svelte
feloy: @slemeur @jeffmaury wdyt? I tried to make the scroll as least disturbing as possible (by enabling it only when the assistant is replying), but it can be frustrating for the user if he tries to see the beginning of the conversation.

axel7083: This is not a priority for now, this can be fixed in a follow up PR, the error management is more important

axel7083 (approved):

vrothberg:
I love the feature, great work!

I am currently testing and face some issues.  Somehow I end up on page with just `0`.

https://github.com/projectatomic/ai-studio/assets/11679875/7a5343d6-f8a4-4033-915d-bf989c99e32a

feloy:
> I love the feature, great work!
> 
> I am currently testing and face some issues. Somehow I end up on page with just `0`.
> 
>  Screen.Recording.2024-03-20.at.08.31.38.mov

This is what was displayed with the previous version. Are you sure, in this case, that you are currently running the correct branch, which is completely built (you need to rebuild at least from frontend, with `yarn build` either from the root of the extension or from `packages/frontend`)?

vrothberg:
Thanks!  I will try it out.  At the moment, I don't manage to get AI Lab to show up at all (latest main branch of Podman Desktop and AI Lab).  Maybe something's off on my system.  Will wait for the system update, then reboot and try again.

PR #591, opened 2024-03-20
feat: system prompt for playground page

### LINKED ISSUE #543: Add "System Prompt" into the configuration options in the playground
We would need to add a new field in the configuration of the playground environment to enable configuring a system prompt.

feloy:
### What does this PR do?

- Adds the setting block (the layout as the Recipe details in the recipe page)
- Adds the system prompt

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/9973512/b6772aea-4ba1-4ba0-bacc-a286b61011e3

Part of #525 
Fixes #543 

### How to test this PR?

Set instructions in the system prompt textarea, and check that it is used for the following prompts (NOTE: for the video above, it happened frequently that the model didn't take in consideration the system prompt for the second and following prompts)

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.ts
@@ -126,16 +126,22 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
     if (systemPrompt) {
       messages.push({ role: 'system', content: systemPrompt });
     }
-    const response = await client.chat.completions.create({
-      messages,
-      stream: true,
-      model: modelInfo.file.file,
-      ...options,
-    });
-    // process stream async
-    this.processStream(playground.id, response).catch((err: unknown) => {
-      console.error('Something went wrong while processing stream', err);
-    });
+    client.chat.completions
+      .create({
+        messages,
+        stream: true,
+        model: modelInfo.file.file,
+        ...options,
+      })
+      .then(response => {
+        // process stream async
+        this.processStream(playground.id, response).catch((err: unknown) => {
+          console.error('Something went wrong while processing stream', err);
+        });
+      })
+      .catch((err: unknown) => {
+        console.error('Something went wrong while creating model reponse', err);
+      });
   }
 
   /**
--- a/packages/backend/src/managers/playgroundV2Manager.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.ts
@@ -87,7 +87,7 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
    * @param userInput the user input
    * @param options the model configuration
    */
-  async submit(playgroundId: string, userInput: string, options?: ModelOptions): Promise<void> {
+  async submit(playgroundId: string, userInput: string, systemPrompt: string, options?: ModelOptions): Promise<void> {
     const playground = this.#playgrounds.get(playgroundId);
     if (playground === undefined) throw new Error('Playground not found.');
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -87,7 +87,7 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
    * @param userInput the user input
    * @param options the model configuration
    */
-  async submit(playgroundId: string, userInput: string, options?: ModelOptions): Promise<void> {
+  async submit(playgroundId: string, userInput: string, systemPrompt: string, options?: ModelOptions): Promise<void> {
```
axel7083, line 90: The systemPrompt is not something we can/should change once the conversation is started ? 

The conversation history should not be changed, as it would lead to incoherence if we are able to edit it while we already have existing messages.

Looking at how works ChatGPT or else, it is always defined at the begining of the conversation. In the same way that the model id is defined at the conversation creation.

The systemprompt should be provided in the creation request and added to the conversation, or it should have its own endpoint to change it, and once the conversation is started you cannot edit it. (disabling text area)

We could imagine 

````ts

setConversationSystemPrompt(conversationId: string, content: string): void {
    const conversation = this.conversations.get(conversationId);
    //ensure exists
    if(conversation .length > 1) throw new Error('cannot change on started conversation');
    // set system prompt value
}
```

feloy: It seems to me that it is not what @slemeur defined in the spec:

```
How does it work? 
- The settings are applied to the NEXT call to the model to be made and will be sent in the payload made to the model.
- That means that, each time we do a call to the model, we read the state of the settings, get the values and append that in the API call.
```

axel7083: IMO The quoted statement does not reflect how a system prompt in LLM conversation works @slemeur

slemeur: The goal is to let the user being able to iterate with the different "settings" on how the prompt is getting configured. 
- Say the user is trying with one set of settings.
- Get the answer and want to see if adjusting the settings will make the answer any better.
Hence the need to apply the settings only to the next prompt sent to the model - so the user can keep tuning;

axel7083: I think there is a misunderstood, changing a settings should only apply to the next messages, the system prompt is something that the model is receiving at the start of the conversation.

If you change the system prompt, you are changing the first message, (system prompt is simply a message, with a role `system`.) 

In chatGPT when you change a message in a conversation, **all the messages bellow are deleted** 

https://github.com/projectatomic/ai-studio/assets/42176370/cd85fdd2-e911-42b0-aa69-a3f998c7f6b6

You cannot expect in an existing covnersation to be able to change the system prompt, as the system prompt is the first message of the conversation. Otherwise you should delete all the messages already sent. 

I am not saying that a chat bot is deterministic, but at least coherent, if you have an existing conversation, and you change the system prompt, then you re-sent a message, the existing conversation is based an a message that is not here anymore. It does not make sense.

feloy: In this conversation, I have set the system prompt as 'reply in french' at the beginning, then changed to 'reply in italian' before sending the second prompt. The model replies as expected.

reply-french-italian

jeffmaury:
I can;t  make it work I have timeout errors even without the system prompt

feloy:
> I can;t make it work I have timeout errors even without the system prompt

OK, I'll look at this timeout problem

jeffmaury (approved):
LGTM

jeffmaury:
> > I can;t make it work I have timeout errors even without the system prompt
> 
> OK, I'll look at this timeout problem

I think you can forget about it as it seems that was related to the health of my Podman machine. Restarting it made it work again

feloy:
> > > I can;t make it work I have timeout errors even without the system prompt
> > 
> > 
> > OK, I'll look at this timeout problem
> 
> I think you can forget about it as it seems that was related to the health of my Podman machine. Restarting it made it work again

I also get a timeout sometimes (and @axel7083 also). I looked at the durations of the different calls, and the completions.create gets some time (generally 2s on my system, but up to 45s during my tests).

The latest commit on this PR should fix the problem

PR #596, opened 2024-03-20
Show a message when there is no playground in the list 

### LINKED ISSUE #595: Showing a message when no playground are available
image

Should display something like "There is not playground for now. Create one by clicking here" or else

> Should also fix the ghost scrollbar

cc @feloy

feloy:
### What does this PR do?

- adds a message when there is no playground in the list
- removes scrollbar when not necessary

Fixes #595


axel7083 (approved):
👍

PR #597, opened 2024-03-20
feat: checkout status check before starting application

axel7083:
### What does this PR do?

Empower the user with information on the repository status when starting an application

Scenarios

(1) repository is not cloned
- cloning repository

(2) repository is cloned and behind
- git fetch
- show warning to the user, do you want to update ?
- 'Update' => git pull
- 'Continue' => use as it is
- 'Cancel' => throw error

(3) repository is cloned and detached
- git fetch
- show warning to the user, do you want to reset ?
- 'Reset' => delete folder & git clone & git checkout
- 'Continue' => use as it is
- 'Cancel' => throw error

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/415a46df-823f-4cdb-a0c6-580cde86d696

Fixes https://github.com/projectatomic/ai-studio/issues/243

### How to test this PR?

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -27,6 +27,9 @@ import {
   type PodInfo,
   type Webview,
   type HostConfig,
+  window,
+  env,
+  Uri,
 } from '@podman-desktop/api';
 import type { AIConfig, AIConfigFile, ContainerConfig } from '../models/AIConfig';
 import { parseYamlFile } from '../models/AIConfig';
@@ -560,9 +563,47 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
     try {
       // We might already have the repository cloned
       if (fs.existsSync(gitCloneInfo.targetDirectory) && fs.statSync(gitCloneInfo.targetDirectory).isDirectory()) {
-        // Update checkout state
-        checkoutTask.name = 'Checking out repository (cached).';
-        checkoutTask.state = 'success';
+
+        const result = await this.git.isRepositoryUpToDate(
+          gitCloneInfo.targetDirectory,
+          gitCloneInfo.repository,
+          gitCloneInfo.ref,
+        );
+
+        if (result.ok) {
+          checkoutTask.name = 'Checkout repository (cached).';
+          checkoutTask.state = 'success';
+        } else {
+          const error =  `The repository "${gitCloneInfo.repository}" seems to already be cloned and is not matching the expected configuration: ${result.error}`;
+
+          if(result.updatable) {
+            const selected = await window.showWarningMessage(
+              `The repository located at "${gitCloneInfo.targetDirectory}" is not up to date. Do you want to update it ?`,
+              'Cancel',
+              'Open Folder',
+              'Yes',
+            );
+
+            switch (selected) {
+              case 'Open Folder':
+                await env.openExternal(Uri.file(gitCloneInfo.targetDirectory));
+              // eslint-disable-next-line no-fallthrough
+              case undefined:
+              case 'Cancel':
+                checkoutTask.state = 'error';
+                checkoutTask.error = error;
+                throw error;
+              case 'Continue':
+                await this.git.pull(gitCloneInfo.targetDirectory);
+                checkoutTask.state = 'success';
+                return;
+            }
+          }
+
+          checkoutTask.error = error;
+          checkoutTask.state = 'error';
+        }
+
       } else {
         // Create folder
         fs.mkdirSync(gitCloneInfo.targetDirectory, { recursive: true });
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -118,12 +118,10 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
         ref: recipe.ref,
         targetDirectory: localFolder,
       };
-      console.log('[applicationManager] before doCheckout');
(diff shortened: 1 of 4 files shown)


packages/backend/src/managers/applicationManager.ts
```diff
@@ -118,10 +118,12 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
         ref: recipe.ref,
         targetDirectory: localFolder,
       };
+      console.log('[applicationManager] before doCheckout');
```
benoitf, line 121: will be removed I suppose ?

packages/backend/src/managers/applicationManager.ts
```diff
@@ -118,10 +118,12 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
...
       await this.doCheckout(gitCloneInfo, {
         'recipe-id': recipe.id,
         'model-id': model.id,
       });
+      console.log('[applicationManager] after doCheckout');
```
benoitf, line 126: same here ?

packages/backend/src/managers/gitManager.ts
```diff
@@ -33,4 +35,137 @@ export class GitManager {
...
+        gitCloneInfo.targetDirectory,
+        gitCloneInfo.repository,
+        gitCloneInfo.ref,
+      );
+
+      if (result.ok) {
+        return;
+      }
+
+      const error = `The repository "${gitCloneInfo.repository}" seems to already be cloned and is not matching the expected configuration: ${result.error}`;
```
benoitf, line 75:
```suggestion
      const error = `The repository "${gitCloneInfo.repository}" appears to have already been cloned and does not match the expected configuration: ${result.error}`;
```

packages/backend/src/managers/gitManager.ts
```diff
@@ -33,4 +35,137 @@ export class GitManager {
...
+
+      // Ask user
+      const selected = await window.showWarningMessage(
+        `${error} By continuing, the AI application may not run as expected. `,
+        'Cancel',
+        'Continue',
+        result.updatable ? 'Update' : 'Reset',
+      );
+
+      if (selected === undefined || selected === 'Cancel') {
```
benoitf, line 85:
```suggestion
      if (!selected || selected === 'Cancel') {
```

packages/backend/src/managers/gitManager.ts
```diff
@@ -33,4 +35,137 @@ export class GitManager {
...
+
+      if (selected === 'Update') {
+        // Update
+        await this.pull(gitCloneInfo.targetDirectory);
+        return;
+      }
+
+      if (selected === 'Reset') {
+        rmSync(gitCloneInfo.targetDirectory, { recursive: true });
+      }
```
benoitf, line 101: looks like we could avoid the return instructions using `if else if else if` pattern

packages/backend/src/managers/gitManager.ts
```diff
@@ -33,4 +35,137 @@ export class GitManager {
...
+    }
+
+    if (status.deleted.length > 0) {
+      return { error: 'The local repository has created files.' };
+    }
+
+    if (status.ahead !== 0) {
+      return {
+        error: `The local repository has ${status.ahead} commit(s) ahead.`,
+      };
```
benoitf, line 151: else if

packages/backend/src/managers/gitManager.ts
```diff
@@ -33,4 +35,137 @@ export class GitManager {
...
+
+    // Ensure working with a clean branch
+    if (!status.isClean()) {
+      return { error: 'The local repository is not clean.' };
+    }
+
+    // If we are not in HEAD
+    if (status.behind !== 0) {
+      return { ok: true, updatable: true };
+    }
```
benoitf, line 169: else if ?

axel7083: yes

packages/backend/src/managers/gitManager.ts
axel7083: yes

packages/backend/src/managers/applicationManager.ts
axel7083: 👍

packages/backend/src/managers/applicationManager.ts
axel7083: 👍

benoitf (approved):

jeffmaury (commented):
I have the dialog even if I did not change the git repo neither the catalog so to me it should be updated

axel7083:
> I have the dialog even if I did not change the git repo neither the catalog so to me it should be updated

I will open an issue and investigate, I don't seems to have it on my machine

vrothberg:
I am facing the dialog at the moment and have troubles understanding what it means.  The message that the "local repository is detached" seems like an implementation detail to me that I do not understand.  I am afraid that users would be overwhelmed with this message and would always click on "Reset".

I don't have a good proposal for an alternative but wanted to highlight the challenge when being faced with the prompt.

Screenshot 2024-03-21 at 11 04 21

axel7083:
> I am facing the dialog at the moment and have troubles understanding what it means. The message that the "local repository is detached" seems like an implementation detail to me that I do not understand. I am afraid that users would be overwhelmed with this message and would always click on "Reset".
> 
> I don't have a good proposal for an alternative but wanted to highlight the challenge when being faced with the prompt.
> 
> Screenshot 2024-03-21 at 11 04 21

See for fix https://github.com/projectatomic/ai-studio/pull/612

PR #599, opened 2024-03-20
fix: use image prop from ai-studio.yaml when defined

### LINKED ISSUE #161: Add option to tag/name a resulting image when pulling a sample
user should have an option to tag/name the resulting image when pulling a sample. 
The images when built by ai studio are listed as docker.io/library/chatbot-example.

Should we add a field in the ai-yaml for image-name so it is used as default?

lstocchi:
### What does this PR do?

It adds support for the new image prop added to ai-studio.yaml file.

### Screenshot / video of UI

N/A

it resolves #161 

### How to test this PR?

1. run any recipe and check that the resulting image have the name defined in its ai-studio. You should delete your local recipe repo as it has to download the new one. If you uses an old repo nothing should change

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -1361,32 +1361,32 @@ describe('pod detection', async () => {
 
 describe('getImageTag', () => {
   const manager = new ApplicationManager(
-      '/path/to/user/dir',
-      {} as GitManager,
-      taskRegistry,
-      {
-        postMessage: mocks.postMessageMock,
-      } as unknown as Webview,
-      {
-        onPodStart: mocks.onPodStartMock,
-        onPodStop: mocks.onPodStopMock,
-        onPodRemove: mocks.onPodRemoveMock,
-        startupSubscribe: mocks.startupSubscribeMock,
-        onMachineStop: mocks.onMachineStopMock,
-      } as unknown as PodmanConnection,
-      {
-        getRecipeById: vi.fn().mockReturnValue({ name: 'MyRecipe' } as Recipe),
-      } as unknown as CatalogManager,
-      {} as ModelsManager,
-      {} as TelemetryLogger,
-      localRepositoryRegistry,
-    );
+    '/path/to/user/dir',
+    {} as GitManager,
+    taskRegistry,
+    {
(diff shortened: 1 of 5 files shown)


packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -1358,3 +1358,59 @@ describe('pod detection', async () => {
...
+      id: 'recipe1',
+    } as Recipe;
+    const container = {
+      name: 'name',
+      image: 'quay.io/repo/image',
+    } as ContainerConfig;
+    const imageTag = manager.getImageTag(recipe, container);
+    expect(imageTag).equals('quay.io/repo/image:latest');
+  });
+})
```
benoitf, line 1416: looks the file is unformatted, don't know if we need pre-commit hooks like in podman-desktop

lstocchi: yes i guess we should add it

lstocchi: opened issue https://github.com/projectatomic/ai-studio/issues/603

sallyom:
This will push the end-to-end story forward, making it possible to take the pod.yaml that ai-lab creates and `podman kube play pod.yaml` on any system. 

(as a follow-up, an option to push the ai-lab built images would be really convenient)

jeffmaury:
Update to the packaging guide is missing

lstocchi:
> Update to the packaging guide is missing

WDYM? I added the image prop description and updated the example -> https://github.com/projectatomic/ai-studio/blob/bdeceea78f90bd8769c71493368320f26b64fce8/PACKAGING-GUIDE.md

jeffmaury:
> > Update to the packaging guide is missing
> 
> WDYM? I added the image prop description and updated the example -> https://github.com/projectatomic/ai-studio/blob/bdeceea78f90bd8769c71493368320f26b64fce8/PACKAGING-GUIDE.md

My fault, sorry

jeffmaury (approved):
LGTM

lstocchi:
> (as a follow-up, an option to push the ai-lab built images would be really convenient)

@sallyom Could you please open an issue so we don't forget about it?  Thank you!!

PR #604, opened 2024-03-20
Feature/inference server using upload

axel7083:
### What does this PR do?

Using the upload system to the podman machine to improve start speed. This PR is possible as https://github.com/projectatomic/ai-studio/pull/583 allows to have a proper async creation process for the inference server

> On my computer I am going from 5 minutes to 30s to start the inference server (not counting the upload time)

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/bc3d2965-aed9-4d77-8494-106537dfdff0

Fixes https://github.com/projectatomic/ai-studio/issues/602

### How to test this PR?

### DIFF
--- a/packages/backend/src/utils/inferenceUtils.ts
+++ b/packages/backend/src/utils/inferenceUtils.ts
@@ -28,7 +28,6 @@ import {
 import type { CreationInferenceServerOptions, InferenceServerConfig } from '@shared/src/models/InferenceServerConfig';
 import { DISABLE_SELINUX_LABEL_SECURITY_OPTION } from './utils';
 import { getFreeRandomPort } from './ports';
-import { ModelInfo } from '@shared/src/models/IModelInfo';
 
 export const LABEL_INFERENCE_SERVER: string = 'ai-studio-inference-server';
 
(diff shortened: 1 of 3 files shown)


packages/backend/src/utils/inferenceUtils.ts
```diff
@@ -138,7 +138,7 @@ export function generateContainerCreateOptions(
     HealthCheck: {
       // must be the port INSIDE the container not the exposed one
       Test: ['CMD-SHELL', `curl -sSf localhost:8000/docs > /dev/null`],
-      Interval: 1_000_000_000 * 15, // 15s
+      Interval: 1_000_000_000 * 5, // 15s
```
benoitf, line 141:
```suggestion
      Interval: 1_000_000_000 * 5, // 5s
```
or maybe have a constant SECOND being 1_000_000_000 and then you use 5 * SECONDS (no need to have the comment that will be become obsolete / invalid at some point)

benoitf (approved):

packages/backend/src/utils/inferenceUtils.ts
axel7083: Yes good idea, I created as you suggest a SECOND constant and removed the comments

PR #610, opened 2024-03-20
feat(playground): adding delete conversation capabilities

axel7083:
### What does this PR do?

Adding a delete button to the conversation lists

### Screenshot / video of UI

image

Fixes https://github.com/projectatomic/ai-studio/issues/526

### How to test this PR?

### DIFF
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -72,7 +72,6 @@ export class StudioApiImpl implements StudioAPI {
       .catch((err: unknown) => {
         console.error(`Something went wrong with confirmation modals`, err);
       });
-
   }
 
   async createPlayground(name: string, model: ModelInfo): Promise<void> {
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -60,8 +60,19 @@ export class StudioApiImpl implements StudioAPI {
     private snippetManager: SnippetManager,
   ) {}
 
-  async deleteConversation(conversationId: string): Promise<void> {
-    this.playgroundV2.deletePlayground(conversationId);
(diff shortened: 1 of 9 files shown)


jeffmaury (changes requested):
Missing confirmation dialog

axel7083:
> Missing confirmation dialog

@jeffmaury confirmation dialog added

packages/backend/src/studio-api-impl.ts
```diff
@@ -60,6 +60,20 @@ export class StudioApiImpl implements StudioAPI {
     private snippetManager: SnippetManager,
   ) {}
 
+  async requestDeleteConversation(conversationId: string): Promise<void> {
+    // Do not wait on the promise as the api would probably timeout before the user answer.
+    podmanDesktopApi.window
+      .showWarningMessage(`Are you sure you want to delete this conversation ?`, 'Confirm', 'Cancel')
```
jeffmaury, line 66: Not sure why it is not playground as this is what user will see in the UI

axel7083: Because currently there is no consensus on what it is, sometime it is a conversation, sometime a playground, there are two objects that should be merged. At the end, it will be conversation because what you are deleting are the messages.

The playground is a dummy object with a duplicate of the conversation id, a name and modelId, which should be on the conversation object.

jeffmaury: My concern is that there is no mention of conversation in UI so user will be confused

axel7083, line 66:
```suggestion
      .showWarningMessage(`Are you sure you want to delete this playground ?`, 'Confirm', 'Cancel')
```

axel7083: Renamed done ✅

PR #612, opened 2024-03-20
feat: improve git detached logic

axel7083:
### What does this PR do?

When the git is detached, we need to get the current commit to check with the ref provided in the ai.json file, to check if they match. If they do, then continue without raising an error


vrothberg (commented):
Nice work!

Fixes the issue on my machine, thanks!

benoitf (approved):

Issue #624, opened 2024-03-21
change text of Dashboard


vrothberg:
At the present, the Dashboard displays the contents of the README.  It's a good starting point but we need a new document for the Dashboard as the user is a different stage in the journey.

Going through the e2e experience, the new text should cover the following:
* Welcome message
* Introduction of AI Lab
* Key features and how to get started
* Guiding the user into opening the Recipes Catalog (entry point for new users)

Screenshots would be a nice visual aid as well.

vrothberg:
Cc: @slemeur

vrothberg:
I'll give it a shot, so have an improved version before the 0.3.0 release.  It won't be perfect but something we can iterate on.  I've no idea how to have links to other menus (e.g., to the Recipe Catalog) but that's something we can worry about later on.

vrothberg:
-> https://github.com/projectatomic/ai-studio/pull/639

vrothberg:
I consider this done by https://github.com/projectatomic/ai-studio/pull/639, closing.

PR #639, opened 2024-03-22
update dashboard description

vrothberg:
### What does this PR do?

In preparation of the 0.3.0 release, I want to update the dashboard. At the present, it displays the contents of the README with installation instructions which are likely confusing users since the extension has already been installed.

The new text is not perfect and does not fully address #624 but I think it's a step in the right direction.  The descriptions will very likely need polishing in the future.  I did not find a way to create links that allow for opening another menu.  For instance, it would be nice to have clickable links/buttons that navigate users directly into the specific menu such as the Recipes Catalog.

Some graphical elements (e.g., screenshots) may be a nice additions as well.  We can add them later on, if needed, when the UI has settled a bit more.

### Screenshot / video of UI
Screenshot 2024-03-22 at 10 09 48

### DIFF
--- a/packages/frontend/src/pages/dashboard.md
+++ b/packages/frontend/src/pages/dashboard.md
@@ -1,49 +1,15 @@
-# AI studio
+# Welcome to AI Studio
 
-## Installing a released version
+The Podman Desktop AI Studio extension simplifies getting started and develop with AI in a local environment. It provides key open-source technologies to start building on AI. A curated catalog of so-called recipes helps navigate the jungle of AI use cases and AI models. AI Studio further ships playgrounds: environments to experiment with and test AI models, for instance, a chat bot.
 
-If you want to install a specific version of the extension, go to Podman Desktop UI > ⚙ Settings > Extensions > Install a new extension from OCI Image.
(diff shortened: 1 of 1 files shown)


axel7083:
I think the model serving part is important enough to have a small paragraph too, emphasising on the OpenAI like api compatible, that we often see in ~other~ products.

vrothberg:
> I think the model serving part is important enough to have a small paragraph too, emphasising on the OpenAI like api compatible, that we often see in ~other~ products.

Can you write a paragraph?  Happy to integrate it into this PR but I don't feel competent enough to write something useful about the serving part.

vrothberg:
Or we merge this PR and you can add a paragraph on model serving.

jeffmaury (approved):
LGTM except the last sentence we don't allow yet customizing the prompt with model specific parameters

vrothberg:
> LGTM except the last sentence we don't allow yet customizing the prompt with model specific parameters

Thanks. I removed this sentence.

vrothberg:
Ready to go?

benoitf (approved):

PR #642, opened 2024-03-22
fix: remove remains of playground v1

axel7083:

### DIFF
--- a/packages/shared/Messages.ts
+++ b/packages/shared/Messages.ts
@@ -17,15 +17,6 @@
  ***********************************************************************/
 
 export enum Messages {
-  /**
-   * @deprecated
-   */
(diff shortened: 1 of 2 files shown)


packages/shared/Messages.ts
```diff
@@ -17,15 +17,6 @@
  ***********************************************************************/
 
 export enum Messages {
-  /**
```
jeffmaury, line 20: Seems unrelated ?

axel7083: There are messages which where used in the Playground v1, there are now useless, as shown by the typecheck not raising any error (my editor was telling me there were unused also)

jeffmaury (approved):
LGTM

PR #643, opened 2024-03-22
ask the system prompt from the Playground creation form

### LINKED ISSUE #617: Prevent modification of the system prompt after having started the playground's discussion
Following the work on #591 we would like to improve the behavior to be more consistent with how system prompt are used.

- The system prompt should be defined ONLY before starting the conversation.
- Once the conversation started, the system prompt is not editable anymore. 
- The user will see a message saying "Modifying the system prompt will significantly impact your entire playground. Unfortunately, you have limited options to make changes once the conversation has begun. Create a new playground and adjust the system prompt prior to initiating the interaction."

feloy:
### What does this PR do?

This PR moves the system prompt textarea to the creation form of the Playground.

### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/9973512/19cbb519-42c0-4490-a2ec-96a2ea3be1c0

Fixes #617

### How to test this PR?

Create a playground with/without system prompt and check that it is displayed in the playground conversation if set.

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.ts
@@ -54,7 +54,7 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
     this.notify();
   }
 
-  async requestCreatePlayground(name: string, model: ModelInfo, systemPrompt: string): Promise<string> {
+  async requestCreatePlayground(name: string, model: ModelInfo, systemPrompt?: string): Promise<string> {
     const trackingId: string = getRandomString();
     const task = this.taskRegistry.createTask('Creating Playground environment', 'loading', {
       trackingId: trackingId,
@@ -94,7 +94,12 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
     return trackingId;
   }
 
-  async createPlayground(name: string, model: ModelInfo, systemPrompt: string, trackingId: string): Promise<string> {
+  async createPlayground(
+    name: string,
+    model: ModelInfo,
+    systemPrompt: string | undefined,
+    trackingId: string,
+  ): Promise<string> {
     const id = `${this.#playgroundCounter++}`;
 
     if (!name) {
@@ -103,7 +108,7 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
 
     this.#conversationRegistry.createConversation(id);
 
-    if (systemPrompt) {
+    if (systemPrompt !== undefined) {
       this.#conversationRegistry.submit(id, {
         content: systemPrompt,
         role: 'system',
--- a/packages/backend/src/managers/playgroundV2Manager.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.ts
@@ -22,7 +22,7 @@ import type { ChatCompletionChunk, ChatCompletionMessageParam } from 'openai/src
 import type { ModelOptions } from '@shared/src/models/IModelOptions';
 import type { Stream } from 'openai/streaming';
 import { ConversationRegistry } from '../registries/conversationRegistry';
-import type { Conversation, PendingChat, UserChat } from '@shared/src/models/IPlaygroundMessage';
+import type { Conversation, PendingChat, SystemPrompt, UserChat } from '@shared/src/models/IPlaygroundMessage';
 import type { PlaygroundV2 } from '@shared/src/models/IPlaygroundV2';
 import { Publisher } from '../utils/Publisher';
 import { Messages } from '@shared/Messages';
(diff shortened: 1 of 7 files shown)


packages/shared/src/StudioAPI.ts
```diff
@@ -143,7 +143,7 @@ export abstract class StudioAPI {
    */
   abstract createSnippet(options: RequestOptions, language: string, variant: string): Promise<string>;
 
-  abstract requestCreatePlayground(name: string, model: ModelInfo): Promise<string>;
+  abstract requestCreatePlayground(name: string, model: ModelInfo, systemPrompt: string): Promise<string>;
```
axel7083, line 146:
```suggestion
  abstract requestCreatePlayground(name: string, model: ModelInfo, systemPrompt?: string): Promise<string>;
```

packages/frontend/src/pages/PlaygroundCreate.svelte
```diff
@@ -55,7 +56,7 @@ async function submit() {
   // disable submit button
   submitted = true;
   try {
-    trackingId = await studioClient.requestCreatePlayground(playgroundName, model);
+    trackingId = await studioClient.requestCreatePlayground(playgroundName, model, systemPrompt ?? '');
```
axel7083, line 59:
```suggestion
    trackingId = await studioClient.requestCreatePlayground(playgroundName, model, systemPrompt ?? undefined);
```

packages/backend/src/studio-api-impl.ts
```diff
@@ -74,9 +74,9 @@ export class StudioApiImpl implements StudioAPI {
       });
   }
 
-  async requestCreatePlayground(name: string, model: ModelInfo): Promise<string> {
+  async requestCreatePlayground(name: string, model: ModelInfo, systemPrompt: string): Promise<string> {
```
axel7083, line 77:
```suggestion
  async requestCreatePlayground(name: string, model: ModelInfo, systemPrompt?: string): Promise<string> {
```

axel7083 (approved):
Great job ! We might want later to use something like 

```
createPlayground(options?: { system?: string, name?: string ....}) 
```

to easily add optional parameters

packages/frontend/src/pages/PlaygroundCreate.svelte
feloy: That was my first attempt, but the problem I can see is that systemPrompt can be either undefined or an empty string, depending on if the user lets the textarea untouched, or edits its content and removes it. I wanted to have only one value (the empty string) to indicate the systemPrompt is empty. But as the two values are handled correctly on the other side, we can do like this if you prefer

axel7083: I think that the empty value should not be allowed, and be replaced with `undefined`

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -103,6 +108,15 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
 
     this.#conversationRegistry.createConversation(id);
 
+    if (systemPrompt !== undefined) {
```
jeffmaury, line 111:
```suggestion
    if (systemPrompt) {
```

jeffmaury (approved):
LGTM

vrothberg (approved):
LGTM, it looks really good

slemeur:
I see this has been merged but the UX does not work for me
and I have a big -1 on this change

slemeur:
We need to agreed that the experience is primarily designed for quick iterations for a developer.

*First time user*
We are asking the user to define something that is not yet known. The change it will do to the way the model interact will not be understood AND it will not be editable "on the go"

*User interacting with the model*
As a user I've defined a system prompt, but now that I'm in the playground environment, I don't see it anymore. So if I start a playground one day and get back to it another day, I'll not know what I had defined as system prompt. 
Exact same bad UX will happen when I'm creating multiple playground environments

*User iterating customizing the use of the model*
The UX proposed and involved by this change is breaking the whole purpose of the screen. 
As a user, I want to be able to adjust and iterate into the definition of how my model must interact. It's a multi-stage purposes.
- User will start without defining a system prompt
- User will adjust the settings and add a system prompt, replay the discussion to observe how it change the behavior of the model
- User will continue to adjust the system prompt and would like to be able to compare the behavior
It's going to be an iterative process, but here, because of this UX, the user is forced to each time get back to the creation of the playground environment.

feloy:
> We need to agreed that the experience is primarily designed for quick iterations for a developer.
> 
> _First time user_ We are asking the user to define something that is not yet known. The change it will do to the way the model interact will not be understood AND it will not be editable "on the go"
> 
> _User interacting with the model_ As a user I've defined a system prompt, but now that I'm in the playground environment, I don't see it anymore. So if I start a playground one day and get back to it another day, I'll not know what I had defined as system prompt. Exact same bad UX will happen when I'm creating multiple playground environments
> 

The system prompt is visible in the chat as the first message of the discussion, with the "System Prompt" title:
 
system-propt-displayed

> _User iterating customizing the use of the model_ The UX proposed and involved by this change is breaking the whole purpose of the screen. As a user, I want to be able to adjust and iterate into the definition of how my model must interact. It's a multi-stage purposes.
> 
> * User will start without defining a system prompt
> * User will adjust the settings and add a system prompt, replay the discussion to observe how it change the behavior of the model
> * User will continue to adjust the system prompt and would like to be able to compare the behavior
>   It's going to be an iterative process, but here, because of this UX, the user is forced to each time get back to the creation of the playground environment.

slemeur:
Ok, I did not see the system prompt getting displayed at the top. Now tested, that's better.
We'll move forward with this solution for 0.3 - but will hopefully be able to refine the experience for 1.0

PR #648, opened 2024-03-25
build: add tool to compute model memory size

jeffmaury:
### What does this PR do?

Add a new tool to compute model memory size from the catalog and a workflow to launch it

### Screenshot / video of UI

N/A

Relates to #479

### How to test this PR?

Run the workflow
An example is given here: https://github.com/jeffmaury/ai-studio/actions/runs/8416856694/job/23044231069

### DIFF
--- a/tools/compute-model-sizes.sh
+++ b/tools/compute-model-sizes.sh
@@ -1,0 +1,7 @@
+echo 'FROM registry.access.redhat.com/ubi9-minimal:9.3' >Containerfile
+echo 'RUN microdnf install -y git make g++' >>Containerfile
+echo 'RUN git clone https://github.com/ggerganov/llama.cpp' >>Containerfile
+echo 'RUN cd llama.cpp && make simple' >>Containerfile
+echo 'RUN pwd' >>Containerfile
+podman build . -q -t get-model-size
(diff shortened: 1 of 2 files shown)


tools/compute-model-sizes.sh
```diff
@@ -0,0 +1,7 @@
+echo 'FROM registry.access.redhat.com/ubi9-minimal:9.3' >Containerfile
+echo 'RUN microdnf install -y git make g++' >>Containerfile
+echo 'RUN git clone https://github.com/ggerganov/llama.cpp' >>Containerfile
+echo 'RUN cd llama.cpp && make simple' >>Containerfile
+echo 'RUN pwd' >>Containerfile
```
vrothberg, line 5: Drive-by comment:
There's a more idiomatic way of creating a file than using `echo` via

```bash
cat >Containerfile<< EOF
# Contents of the file
EOF
```

This pattern is used a lot in Podman's system tests, for instance here:
https://github.com/containers/podman/blob/2aad385b51481008e337ec17950cfaace553ad79/test/system/255-auto-update.bats#L489-L496

vrothberg (commented):
LGTM

feloy (approved):

PR #650, opened 2024-03-25
feat: display model icon in icons list

feloy:
### What does this PR do?

Displays a model icon in the models list. 

The icon is green when the model is downloaded.

### Screenshot / video of UI

model-icon

Part of #479


slemeur:
I like being able to more easily see the status of the model.
This icon is placeholder? Could we take the icons from HF?

jeffmaury (approved):
LGTM. Is there a way to have a unit test (switch between model downloaded or not ?)

slemeur:
LGTM

axel7083:
shoudn't the icon be green when the model is in use ? like the volume in podman desktop ? I feel like it would be a bit confusing ?

feloy:
> shoudn't the icon be green when the model is in use ? like the volume in podman desktop ? I feel like it would be a bit confusing ?

@jeffmaury @slemeur WDYT?

jeffmaury:
> > shoudn't the icon be green when the model is in use ? like the volume in podman desktop ? I feel like it would be a bit confusing ?
> 
> @jeffmaury @slemeur WDYT?

Could be a 1.0 follow up PR

PR #654, opened 2024-03-25
fix(playground): disable the system prompt textarea when loading state

axel7083:
### What does this PR do?

Disable the system prompt textarea when loading state

Fixes https://github.com/projectatomic/ai-studio/issues/653


axel7083:
Waiting for more instructions about the system prompt position in the form, #643 could be reverted apprently

PR #656, opened 2024-03-25
fix(inference-details-ui): forcing refresh on health change

axel7083:
### What does this PR do?

The health status is a nested properties in the `InferenceServer` object, leading to miss-render on properties change. This PR is forcing the re-render using svelte `key`.

Fixes https://github.com/projectatomic/ai-studio/issues/655

### How to test this PR?

- [x] unit test provided

### DIFF
--- a/packages/frontend/src/pages/InferenceServerDetails.spec.ts
+++ b/packages/frontend/src/pages/InferenceServerDetails.spec.ts
@@ -88,7 +88,11 @@ beforeEach(() => {
 
   mocks.getInferenceServersMock.mockReturnValue([
     {
-      health: undefined,
+      health: {
+        Status: 'healthy',
+        Log: [],
+        FailingStreak: 0,
+      },
       models: [],
       connection: { port: 9999 },
       status: 'running',
@@ -135,4 +139,13 @@ test('on mount should call createSnippet', async () => {
   });
 
   expect(studioClient.createSnippet).toHaveBeenCalled();
+});
+
+test('ensure status is visible when running', async () => {
+  render(InferenceServerDetails, {
+    containerId: 'dummyContainerId',
+  });
+
+  const status = screen.getByRole('status');
+  expect(status).toBeDefined();
 });
--- a/packages/frontend/src/pages/InferenceServerDetails.svelte
+++ b/packages/frontend/src/pages/InferenceServerDetails.svelte
@@ -84,7 +84,9 @@ onMount(() => {
               <span class="text-sm">Container</span>
               <div class="w-full bg-charcoal-600 rounded-md p-2 flex items-center">
                 <div class="grow ml-2 flex flex-row">
-                  <ServiceStatus object="{service}" />
+                  {#key service.health?.Status}
+                    <ServiceStatus object="{service}" />
+                  {/key}
                   <div class="flex flex-col text-xs ml-2">
                     <span>{service.container.containerId}</span>
                   </div>


benoitf:
I think there is a missing reassignment and I would prefer to fix the missing assignment rather than adding keys as you may face issues at other places as well.

axel7083:
> I think there is a missing reassignment and I would prefer to fix the missing assignment rather than adding keys as you may face issues at other places as well.

After some investigation, I don't see where the missing re-assigment could come from. In the component here is how I am getting the service:

https://github.com/projectatomic/ai-studio/blob/1aa3f60f11fa64693ea135a0a61572a622dc484c/packages/frontend/src/pages/InferenceServerDetails.svelte#L16-L17

but still, it is not updating the UI, the store is updated, there is no doubt on that as when used in the service list, it is updated...

https://github.com/projectatomic/ai-studio/assets/42176370/50a8cc89-8931-4ccb-8946-10c253633af2

and the code in the table, do not add any logic

https://github.com/projectatomic/ai-studio/blob/1aa3f60f11fa64693ea135a0a61572a622dc484c/packages/frontend/src/pages/InferenceServers.svelte#L21

https://github.com/projectatomic/ai-studio/blob/1aa3f60f11fa64693ea135a0a61572a622dc484c/packages/frontend/src/pages/InferenceServers.svelte#L30

I do not really know why, but this nested properties does not trigger the re-render. If you have any idea, I am open because I've looked and can't find it

feloy (approved):

benoitf:
are you sure the store is updated with a new object and not a property being updated when a property is updated 

I see that ServiceStatus already has a `#key` pattern so IMHO there is still a problem.
If the object is replaced, it'll retrigger updates on the whole chain

so even in ServiceStatus it should not need a #key there
  https://github.com/projectatomic/ai-studio/blob/a93e66ee09d3d8bda3246867c6f36ede9f06b101/packages/frontend/src/lib/table/service/ServiceStatus.svelte#L31

axel7083:
> are you sure the store is updated with a new object and not a property being updated when a property is updated
> 
> I see that ServiceStatus already has a `#key` pattern so IMHO there is still a problem. If the object is replaced, it'll retrigger updates on the whole chain
> 
> so even in ServiceStatus it should not need a #key there
> 
> https://github.com/projectatomic/ai-studio/blob/a93e66ee09d3d8bda3246867c6f36ede9f06b101/packages/frontend/src/lib/table/service/ServiceStatus.svelte#L31

Yes the store is updated, otherwise the `#key` would not have any impact. This is what the store look like with one item:

```json
[
  {
    "container": {
      "containerId": "e9aaa99abf944a62643dd7dfbc6cb1c9d401c9942f436edc7972be9b79358373",
      "engineId": "podman.Podman Machine"
    },
    "connection": {
      "port": 52918
    },
    "status": "running",
    "models": [
      {
        "id": "hf.TheBloke.mistral-7b-instruct-v0.1.Q4_K_M",
        "name": "TheBloke/Mistral-7B-Instruct-v0.1-GGUF",
        "description": "The Mistral-7B-Instruct-v0.1 Large Language Model (LLM) is a instruct fine-tuned version of the [Mistral-7B-v0.1](https://huggingface.co/mistralai/Mistral-7B-v0.1) generative text model using a variety of publicly available conversation datasets. For full details of this model please read our [release blog post](https://mistral.ai/news/announcing-mistral-7b/)",
        "hw": "CPU",
        "registry": "Hugging Face",
        "license": "Apache-2.0",
        "url": "https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf",
        "file": {
          "file": "mistral-7b-instruct-v0.1.Q4_K_M.gguf",
          "path": "C:\\Users\\axels\\podman-desktop\\ai-studio\\models\\hf.TheBloke.mistral-7b-instruct-v0.1.Q4_K_M",
          "size": 4368438944,
          "creation": "2024-03-25T14:22:42.906Z"
        }
      }
    ],
    "health": {
      "Status": "healthy",
      "FailingStreak": 0,
      "Log": [
        {
          "Start": "2024-03-26T11:03:21.2971906+01:00",
          "End": "2024-03-26T11:03:21.407583699+01:00",
          "ExitCode": 0,
          "Output": ""
        },
        {
          "Start": "2024-03-26T11:03:26.772686014+01:00",
          "End": "2024-03-26T11:03:26.84352236+01:00",
          "ExitCode": 0,
          "Output": ""
        },
        {
          "Start": "2024-03-26T11:03:32.303293441+01:00",
          "End": "2024-03-26T11:03:32.438443661+01:00",
          "ExitCode": 0,
          "Output": ""
        },
        {
          "Start": "2024-03-26T11:03:38.301139561+01:00",
          "End": "2024-03-26T11:03:38.396622913+01:00",
          "ExitCode": 0,
          "Output": ""
        },
        {
          "Start": "2024-03-26T11:03:44.299211874+01:00",
          "End": "2024-03-26T11:03:44.420580661+01:00",
          "ExitCode": 0,
          "Output": ""
        }
      ]
    }
  }
]
```

IMO the health property `service.health.status` is too nested for it tricker the re-render, as it is the only properties that is updated in the store, all the others are the same, maybe I am wrong, but the store is updated, and using key on nested properties makes it re-render, so IMO it is a svelte issue

benoitf:
I won't have time to investigate pretty soon so I don't want to hold the merge as well

I still believe the object is not somehow updated properly but finding will take time and at the end I can be wrong :-)

PR #674, opened 2024-03-25
add models memory footprint to catalog

feloy:
### What does this PR do?

Adds memory footprint of models, as computed by the worflow `.github/workflows/compute-model-sizes.yaml`

Part of #479

### DIFF
--- a/packages/backend/src/assets/ai.json
+++ b/packages/backend/src/assets/ai.json
@@ -84,7 +84,7 @@
       "registry": "Hugging Face",
       "license": "Apache-2.0",
       "url": "https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf",
-      "memory": 4268032
+      "memory": 4370464768
     },
     {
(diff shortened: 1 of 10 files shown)


feloy:
I forgot a factor of 1024 for values, I'll change ...

axel7083 (approved):
LGTM

packages/backend/src/assets/ai.json
```diff
@@ -83,7 +83,8 @@
       "hw": "CPU",
       "registry": "Hugging Face",
       "license": "Apache-2.0",
-      "url": "https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf"
+      "url": "https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf",
+      "memory": 4370464768
```
jeffmaury, line 87: 4,07 * 1024 * 1024 * 1024 = 4 370 129 224 after rounding on my side

feloy: oh right, I've first rounded at the MiB, then converted to bytes. That should be ok now. Thanks

jeffmaury (approved):
LGTM

Issue #551, opened 2024-03-14
Add breadcrumbs for consistent navigation patterns


slemeur:
In Podman Desktop, when you have a list and get into the details of one of its item, you get a breadcrumb that facilitates the navigation.

<img width="477" alt="Screenshot 2024-03-14 at 16 37 37" src="https://github.com/projectatomic/ai-studio/assets/1636769/0c272e51-57f1-4dde-a86a-c2f513873fee">

We are missing it for Recipes:
<img width="569" alt="Screenshot 2024-03-14 at 16 37 52" src="https://github.com/projectatomic/ai-studio/assets/1636769/989ffeb5-b436-4872-b981-26f8b192a844">

We are missing it for Models:
<img width="397" alt="Screenshot 2024-03-14 at 16 38 38" src="https://github.com/projectatomic/ai-studio/assets/1636769/cdf4c21e-a86a-4b02-b7c6-4c577aa454be">

We are missing it for Services too.

vrothberg:
@feloy can we close this given #697 merged?

feloy:
fixed by #697

PR #697, opened 2024-03-26
feat: adding breadcrumbs

axel7083:
### What does this PR do?

Adding Breadcrumbs in a similar manner than Podman-Desktop.

> Implementation details: Podman-Desktop use a dedicated store to store the last page, however the webview of AI-Studio can and will be destoryed, re-created many time. making it complicated to save the last page nor relevant with the current UI. In this PR, for each page, we can optionally define a `lastpage` variable, that will display the breadcrumb corresponding.

### Screenshot / video of UI

**Model Service Details**

image

**Create Model Service**

image

**Create Playground**

image

**Playground Details**

image

**Recipe details**

image

**model details**

image

Fixes https://github.com/projectatomic/ai-studio/issues/696

### How to test this PR?

Navigate in the pages of AI-Studio

### DIFF
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -53,7 +53,7 @@ onMount(() => {
       </Route>
 
       <!-- Playgrounds -->
-      <Route path="/playgrounds" >
+      <Route path="/playgrounds">
         <Playgrounds />
       </Route>
       <Route path="/playground/:id/*" let:meta>
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -32,31 +32,31 @@ onMount(() => {
 });
 </script>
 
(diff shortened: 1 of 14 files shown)


feloy:
Also fixes #551 ?

feloy:
> **Model Service Details**
> 
> image
> 

It would be nice to have the name of the model, instead of 'Service Details'


Also, the Playground list page displays a breadcrumb `Playgrounds > Playgrounds environments`, which is not consistent with other pages at the same level, where there is no breadcrumb.

axel7083:
> > **Model Service Details**
> > image
> 
> It would be nice to have the name of the model, instead of 'Service Details'
> 
> Also, the Playground list page displays a breadcrumb `Playgrounds > Playgrounds environments`, which is not consistent with other pages at the same level, where there is no breadcrumb.

I agree, but this should be made in a follow up PR as it changes the name of the pages properties.

jeffmaury (approved):
LGTM

PR #703, opened 2024-03-26
fix: do not handle pods without ai-studio labels

### LINKED ISSUE #702: A pod created is added to the list of AI Apps
https://github.com/projectatomic/ai-studio/assets/9973512/44ab8736-6759-4ede-b349-db0b6a474c9d

feloy:

Fixes #702 

### How to test this PR?

- Create a pod from podman desktop and check that it does not appear in the list of AI Apps
- Create AI Apps and check they appear in the list


axel7083 (approved):
Oh that was a big bug, nice having it fixed ! LGTM

lstocchi (approved):
LGTM. Tested and works fine 🚀

vrothberg (commented):
Should the label now be `ai-lab`?

feloy:
> Should the label now be `ai-lab`?

Yes, the constants `LABEL_RECIPE_ID` and similar will need to be renamed to contain `ai-lab` instead of `ai-studio`. I can see it is done as part of #713

PR #713, opened 2024-03-27
fix: rename Studio to Lab

lstocchi:
### What does this PR do?

It replaces the old ai-studio with the new ai-lab naming

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

it resolves #571 

### How to test this PR?

1. it should work as before

PR #715, opened 2024-03-27
feat: check ai-app pod healthcheck

feloy:
### What does this PR do?

Check the healthcheck of the AI apps' pods and adds a `health` field to the data sent to the front. 

Part of #265

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -118,6 +118,9 @@ vi.mock('@podman-desktop/api', () => ({
     stopPod: mocks.stopPodMock,
     removePod: mocks.removePodMock,
   },
+  Disposable: {
+    create: vi.fn(),
+  },
 }));
 
 const telemetryLogger = {
@@ -934,6 +937,12 @@ describe('createAndAddContainersToPod', () => {
       start: false,
       name: 'name',
       pod: 'id',
+      HealthCheck: {
+        Interval: 5000000000,
+        Retries: 20,
+        Test: ['CMD-SHELL', 'curl -s localhost:8080 > /dev/null'],
+        Timeout: 2000000000,
+      },
     });
     expect(mocks.createContainerMock).toHaveBeenNthCalledWith(2, 'engine', {
       Image: 'id2',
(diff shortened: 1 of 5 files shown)


packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -1268,4 +1284,58 @@ describe('getImageTag', () => {
...
     expect(imageTag).equals('quay.io/repo/image:latest');
   });
+
+  test('adoptRunningApplications should check pods health', async () => {
+    mocks.listPodsMock.mockResolvedValue([
+      {
+        Id: 'pod1',
+        engineId: 'engine1',
+        Labels: {
+          'ai-studio-recipe-id': 'recipe-id-1',
```
feloy, line 1294: This will need to be changed when #713 is merged

axel7083:
I've noticed pretty recently than podman desktop is sending health events. 

image

I used the same inspect technique with the Inference Manager, and at some point, I am wondering if we should rely on our own pulling update with the inspect mecanism, or wait for podman desktop update maybe @benoitf have an opinion ?

From my understanding, when adding an health check to a container, podman is creating some kind of background services[^1], that will execute the health command on our container, and it seems that this is triggering an event.

I do not think that should be a blocker for this PR, but might be a good way to simplify our code, both in the Inference Manager, and the application manager in the future

[^1]: https://developers.redhat.com/blog/2019/04/18/monitoring-container-vitality-and-availability-with-podman#what_are_healthchecks_

packages/backend/src/managers/applicationManager.ts
```diff
@@ -701,6 +734,39 @@ export class ApplicationManager extends Publisher<ApplicationState[]> {
...
+      );
+      const healthyPod = !containerStates.find(
+        s => s.Status !== 'running' || (s.Health.Status !== '' && s.Health.Status !== 'healthy'),
+      );
+      const recipeId = pod.Labels[LABEL_RECIPE_ID];
+      const modelId = pod.Labels[LABEL_MODEL_ID];
+      if (!this.#applications.has({ recipeId, modelId })) {
+        // a fresh pod could not have been added yet, we will handle it at next iteration
+        continue;
+      }
```
lstocchi, line 756: Should this be moved before inspecting all containers?

lstocchi (approved):
LGTM

PR #716, opened 2024-03-27
feat: adding hidden utility class to experiment with GPU

axel7083:
### What does this PR do?

This PR is adding an hidden, experimental manager called `GPUManager`. This class is responsible to detect if GPUs are available on the host machine.

We cannot rely on host configuration, and need to create a container, mounting properly the expected devices, especially in WSL, then to detect the gpus a container can access to. 

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/60d48b77-9a16-4052-a76d-4417727617b2

### How to test this PR?

> hint: run `podman pull nvcr.io/nvidia/cuda:12.3.2-devel-ubi8` before, as this can take a looooot of time.

- (1) Open the command palette (F1)
- (2) Select the `AI-Lab: Collect GPU(s)`.
- (3) Wait for the notification popup

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -33,13 +33,7 @@
           "icon": "${brain-icon}"
         }
       ]
-    },
-    "commands": [
-      {
-        "command": "ai-lab.gpu.collect",
-        "title": "AI-Lab: Collect GPU(s)"
-      }
-    ]
+    }
   },
   "scripts": {
     "build": "vite build",
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -33,7 +33,13 @@
           "icon": "${brain-icon}"
         }
       ]
-    }
+    },
+    "commands": [
+      {
+        "command": "ai-lab.gpu.collect",
+        "title": "AI-Lab: Collect GPU(s)"
+      }
+    ]
   },
   "scripts": {
     "build": "vite build",
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -44,6 +44,7 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
+    "fast-xml-parser": "^4.3.6",
     "mustache": "^4.2.0",
     "openai": "^4.29.2",
     "postman-code-generators": "^1.9.0",
--- a/packages/backend/src/managers/GPUManager.spec.ts
+++ b/packages/backend/src/managers/GPUManager.spec.ts
@@ -16,9 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 import { expect, test, vi, beforeEach } from 'vitest';
-import {
-  containerEngine,
-  env} from '@podman-desktop/api';
+import { containerEngine, env } from '@podman-desktop/api';
 import type {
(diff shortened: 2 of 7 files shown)


lstocchi (commented):
I just tried and it's not able to give me an answer. I don't have an nvidia gpu so i would expect the "No GPU was found on this machine" message but it error with timeout.
I'll try to investigate, maybe if you don't have an nvidia gpu it doesn't work? Is it fast on your machine?

axel7083:
> I just tried and it's not able to give me an answer. I don't have an nvidia gpu so i would expect the "No GPU was found on this machine" message but it error with timeout. I'll try to investigate, maybe if you don't have an nvidia gpu it doesn't work? Is it fast on your machine?

I investigate, and tried on my laptop with a Intel GPU, and the issue is that I am not detecting that the container is exiting with non-zero code. I will work on improving this scenario.

axel7083:
> I just tried and it's not able to give me an answer. I don't have an nvidia gpu so i would expect the "No GPU was found on this machine" message but it error with timeout. I'll try to investigate, maybe if you don't have an nvidia gpu it doesn't work? Is it fast on your machine?

I upda

benoitf:
should it be moved as draft ?

axel7083:
> should it be moved as draft ?

Why ? I would rather have it merge, than keeping it draft for weeks, then have to deal with tons of conflicts later. 

> The close/re-open is a miss-click on my side. I wanted to sent the message but click on the wrong button :/

lstocchi (commented):
Tried again and the response is `nvidia CUDA Container exited with code 127` so now i have a feedback 👍 
I can just verify the code. @feloy @jeffmaury do you have an nvidia on your win machine?

feloy:
> Tried again and the response is `nvidia CUDA Container exited with code 127` so now i have a feedback 👍 I can just verify the code. @feloy @jeffmaury do you have an nvidia on your win machine?

No, I don't have.

What is the expected result on Mac, is it worth testing on mac?

axel7083:
> > Tried again and the response is `nvidia CUDA Container exited with code 127` so now i have a feedback 👍 I can just verify the code. @feloy @jeffmaury do you have an nvidia on your win machine?
> 
> No, I don't have.
> 
> What is the expected result on Mac, is it worth testing on mac?

No on mac, you should have an error saying you are not a windows machine. So there is no interest for this platform.

lstocchi (approved):
BTW code LGTM

axel7083:
@lstocchi @feloy do you think we should keep the command ? or delete it to keep it hidden ? 🤔

lstocchi:
> @lstocchi @feloy do you think we should keep the command ? or delete it to keep it hidden ? 🤔

I recalled we had something like a when clause for commands (when: false) or something similar to hide a command from the command palette? But i cannot find it so maybe i'm confused with something else. @benoitf do we have such a prop for commands?

benoitf:
https://github.com/containers/podman-desktop/issues/4107

axel7083:
I don't really like hidding it like this, so I'm removing the command for now. We might have later some kind of experimental flag.

feloy (approved):

Issue #724, opened 2024-03-28
Creating CancellationTokenRegistry


axel7083:
Required for the following issues

- https://github.com/containers/podman-desktop-extension-ai-lab/issues/520
- https://github.com/containers/podman-desktop-extension-ai-lab/issues/467

PR #725, opened 2024-03-28
feat: introducing CancellationTokenRegistry

axel7083:
### What does this PR do?

This PR is adding a _very similar_ CancellationTokenRegistry. It is using the class CancellationTokenSource from the api to do so.

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/724

### How to test this PR?


lstocchi (commented):
Missing a function to create a new cancel token?
i see you just created one to ask the cancellation of an existing token

packages/backend/src/registries/CancellationTokenRegistry.ts
```diff
@@ -0,0 +1,59 @@
...
+export class CancellationTokenRegistry implements Disposable {
+  #callbackId: number;
+  #callbacksCancellableToken: Map<number, ICancellationTokenSource>;
+
+  constructor() {
+    this.#callbackId = 0;
+    this.#callbacksCancellableToken = new Map<number, ICancellationTokenSource>();
+  }
+
+  createCancellationTokenSource(): number {
```
axel7083, line 30: @lstocchi

axel7083: This are the same methods use in PodmanDesktop, I wanted to stay close to their implementation

lstocchi: I express myself badly. I was referring to the studioImpl. You only exposed the cancelToken but not the create. I would expect a `requestNewToken  `https://github.com/containers/podman-desktop-extension-ai-lab/pull/725/files#diff-3f3c50af66ddba22163580bcd3807fbb24558686b62bebd09e120bc6be130a76R394-R398

axel7083: IMO the frontend should not be responsible of creating those token, but should be attached to a task. 

Let's take the download model example, the `ModelManager` will create a cancellation token, set to the download task a `cancellationId` that the frontend will be able to see. But there is no reason why the front should create those tokens

lstocchi: I'm not so convinced but we can always add it later 👍

lstocchi (approved):
LGTM

benoitf (commented):
AFAIK `CancellationTokenSource` should be available from the extension API

(CancellationToken, CancellationTokenSource)

benoitf (commented):
basically CancellationTokenSource should be exposed as a class, not an interface from the Podman Desktop API

axel7083:
> AFAIK `CancellationTokenSource` should be available from the extension API
> 
> (CancellationToken, CancellationTokenSource)

It is available, but as I mention, as an interface not a class, is this what you are pointing out ? It should be available as a class ?

axel7083:
> basically CancellationTokenSource should be exposed as a class, not an interface from the Podman Desktop API

Okey, I will open a ticket on the Podman Desktop side, but for our timeframe and the version we support, I guess we cannot rely on the api to handle this for us

benoitf:
> Okey, I will open a ticket on the Podman Desktop side, but for our timeframe and the version we support, I guess we cannot rely on the api to handle this for us

I don't know for timeframe, there is a release planned April 2nd (1.9.0 of Podman Desktop)

axel7083:
> > Okey, I will open a ticket on the Podman Desktop side, but for our timeframe and the version we support, I guess we cannot rely on the api to handle this for us
> 
> I don't know for timeframe, there is a release planned April 2nd (1.9.0 of Podman Desktop)

@slemeur do we want to make `AI-Lab` only compatible with 1.9 ?

benoitf (approved):

PR #733, opened 2024-03-28
feat: disable send prompt when model not served + display status

### LINKED ISSUE #646: Wait for the inference server to be running before to activate the "Send Prompt" (and display inference server in playground page?)

feloy:
### What does this PR do?

- in the playground page, displays the status of the model server. 
-  if the server model is not running/healthy, the 'send prompt' button is disabled and the tooltip indicates the status of the server.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/9973512/307d806a-0021-495a-8063-bf4a0c26a03d

Fixes #646

### How to test this PR?

- Start a playground on a model with no server running and check the model server is marked as not healthy
- stop a model server, and check that is is displayed as not running on a related playground

### DIFF
--- a/packages/frontend/src/pages/Playground.svelte
+++ b/packages/frontend/src/pages/Playground.svelte
@@ -13,6 +13,7 @@ import Fa from 'svelte-fa';
 
 import ChatMessage from '../lib/conversation/ChatMessage.svelte';
 import SystemPromptBanner from '/@/lib/conversation/SystemPromptBanner.svelte';
+import { inferenceServers } from '/@/stores/inferenceServers';
 
 export let playgroundId: string;
 let prompt: string;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/pages/Playground.svelte
```diff
@@ -71,6 +72,37 @@ afterUpdate(() => {
...
+  switch (status) {
+    case 'running':
+      switch (health) {
+        case 'healthy':
+          return 'Model Service healthy';
+        default:
+          return 'Model Service not healthy';
+      }
+    default:
+      return 'Model Service not running';
```
lstocchi, line 94: From a user perspective idk if it makes sense to use the term healthy. Maybe we can just show `Model Service running` - `Model Service not running` and add further explanation/hints in the tooltip (if we can have any).

lstocchi (approved):
LGTM, just a thought on the text shown.

packages/frontend/src/pages/Playground.svelte
feloy: Yes, you are right. I also introduced a "Model Service starting" state, as discussed for the AI App today during UX call

lstocchi (approved):

slemeur:
Oh yeah, this one is very nice!

PR #734, opened 2024-03-28
feat: more details for pod health

feloy:
### What does this PR do?

Provides a more precise value for the pod health ('', healthy, unhealthy, starting which are the same values as for the container health).

Part of #265

### DIFF
--- a/packages/shared/src/models/IApplicationState.ts
+++ b/packages/shared/src/models/IApplicationState.ts
@@ -18,7 +18,7 @@
 
 import type { PodInfo } from '@podman-desktop/api';
 
-export type PodHealth = '' | 'starting' | 'healthy' | 'unhealthy';
+export type PodHealth = 'none' | 'starting' | 'healthy' | 'unhealthy';
 
 export interface ApplicationState {
   recipeId: string;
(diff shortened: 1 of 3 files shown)


packages/shared/src/models/IApplicationState.ts
```diff
@@ -18,11 +18,13 @@
 
 import type { PodInfo } from '@podman-desktop/api';
 
+export type PodHealth = '' | 'starting' | 'healthy' | 'unhealthy';
```
axel7083, line 21: Why are you using '' it is not a valid value ? if we want to know when we do not have this value, we should use health?: PodHealth bellow

feloy: the empty string is the value returned by Inspect when a container has no healthcheck configured (or 'none' as documented in https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/ContainerInspect)

My idea is to have the same value as for a container

axel7083: I understand, as much as I hate to see there api to provide that value, your implementation make sense with it thanks for pointing this out

feloy: I can replace the empty string with the `'none'` one to completely stick with the documented values

axel7083: I also prefer none because this seems to be also what podman return https://github.com/containers/podman/blob/8ec821452581120b73625daff68b5eb4829f44f6/pkg/api/server/register_containers.go#L79

axel7083 (approved):

Issue #735, opened 2024-03-28
feat(playground): regenerate assistant message


axel7083:
When changing the parameters, we should be able to regenerate the assistant responses using new parameters.

hint: all message following it including itself should be deleted.

PR #736, opened 2024-03-28
feat: adding regenerate button for assistant chats

axel7083:
### What does this PR do?

Adding a `regenerate` button bellow the assistant message. Allowing to regenerate an assistant message using current parameters.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/a0727f69-c3c0-4780-b987-c5b3814630e7

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/735

### How to test this PR?

**manually**

- Start a playground
- Ask stuff to the assistant
- Use the regenerate button

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.spec.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.spec.ts
@@ -485,6 +485,57 @@ test('requestCreatePlayground should call createPlayground and createTask, then 
   });
 });
 
+describe('regenerate', () => {
+  test('when regenerate called the system message should be deleted', async () => {
+    vi.mocked(inferenceManagerMock.getServers).mockReturnValue([
+      {
+        status: 'running',
+        health: {
+          Status: 'healthy',
+        },
+        models: [
+          {
+            id: 'dummyModelId',
+            file: {
+              file: 'dummyModelFile',
+            },
+          },
+        ],
+        connection: {
+          port: 8888,
+        },
+      } as unknown as InferenceServer,
+    ]);
+    const createMock = vi.fn().mockResolvedValue([]);
+    vi.mocked(OpenAI).mockReturnValue({
+      chat: {
+        completions: {
+          create: createMock,
+        },
+      },
+    } as unknown as OpenAI);
+
+    const manager = new PlaygroundV2Manager(webviewMock, inferenceManagerMock, taskRegistryMock);
+    await manager.createPlayground('playground 1', { id: 'dummyModelId' } as ModelInfo, undefined, 'tracking-1');
+
+    const date = new Date(2000, 1, 1, 13);
+    vi.setSystemTime(date);
+
+    const playgrounds = manager.getPlaygrounds();
+    await manager.submit(playgrounds[0].id, 'dummyUserInput');
+
+    // Wait for assistant message to be completed
+    await vi.waitFor(() => {
+      expect(manager.getConversations()[0].messages[1].content).toBeDefined();
+    });
+
+    const messageId = manager.getConversations()[0].messages[1].id;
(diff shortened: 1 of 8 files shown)


slemeur:
A couple of feedback:
- Could we display the "regenerate button" only when the settings have been changed?
- I think it would be nice to display a message to the user, than once they change the settings, they need to click on the "regenerate" icon, if they want to apply the settings at a specific point of the conversation. WDYT?
- A tooltip explaining what the icon means could be nice

axel7083:
> A couple of feedback:
> 
> * Could we display the "regenerate button" only when the settings have been changed?

LLM are not deterministic, they have the ability to give a different response with the same parameters. E.g. ChatGPT.

> * I think it would be nice to display a message to the user, than once they change the settings, they need to click on the "regenerate" icon, if they want to apply the settings at a specific point of the conversation. WDYT?

This would be confusing, as after changing the parameters the user has two action possible: send a new message or regenerate an existing one. Other chat bot don't do that, so let's keep it simple 

> * A tooltip explaining what the icon means could be nice

There is :) !
image

slemeur:
OK ! LGTM

benoitf:
I would expect regenerate on a card regenerate the content only in that box/card but keep the other boxes

so in your example it should keep the "Tell me short things" question and reply

lstocchi (approved):
Super cool feature. LGTM 🚀

axel7083:
> I would expect regenerate on a card regenerate the content only in that box/card but keep the other boxes
> 
> so in your example it should keep the "Tell me short things" question and reply

Regenerating a message at the top of the conversation should, and will delete all the message after this one. As all subsequence messages are based on the previous, it makes no sense to keep them. This is how all the ChatBot are working today (E.g. ChatGPT). We can think the regenerate as a fork, branching out of the existing conversation to a new one.

> I would be disappointed if by clicking on regenerate at the top card I'm losing all the other questions I asked (without knowing it'll drop all the history)
> 
> (or it creates a new card at the bottom: like more a 'ask/query again') if you can't update the content of a given card

I understand the possible frustration, but as this is a _common_ feature in other chatbot, I feel adding some kind of warning would add too much sludge to the action that should be fast IMO.

benoitf:
> This is how all the ChatBot are working today (E.g. ChatGPT).

AFAIK ChatGPT regenerate icon is only possible on the last call
image


And then if I edit a question, I still have access to all my previous questions using the prev/next
image

image

image

benoitf:
And when I click on regenerate on the last card, I still have all my previous cards staying

axel7083:
> > This is how all the ChatBot are working today (E.g. ChatGPT).
> 
> AFAIK ChatGPT regenerate icon is only possible on the last call 

I am surprised about that, my bad I thought it was available for all.

> And then if I edit a question, I still have access to all my previous questions using the prev/next

The edit would be the next PR, we do not support that yet, but yes the behaviour would be hopefully something similar 

--- 

If we want to make something similar with the history, we would need to change the representation of a conversation from an array of message to tree with branches. 

I would agree to only show the regenerate button on the latest message even tho I still think it would make some kind of sense to regenerate any of the assistant message.

benoitf:
Also I would avoid the term "a common feature in other chatbot" or "how all the ChatBot are working" as IMHO this is a personal interpretation b/c ChatGPT has the icon only on the last card, and even editing questions and having new history I have access to all questions I asked before.

If you want to introduce regenerate, if you're not able to keep/have a way to browse the history, then regenerate should only be available on the last question as it will keep all the previous questions

benoitf:
> I would agree to only show the regenerate button on the latest message even tho I still think it would make some kind of sense to regenerate any of the assistant message.

I think it's a good tradeoff until the graph is implemented

PR #742, opened 2024-03-29
feat: display ai app pod's status using health

feloy:
### What does this PR do?

- Use the health of the pod in addition to the general status to display more accurately the status of an AI App, in both Recipe details and in AI Apps list.
- In the AI Apps list, removes the Status column displaying the latest task, and move it to the left with the icon indicating the status 

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/9973512/73681d3a-603c-4e4c-82f5-01bfc93a7880

Part of #265 

### How to test this PR?

Start a recipe and check the status is correct


packages/frontend/src/pages/applications.ts
```diff
@@ -17,13 +17,38 @@
...
 
-export interface ApplicationCell {
-  tasks?: Task[];
-  appState: ApplicationState;
-  recipeId: string;
-  modelId: string;
-  appPorts: number[];
-  modelPorts: number[];
+/* returns the status of the AI application, to be used by <IconStatus> */
+export function getApplicationStatus(
```
benoitf, line 22: should we have a new type being introduced for the return type

`'RUNNING' | 'DEGRADED' | 'STARTING' | 'USED' | 'DELETING' | ''`

also instead of '' should it be `UNKNOWN` (or another value)

IMHO it looks strange to have an empty state

feloy: These values are the ones accepted by the component StatusIcon. I think we would need to define a type on the component first.

I'm using the empty string, as it is the default value for `status` of the StatusIcon component.

feloy: I have made some changes in the latest commit.

I'm not sure if declaring this has a sense? (we want to accept all values, but some of them are "privileged", as they will have specific effect on the component)
```
export let status: knownStatus | string = '';
```

packages/frontend/src/lib/StatusIcon.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export type knownStatus = 'RUNNING' | 'STARTING' | 'USED' | 'CREATED' | 'DELETING' | 'DEGRADED';
```
benoitf, line 19:
```suggestion
export type KnownStatus = 'RUNNING' | 'STARTING' | 'USED' | 'CREATED' | 'DELETING' | 'DEGRADED';
```

I would use an uppercase for the first letter

benoitf (approved):

packages/frontend/src/lib/StatusIcon.svelte
```diff
@@ -1,9 +1,10 @@
 <script lang="ts">
+import type { knownStatus } from './StatusIcon';
 // status: one of RUNNING, STARTING, USED, CREATED, DELETING, or DEGRADED
 // any other status will result in a standard outlined box
 import Spinner from '/@/lib/button/Spinner.svelte';
 
-export let status = '';
+export let status: knownStatus | string = '';
```
benoitf, line 7: not sure why we need the "else string".

packages/frontend/src/pages/applications.ts
benoitf: well "anything that is a string" | string is just a string

it's like `"foo" | string | any` at the end it's any that is winning

so IMHO either you put 'KnownStatus" or just string 

using KnownStatus | string people will never know that there are like "privilege" components

or you have a component like `<KnownStatusComponent>` accepting only `KnownStatus` but that will use `<StatusComponent>` "behind the scene"

packages/frontend/src/lib/StatusIcon.svelte
feloy: the doc indicates that the component can accept any other value:
```
// any other status will result in a standard outlined box
```

axel7083 (commented):
I'm stuck on loading spinner

image


https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/a1b084c9-433a-424a-8cc0-5b9cc0b0b51f

The app is working in the browser, seems all healthy, but still loading state endlessly

feloy:
> I'm stuck on loading spinner
> 

This is happening every time you start/restart an app, or is it something flaky?

axel7083:
> > I'm stuck on loading spinner
> 
> This is happening every time you start/restart an app, or is it something flaky?

Everytime

feloy:
> > > I'm stuck on loading spinner
> > 
> > 
> > This is happening every time you start/restart an app, or is it something flaky?
> 
> Everytime

Is it also happening in the Recipe page, in the AI App details?

axel7083:
Here is a full video.
- I checkout latest 

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/cc970bd7-a79a-4824-b6c8-b34db0f6dab1

The error in the log

image

axel7083 (changes requested):
Introduced by #715 an issue with health checks (see https://github.com/containers/podman-desktop-extension-ai-lab/pull/742#issuecomment-2027314597) it should probably be better to fix it before merging this

feloy:
> Introduced by #715 an issue with health checks (see [#742 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/742#issuecomment-2027314597)) it should probably be better to fix it before merging this

OK, thanks, I'll work on this issue first

axel7083:
> > Introduced by #715 an issue with health checks (see [#742 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/742#issuecomment-2027314597)) it should probably be better to fix it before merging this
> 
> OK, thanks, I'll work on this issue first

Thanks you !! And sorry for the response delay ^^

feloy:
> > > Introduced by #715 an issue with health checks (see [#742 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/742#issuecomment-2027314597)) it should probably be better to fix it before merging this
> > 
> > 
> > OK, thanks, I'll work on this issue first
> 
> Thanks you !! And sorry for the response delay ^^

No problem, thanks for helping find the root cause of the issue. 

This PR is now rebased with #746 merged

Issue #728, opened 2024-03-28
Generated model service seems to have a wrong coding used in client code editor using curl


odockal:
I have created a service model and wanted to try the sending the request using curl, the url endpoint coding is somehow messed up. Or not yet loaded correctly. 

![ai-lab-model-service-curl-coding-broken](https://github.com/containers/podman-desktop-extension-ai-lab/assets/19164299/3425307c-ff33-4977-8076-bda5a53b1aca)

If you switch the page and go back, it gets fixed, or switching to using different language.
Another thing is that I do not have a context menu offered which would be nice to have enhancement.

PR #743, opened 2024-03-29
fix(service-details): port undefined in snippet on mount

axel7083:
### What does this PR do?

Thanks @odockal for raising the issue.

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/728

### How to test this PR?

- [x] regression test has been added

### DIFF
--- a/packages/frontend/src/pages/InferenceServerDetails.svelte
+++ b/packages/frontend/src/pages/InferenceServerDetails.svelte
@@ -67,7 +67,7 @@ const generate = async (language: string, variant: string) => {
 };
 
 $: {
-  if (snippet === undefined && service !== undefined) {
+  if (!snippet && service) {
     generate('curl', 'cURL');
   }
 }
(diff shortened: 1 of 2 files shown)


feloy (approved):

packages/frontend/src/pages/InferenceServerDetails.svelte
```diff
@@ -67,9 +66,11 @@ const generate = async (language: string, variant: string) => {
   );
 };
 
-onMount(() => {
-  generate('curl', 'cURL');
-});
+$: {
+  if (snippet === undefined && service !== undefined) {
```
benoitf, line 70:
```suggestion
  if (!snippet && service) {
```

benoitf (approved):
I think we can avoid the usage of undefined in the if

packages/frontend/src/pages/InferenceServerDetails.svelte
axel7083: Let me try it

PR #747, opened 2024-03-29
fix: display status of recipe/model

### LINKED ISSUE #574: No confirmation getting displayed to the end user when testing a new model with a recipe
When the user is trying to test a new model with a recipe there is no confirmation asked to the end user. As a result the user doesn't really know if something happened.

We should be displaying a confirmation message to the end user, so that it restarts the AI app with the newly selected model

feloy:
### What does this PR do?

Display the state of an application for a recipe AND a given model, not for the first one found for a recipe.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/9973512/03a6223b-6d4c-468d-99bd-d3b5e85bcf57

fixes #574 

### How to test this PR?

- Start a recipe with a first model
- switch to another model, the button "Start AI App" should appear
- start the recipe with the second model
- navigating to one or other model should display the information of the running app with the selected model (for example the link should point to the good instance - check the port)

### DIFF
--- a/packages/frontend/src/lib/RecipeDetails.spec.ts
+++ b/packages/frontend/src/lib/RecipeDetails.spec.ts
@@ -281,3 +281,50 @@ test('local clone and delete local clone buttons should be visible if local repo
 
   expect(mocks.requestDeleteLocalRepositoryMock).toBeCalled();
 });
+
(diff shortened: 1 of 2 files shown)


benoitf (approved):
LGTM but looks like it should have a test case

Issue #730, opened 2024-03-28
Cannot delete the model with operation not permitted


odockal:
I had a chatbot app nice and working. I have stopped the app, deleted the model service and tried to delete the model.

![deleting_model_operation_not_permitted](https://github.com/containers/podman-desktop-extension-ai-lab/assets/19164299/c66d04dc-9fb7-453a-a9e3-14caf81ceaaf)

Error:
```
Error deleting model hf.TheBloke.mistral-7b-instruct-v0.1.Q4_K_M. Error: EPERM: operation not permitted, lstat 'C:\Users\Ondrej\podman-desktop\ai-studio\models\hf.TheBloke.mistral-7b-instruct-v0.1.Q4_K_M\mistral-7b-instruct-v0.1.Q4_K_M.gguf'
```

Error in the dev console:
```
VM5:2 main ↪️ [ai-studio] Something went wrong while getting file stats (probably in use). Error: EPERM: operation not permitted, stat 'C:\Users\Ondrej\podman-desktop\ai-studio\models\hf.TheBloke.mistral-7b-instruct-v0.1.Q4_K_M\mistral-7b-instruct-v0.1.Q4_K_M.gguf'
    at Object.stat (node:internal/fs/sync:58:25)
    at statSync (node:fs:1669:17)
    at t.statSync (node:electron/js2c/node_init:2:5052)
    at ModelsManager.getLocalModelsFromDisk (C:\Users\Ondrej\.local\share\containers\podman-desktop\plugins\ghcrioprojectatomicaistudio\dist\extension.js:10275:19)
    at ModelsManager.deleteLocalModel (C:\Users\Ondrej\.local\share\containers\podman-desktop\plugins\ghcrioprojectatomicaistudio\dist\extension.js:10327:14)
```

PR #748, opened 2024-03-29
fix: model deletion

axel7083:
### What does this PR do?

When trying to delete a model we faced "EPERM: operation not permitted". When the deletion start, it was blocking itself trying to delete the directory with the model in it. I was not able to find exactly why there was this problem, but deleting first the file, then the directory solved the issue.

> I suspect some kind of weird behaviour with the filesystem watcher and the rm retry/force arguments.

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/730

### How to test this PR?

- [x] unit tests cover the chage

**Please test manually**

#### Scenario 1

- (1) download a model
- (2) delete it

#### Scenario 2

- (1) download a model
- (2) start a recipe/model service
- (3) stop the model service/recipe
- (4) delete the mode

Please update the following after testing

- [x] Windows (Axel)
- [x] Mac (Philippe)
- [ ] Linux

### DIFF
--- a/packages/backend/src/managers/modelsManager.spec.ts
+++ b/packages/backend/src/managers/modelsManager.spec.ts
@@ -400,7 +400,7 @@ describe('deleting models', () => {
     await manager.deleteModel('model-id-1');
     // check that the model's folder is removed from disk
     if (process.platform === 'win32') {
-      expect(rmSpy).toBeCalledWith('C:\\home\\user\\appstudio-dir\\model-id-1\\model-id-1-model');
+      expect(rmSpy).toBeCalledWith(expect.stringContaining('\\home\\user\\appstudio-dir\\model-id-1\\model-id-1-model'));
     } else {
       expect(rmSpy).toBeCalledWith('/home/user/appstudio-dir/model-id-1/model-id-1-model');
     }
--- a/packages/backend/src/managers/modelsManager.spec.ts
+++ b/packages/backend/src/managers/modelsManager.spec.ts
@@ -400,7 +400,9 @@ describe('deleting models', () => {
     await manager.deleteModel('model-id-1');
     // check that the model's folder is removed from disk
     if (process.platform === 'win32') {
-      expect(rmSpy).toBeCalledWith(expect.stringContaining('\\home\\user\\appstudio-dir\\model-id-1\\model-id-1-model'));
+      expect(rmSpy).toBeCalledWith(
+        expect.stringContaining('\\home\\user\\appstudio-dir\\model-id-1\\model-id-1-model'),
+      );
     } else {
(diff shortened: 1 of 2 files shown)


feloy:
Both scenarios work for me on mac

lstocchi (commented):
I just tried on win but i still see this 
image

I open AI lab, go to summarizer ,select a model which i don't have on my machine, start the app, wait it started, stop it, try to delete the model.

feloy:
> I just tried on win but i still see this image
> 
> I open AI lab, go to summarizer ,select a model which i don't have on my machine, start the app, wait it started, stop it, try to delete the model.

It seems there is an inconsistent behaviour: if I try to start a recipe with the **default** model not being on disk, I get an error saying that the model is not downloaded.

But I can do this with a non-default model, the model will be downloaded during the AI App start.

axel7083:
> I just tried on win but i still see this image
> 
> I open AI lab, go to summarizer ,select a model which i don't have on my machine, start the app, wait it started, stop it, try to delete the model.

😢 do you have other file in your folder ? tmp ? can you ls ?

lstocchi:
> @lstocchi do you have other file in your folder ? tmp ? can you ls ? 😢

I tried again after emptying the whole models folder to being sure i didn't have anything dirty. Still the same :(

axel7083:
> > @lstocchi do you have other file in your folder ? tmp ? can you ls ? 😢
> 
> I tried again after emptying the whole models folder to being sure i didn't have anything dirty. Still the same :( BTW there is only the gguf file inside the folder

Managing files on windows is a nightmare, I'm tired of it. rmdir is after rm, rm is suppose to delete the file, then rmdir delete the folder. rmdir is not suppose to be call if rm fails. rm file seems to do some kind of magic trick where it consider the file is properly deleted when it is not, and the file is locked, and only on process exit, the file is deleted not before, only on exit....

@benoitf any idea ? I'm starting to get lost with this

feloy:
I was reinstalling my windows machine today, I should be able to have a look on Tuesday. Files on Windows are a nightmare, I agree 🥵

benoitf:
I would tend to say to always use `rm` which is the newer API.

That said, on Windows, most of the issue are around file locking. So before deleting we need to be sure that no process are reading the file, etc.

Are you able to list who is owning the lock on the file ?

In the past, I know that for some projects we used like some folders with timestamp to be sure like for example to not try to write again in a folder with a previous lock.

And then, a scheduler would perform cleanup asynchronously. Files/Folders are just unwatched/removed from their availability.

so when we delete we may not really delete immediately but schedule for deletion, but for sure we remove for a resource so the file are no longer being available

axel7083:
Closing for now as I don't see easy solution

PR #751, opened 2024-03-29
refactor: merging conversation and playground objects

axel7083:
### What does this PR do?

We had two objects defined as followed

<table>
<tr>
<td> IPlayground</td> <td> IConversation </td>
</tr>
<tr>
<td> 

```ts
export interface Playground {
  id: string;
  modelId: string;
  name: string;
}
```

</td>
<td>

```ts
export interface Conversation {
  id: string;
  messages: ChatMessage[];
}
```

</td>
</tr>
</table>

Reason to merge them in a single one.

- They have **the same id**, a **playgroundId** was always matching a **conversationId**
- They force us to have two stores in the frontend
- When we were making change, we had to notify twice
- Less code to maintain 
- Easier to make changes

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/750

### How to test this PR?

- [x] existing unit tests have been updated and ensure no regression

Manually:

You can test the playground manually, no change should be visible


packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -101,19 +92,19 @@ export class PlaygroundV2Manager extends Publisher<PlaygroundV2[]> implements Di
     systemPrompt: string | undefined,
     trackingId: string,
   ): Promise<string> {
-    const id = `${this.#playgroundCounter++}`;
-
     if (!name) {
-      name = this.getFreeName();
+      name = `playground ${this.getUniqueId()}`;
```
feloy, line 96: This does not ensure unicity, as the user can set the name itself with a value that would be the same as a future unique Id.

axel7083: I restore the getFreeName method ✅

feloy (approved):

Issue #765, opened 2024-04-02
feat: ensure version of podman desktop is >=1.8.0


axel7083:
We currently have no mechanism ensuring the extension is running in version of podman desktop above 1.8.0. We are using a lot of specific apis which are only available in version 1.8.x.

We need to have a basic check ensuring the requirements are met.

deboer-tim:
Extension container files typically have `io.podman-desktop.api.version=">= 1.8.0"` (AI Lab says >=1.7.0) and package.jsons list API dependencies. Shouldn't we have one path and solve this once in Podman Desktop vs every extension being loaded and having to put in a check?

benoitf:
to me it's a nice thing (not the recommended one) as even if we add that check in Podman Desktop like 1.10 people might try the extension on previous versions of Podman Desktop that doesn't implement the check

it's like a quick win.

PR #766, opened 2024-04-02
feat: ensuring podman desktop version is >= 1.8.0

axel7083:
### What does this PR do?

This PR is adding a check for the PodmanDesktop version used. If bellow 1.8.0 we throw an error in the extension activate.

### Screenshot / video of UI

image

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/765

### How to test this PR?

**manually**

launch the extension using an old version of PodmanDesktop (E.g. 1.7.0)

### DIFF
--- a/packages/backend/src/studio.spec.ts
+++ b/packages/backend/src/studio.spec.ts
@@ -20,7 +20,7 @@
 
 import { afterEach, beforeEach, expect, test, vi } from 'vitest';
 import { Studio } from './studio';
-import { type ExtensionContext, EventEmitter } from '@podman-desktop/api';
+import { type ExtensionContext, EventEmitter, version } from '@podman-desktop/api';
 
 import * as fs from 'node:fs';
 
@@ -36,10 +36,12 @@ const mocks = vi.hoisted(() => ({
   listContainers: vi.fn(),
   getContainerConnections: vi.fn(),
   postMessage: vi.fn(),
+  logErrorMock: vi.fn(),
 }));
 
 vi.mock('@podman-desktop/api', async () => {
   return {
+    version: '1.8.0',
     fs: {
       createFileSystemWatcher: vi.fn(),
     },
@@ -60,6 +62,7 @@ vi.mock('@podman-desktop/api', async () => {
     env: {
       createTelemetryLogger: () => ({
         logUsage: vi.fn(),
+        logError: mocks.logErrorMock,
       }),
     },
     containerEngine: {
@@ -100,7 +103,8 @@ afterEach(() => {
   console.log = originalConsoleLog;
 });
 
-test('check activate ', async () => {
+test('check activate', async () => {
+  expect(version).toBe('1.8.0');
   mocks.listContainers.mockReturnValue([]);
   mocks.getContainerConnections.mockReturnValue([]);
   vi.spyOn(fs.promises, 'readFile').mockImplementation(() => {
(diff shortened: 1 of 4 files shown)


benoitf (commented):
I'm wondering if it needs to be an activation error or a warning being displayed by the extension (so the extension runs but is not functional)

packages/backend/src/studio.ts
```diff
@@ -66,6 +67,16 @@ export class Studio {
 
   public async activate(): Promise<void> {
     console.log('starting AI Lab extension');
+    this.telemetry = env.createTelemetryLogger();
+
+    // Ensure version is above 1.8.0
+    if (!version || lt(version, '1.8.0')) {
+      this.telemetry.logError('start.incompatible', {
+        version: version,
+        message: 'error activating extension on version bellow 1.8.0',
```
benoitf, line 76: there is a typo on bellow
```suggestion
        message: 'error activating extension on version below 1.8.0',
```

packages/backend/src/studio.ts
```diff
@@ -66,6 +67,16 @@ export class Studio {
...
+      });
+      throw new Error('Extension is not compatible with PodmanDesktop version bellow 1.8.');
```
benoitf, line 78:
```suggestion
      throw new Error('Extension is not compatible with Podman Desktop version below 1.8.');
```

benoitf: test would need to be fixed as well

axel7083:
> I'm wondering if it needs to be an activation error or a warning being displayed by the extension (so the extension runs but is not functional)

I have a preference for an error, as I would rather avoid putting warnings, and if( version <= 1.8.0 everywhere ) :/

packages/backend/src/studio.spec.ts
```diff
@@ -112,6 +116,19 @@ test('check activate ', async () => {
   expect(consoleLogMock).toBeCalledWith('starting AI Lab extension');
 });
 
+test('check activate incompatible', async () => {
+  (version as string) = '1.7.0';
+  await expect(async () => {
+    await studio.activate();
+  }).rejects.toThrowError('Extension is not compatible with PodmanDesktop version bellow 1.8.');
```
axel7083, line 123:
```suggestion
  }).rejects.toThrowError('Extension is not compatible with Podman Desktop version below 1.8.');
```

packages/backend/src/studio.ts
```diff
@@ -66,6 +67,16 @@ export class Studio {
 
   public async activate(): Promise<void> {
     console.log('starting AI Lab extension');
+    this.telemetry = env.createTelemetryLogger();
+
+    // Ensure version is above 1.8.0
+    if (!version || lt(version, '1.8.0')) {
```
benoitf, line 73: maybe read from package.json the value `engines.podman-desktop` instead of hardcoding it again <1.8.0 there

so at least check is consistent

axel7083: yes good idea, I updated it, and replace with `>=1.8.0` in the package json. I used then the satisfies method of the semver package.

packages/backend/src/studio.ts
```diff
@@ -66,6 +68,16 @@ export class Studio {
 
   public async activate(): Promise<void> {
     console.log('starting AI Lab extension');
+    this.telemetry = env.createTelemetryLogger();
+
+    // Ensure version is above minimum podman desktop version supported
+    if (!version || !satisfies(version, engines['podman-desktop'])) {
+      this.telemetry.logError('start.incompatible', {
+        version: version,
+        message: 'error activating extension on version below 1.8.0',
```
benoitf, line 77: you need to change the error message and replace 1.8.0 by the value of `engines['podman-desktop']`

axel7083: To avoid making the tests depend on the package.json values, which are going to change, I mocked the imports.

PR #771, opened 2024-04-02
feat: allow to import local models

### LINKED ISSUE #173: Import model

### LINKED ISSUE #814: Downloading model fails when using a user's catalog
Steps to reproduce:
- copy https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/backend/src/assets/ai.json to ~/podman-desktop/ai-lab/catalog.json
- go to a recipe page, start the ai App

A local model should be detected, or the model should be downloaded if not present locally, then the images should be built.

When a user's catalog.json file is present, the local model is not detected and is reloaded every time, and the start of the recipe fails before to build the images with an error message 'model is not available locally'

lstocchi:
### What does this PR do?

It allows to import .gguf models that have already been downloaded by the user and that are on his machine.
After imported, the user can start a service, open its folder and delete it

### Screenshot / video of UI

import_model

it resolves #173 

Fixes #814 

### How to test this PR?

1. download a gguf model which is not listed in the catalog (e.g. take the first ones from here https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/tree/main 
2. go to models, import, select the gguf file
3. once imported, try to play with it (inference server + playground)
4. then delete it (it will be removed from catalog and disk)

### DIFF
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -64,7 +64,16 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
   }
 
   private onCatalogUpdated(content: ApplicationCatalog): void {
+    // when reading the content on the catalog, the creation is just a string and we need to convert it back to a Date object
+    content.models
+      .filter(m => m.file?.creation)
+      .forEach(m => {
+        if (m.file?.creation) {
+          m.file.creation = new Date(m.file.creation);
+        }
+      });
     this.catalog = content;
+
     this.#catalogUpdateListeners.forEach(listener => listener());
     this.notify();
   }
@@ -123,6 +132,8 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
         file: {
           path: path.dirname(model.path),
           file: path.basename(model.path),
+          size: statFile.size,
+          creation: statFile.mtime,
         },
         memory: statFile.size,
       });
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -135,6 +135,7 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
           size: statFile.size,
           creation: statFile.mtime,
         },
+        memory: statFile.size,
       });
     }
 
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -23,7 +23,7 @@ import defaultCatalog from '../assets/ai.json';
 import type { Recipe } from '@shared/src/models/IRecipe';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import { Messages } from '@shared/Messages';
-import { type Disposable, type Webview } from '@podman-desktop/api';
+import { Disposable, type Webview } from '@podman-desktop/api';
 import { JsonWatcher } from '../utils/JsonWatcher';
 import { Publisher } from '../utils/Publisher';
 import type { LocalModelImportInfo } from '@shared/src/models/ILocalModelInfo';
@@ -69,8 +69,11 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
     this.notify();
   }
 
-  onCatalogUpdate(listener: catalogUpdateHandle): void {
+  onCatalogUpdate(listener: catalogUpdateHandle): Disposable {
     this.#catalogUpdateListeners.push(listener);
+    return Disposable.create(() => {
+      this.#catalogUpdateListeners.splice(this.#catalogUpdateListeners.indexOf(listener), 1);
+    });
   }
 
   dispose(): void {
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 import type { ApplicationCatalog } from '@shared/src/models/IApplicationCatalog';
+import fs from 'node:fs';
 import path from 'node:path';
 import defaultCatalog from '../assets/ai.json';
 import type { Recipe } from '@shared/src/models/IRecipe';
@@ -25,9 +26,13 @@ import { Messages } from '@shared/Messages';
 import { type Disposable, type Webview } from '@podman-desktop/api';
 import { JsonWatcher } from '../utils/JsonWatcher';
 import { Publisher } from '../utils/Publisher';
+import type { LocalModelImportInfo } from '@shared/src/models/ILocalModelInfo';
+
+export type catalogUpdateHandle = () => void;
 
 export class CatalogManager extends Publisher<ApplicationCatalog> implements Disposable {
   private catalog: ApplicationCatalog;
+  #catalogUpdateListeners: catalogUpdateHandle[];
   #disposables: Disposable[];
 
   constructor(
@@ -42,6 +47,7 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
       recipes: [],
     };
 
+    this.#catalogUpdateListeners = [];
     this.#disposables = [];
   }
 
@@ -59,7 +65,12 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
 
   private onCatalogUpdated(content: ApplicationCatalog): void {
     this.catalog = content;
+    this.#catalogUpdateListeners.forEach(listener => listener());
     this.notify();
+  }
+
+  onCatalogUpdate(listener: catalogUpdateHandle): void {
+    this.#catalogUpdateListeners.push(listener);
   }
 
   dispose(): void {
@@ -92,5 +103,39 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
       throw new Error(`No recipe found having id ${recipeId}`);
     }
     return recipe;
+  }
+
+  async addLocalModelsToCatalog(models: LocalModelImportInfo[]): Promise<void> {
+    // we copy the current catalog in another object and update it with the model
+    // then write it to the custom catalog path. If it exists it will be overwritten by default
+    const tmpCatalog: ApplicationCatalog = Object.assign({}, this.catalog);
+
+    for (const model of models) {
+      const statFile = await fs.promises.stat(model.path);
+      tmpCatalog.models.push({
+        id: model.path,
+        name: model.name,
+        description: `Model imported from ${model.path}`,
+        hw: 'CPU',
+        file: {
+          path: path.dirname(model.path),
+          file: path.basename(model.path),
+        },
+        memory: statFile.size,
+      });
+    }
+
+    const customCatalog = path.resolve(this.appUserDirectory, 'catalog.json');
+    return fs.promises.writeFile(customCatalog, JSON.stringify(tmpCatalog, undefined, 2), 'utf-8');
+  }
+
+  async removeLocalModelFromCatalog(modelId: string): Promise<void> {
+    // we copy the current catalog in another object and remove from it the model with modelId
+    // then write it to the custom catalog path.
+    const tmpCatalog: ApplicationCatalog = Object.assign({}, this.catalog);
+    tmpCatalog.models = tmpCatalog.models.filter(m => m.url !== '' && m.id !== modelId);
+
+    const customCatalog = path.resolve(this.appUserDirectory, 'catalog.json');
+    return fs.promises.writeFile(customCatalog, JSON.stringify(tmpCatalog, undefined, 2), 'utf-8');
   }
 }
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 
 import type { ApplicationCatalog } from '@shared/src/models/IApplicationCatalog';
-import fs from 'node:fs';
+import { promises } from 'node:fs';
 import path from 'node:path';
 import defaultCatalog from '../assets/ai.json';
 import type { Recipe } from '@shared/src/models/IRecipe';
@@ -111,7 +111,7 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
     const tmpCatalog: ApplicationCatalog = Object.assign({}, this.catalog);
 
     for (const model of models) {
-      const statFile = await fs.promises.stat(model.path);
+      const statFile = await promises.stat(model.path);
       tmpCatalog.models.push({
         id: model.path,
         name: model.name,
@@ -126,7 +126,7 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
     }
 
     const customCatalog = path.resolve(this.appUserDirectory, 'catalog.json');
-    return fs.promises.writeFile(customCatalog, JSON.stringify(tmpCatalog, undefined, 2), 'utf-8');
+    return promises.writeFile(customCatalog, JSON.stringify(tmpCatalog, undefined, 2), 'utf-8');
   }
 
   async removeLocalModelFromCatalog(modelId: string): Promise<void> {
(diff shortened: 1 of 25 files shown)


slemeur:
Nice!

lstocchi:
There is an issue when you delete a model having an inference server using it and you stop/start ai lab again. Fixing it

packages/shared/src/uri/Uri.ts
```diff
@@ -0,0 +1,78 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import type { Uri as APIUri } from '@podman-desktop/api';
+
+export class Uri {
```
axel7083, line 20: Could you explain why we need to have our own implementation of the Uri class ? I am not sure to get it

packages/backend/src/managers/inference/inferenceManager.ts
```diff
@@ -383,11 +392,12 @@ export class InferenceManager extends Publisher<InferenceServer[]> implements Di
               port: !!containerInfo.Ports && containerInfo.Ports.length > 0 ? containerInfo.Ports[0].PublicPort : -1,
             },
             status: containerInfo.Status === 'running' ? 'running' : 'stopped',
-            models: modelsId.map(id => this.modelsManager.getModelInfo(id)),
+            models: models,
```
axel7083, line 395: I think we can avoid having all this try catch with the `getModelInfo` potentially throwing error by filtering the models Id using the `isModelOnDisk` from the `ModelsManager`

```suggestion
            models: modelsId.filter(id => this.modelsManager.isModelOnDisk(id)).map(id => this.modelsManager.getModelInfo(id)),,
```

packages/backend/src/managers/catalogManager.ts
```diff
@@ -93,4 +104,38 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
...
+        hw: 'CPU',
+        file: {
+          path: path.dirname(model.path),
+          file: path.basename(model.path),
+        },
+        memory: statFile.size,
+      });
+    }
+
+    const customCatalog = path.resolve(this.appUserDirectory, 'catalog.json');
```
axel7083, line 128: I think we should have two catalog here. I am not in favor of overwriting the _native_ catalog, I feel it would be more appropriate to have some kind of `user-models.json` in the `appUserDirectory` what do you think ?

axel7083 (commented):
Nice feature ! I added some comment

packages/shared/src/uri/Uri.ts
lstocchi: The Uri object which is returned when calling the api is serialized using private properties bc the Uri constructor uses `private _scheme ..., private _path ...`. So when we receive the resulting object in ai lab we have 
```
_path: '...',
_scheme: '...',
```
In this way it is not kinda usable so we need to "revive" it so that its fields are exposed through methods like the original Uri.
The same thing is done on Desktop.

axel7083: What are the properties that we need ?

packages/backend/src/managers/catalogManager.ts
lstocchi: From my point of view the operation that the user is doing here it is just updating the custom catalog.json file that we already support. It is just that we guide them. But if someone wants to add a new model manually we are already saying that they need to create a `catalog.json` file (copy/paste the `assets/ai.json`) inside the ai-lab folder and update it. 

So i wouldn't create a new file as it may overcomplicate stuff, but let's see what others think

axel7083: How do you handle the case where we update on our side the catalog ? Are we losing the user imported models ?

lstocchi: When importing, if the `catalog.json` file already exists, it just adds the new models in there and update the file on disk. If it does not exists it creates it on disk by copying the content from the native `ai.json` and adding the imported models. 

So if the user wants to update the catalog after having imported models he would work as before. Open the `catalog.json` file which would also contain the imported models and edit it.

packages/backend/src/managers/inference/inferenceManager.ts
lstocchi: Your suggestion does a different thing but i don't know which should be the correct behavior tbh.

In the current code i skip the containers that are associated to a model which is not on the catalog/disk anymore.
In your case you're still detecting the inference server but its models list can be empty.

So not sure how it should work. Should we not show the inference servers that have been created with a no-more-existing model or we should show them and associate an empty models list? Not sure if the latter implies some misbehavior though.

packages/backend/src/managers/catalogManager.ts
```diff
@@ -59,9 +65,14 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
 
   private onCatalogUpdated(content: ApplicationCatalog): void {
     this.catalog = content;
+    this.#catalogUpdateListeners.forEach(listener => listener());
     this.notify();
   }
 
+  onCatalogUpdate(listener: catalogUpdateHandle): void {
+    this.#catalogUpdateListeners.push(listener);
+  }
```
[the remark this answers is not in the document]
lstocchi: Done, Thanks!!

packages/backend/src/managers/catalogManager.ts
axel7083: What if some day we want to remove a model ? like we did with llama2 ? the mechanism you describe would not be suitable ?

packages/backend/src/managers/inference/inferenceManager.ts
lstocchi: Agree, opened https://github.com/containers/podman-desktop-extension-ai-lab/issues/791

packages/backend/src/managers/catalogManager.ts
lstocchi: This is the same thing that may happen today. If i create my custom catalog.json and we update ai.json, ai-lab will still point to the custom catalog. We have to refactor how the custom catalog + ai.json work or inform the user any update. Opening an issue

lstocchi: -> https://github.com/containers/podman-desktop-extension-ai-lab/issues/792

packages/backend/src/managers/inference/inferenceManager.ts
axel7083: SO for now I would rather have the servers showed even tho we don't have the model, so the user can do cleanup from ai studio

axel7083 (commented):
Almost looking good just some stuff I find out while testing it

## Breadcrumb

image

Could we have the breadcrumb allowing to go back to models page ?

## Import timeout

Taking very long time to select a file in explorer gave a timeout

image

We probably need to find out some kind of way to get back this value and update the form. (This could be skipped, but would need to be address at some point)

## Selected file

Maybe I should not be able to select files inside the models folder ?

image

If I import my own file, it gets weird

image

packages/backend/src/managers/inference/inferenceManager.ts
lstocchi: Done

lstocchi:
Nice catches!!

> Almost looking good just some stuff I find out while testing it
> 
> ## Breadcrumb
 
Done

> ## Import timeout
> Taking very long time to select a file in explorer gave a timeout

I'll give a look on desktop side as the timeout is really tight. For the moment i added an error message if something unexpected happen so the user is informed.

> ## Selected file
> Maybe I should not be able to select files inside the models folder ?

Added a check to verify that the selected model is not already in the catalog.

jeffmaury:
The size information is missing and I'm wondering how the RAM usage is computed

packages/backend/src/managers/catalogManager.ts
jeffmaury: What about having the 2 files but merging them instead of choosing one over the other

jeffmaury (commented):
The timeout issue is to be fixed because it breaks the user experience

lstocchi:
> The size information is missing and I'm wondering how the RAM usage is computed

The RAM usage is = size of the file as it seems to me that that is the value we use in the catalog, no?

lstocchi:
@jeffmaury timeout removed and size + creation date added

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -124,6 +124,7 @@ export class RpcBrowser {
   counter: number = 0;
   promises: Map<number, { resolve: (value: unknown) => unknown; reject: (value: unknown) => void }> = new Map();
   subscribers: Map<string, (msg: any) => void> = new Map();
+  noTimeoutChannels: string[] = ['openDialog'];
```
axel7083, line 127: please make it available somewhere like the Messages ids in consts file or something

axel7083 (changes requested):

feloy:
> > The size information is missing and I'm wondering how the RAM usage is computed
> 
> The RAM usage is = size of the file as it seems to me that that is the value we use in the catalog, no?

No, the ram usage does not come from the file size. For the models in the catalog, it is computed with the script in `.github/workflows/compute-model-sizes.yaml` (see #674)

packages/shared/src/messages/MessageProxy.ts
lstocchi: Done

lstocchi:
> > > The size information is missing and I'm wondering how the RAM usage is computed
> > 
> > 
> > The RAM usage is = size of the file as it seems to me that that is the value we use in the catalog, no?
> 
> No, the ram usage does not come from the file size. For the models in the catalog, it is computed with the script in `.github/workflows/compute-model-sizes.yaml` (see #674)

Ah I completely missed that part. Thanks for telling me. Removed the memory property for the moment then. 👍

packages/shared/src/models/NoTimeoutChannels.ts
```diff
@@ -0,0 +1,19 @@
+/**********************************************************************
```
axel7083, line 1: Why is it in the `src/models` folder ? could we move it to the messages folder ?

packages/frontend/src/pages/ImportModels.svelte
```diff
@@ -0,0 +1,147 @@
...
+  if (importError === '') {
+    router.goto(`/models`);
+  }
+}
+</script>
+
+<NavPage lastPage="{{ name: 'Models', path: '/models' }}" title="Import Models" searchEnabled="{false}">
+  <svelte:fragment slot="content">
+    <div class="p-5 min-w-full h-fit">
+      <div class="bg-charcoal-600 px-6 py-4 space-y-2 lg:px-8 sm:pb-6 xl:pb-8">
```
axel7083, line 94:
```suggestion
      <div class="py-4 space-y-2">
```

**before**
image

**after**
image

axel7083, line 94:
```suggestion
      <div class="bg-charcoal-800 rounded-lg px-6 py-4 space-y-2">
```

**before**

image

**after**
image

packages/frontend/src/pages/ImportModels.svelte
```diff
@@ -0,0 +1,147 @@
...
+            <Button type="link" on:click="{() => deleteModelToImport(index)}" icon="{faMinusCircle}" />
+          </div>
+        {/each}
+
+        <div aria-label="importError">
+          {#if importError !== ''}
+            <ErrorMessage class="py-2 text-sm" error="{importError}" />
+          {/if}
+        </div>
+        <Button
```
axel7083, line 135:
```suggestion
        <div class="pt-5 border-zinc-600 border-t-2"/>
        <Button
```

Adding a divider to match the design container import

packages/frontend/src/pages/ImportModels.svelte
```diff
@@ -0,0 +1,147 @@
...
+  if (importError === '') {
+    router.goto(`/models`);
+  }
+}
+</script>
+
+<NavPage lastPage="{{ name: 'Models', path: '/models' }}" title="Import Models" searchEnabled="{false}">
+  <svelte:fragment slot="content">
+    <div class="p-5 min-w-full h-fit">
+      <div class="bg-charcoal-800 rounded-lg px-6 py-4 space-y-2">
```
axel7083, line 94:
```suggestion
      <div class="bg-charcoal-800 rounded-lg px-6 py-6 space-y-2">
              <span class="block mb-2 text-sm font-medium text-gray-400">Models to import:</span>
```

To match the design of the import containers

Here is the example
image

with the changes you get
image

packages/frontend/src/lib/table/model/ModelColumnName.svelte
```diff
@@ -8,9 +8,18 @@ function openDetails() {
...
 </script>
 
-<button class="flex flex-col w-full" title="{object.name}" on:click="{() => openDetails()}">
-  <div class="text-sm text-gray-300 overflow-hidden text-ellipsis w-full">
+<button
+  class="flex flex-col w-full"
+  title="{object.name}"
+  on:click="{() => openDetails()}"
+  aria-label="Open Model Details">
+  <div class="text-sm text-gray-300 overflow-hidden text-ellipsis w-full" aria-label="Model Name">
```
axel7083, line 16:
```suggestion
  <div class="text-sm text-gray-300 overflow-hidden text-ellipsis w-full text-left" aria-label="Model Name">
```

When the name is too short we have some weird alignements

image

jeffmaury:
> > > > The size information is missing and I'm wondering how the RAM usage is computed
> > > 
> > > 
> > > The RAM usage is = size of the file as it seems to me that that is the value we use in the catalog, no?
> > 
> > 
> > No, the ram usage does not come from the file size. For the models in the catalog, it is computed with the script in `.github/workflows/compute-model-sizes.yaml` (see #674)
> 
> Ah I completely missed that part. Thanks for telling me. Removed the memory property for the moment then. 👍

I would rather default the memory field to the file size

axel7083:
> > > > > The size information is missing and I'm wondering how the RAM usage is computed
> > > > 
> > > > 
> > > > The RAM usage is = size of the file as it seems to me that that is the value we use in the catalog, no?
> > > 
> > > 
> > > No, the ram usage does not come from the file size. For the models in the catalog, it is computed with the script in `.github/workflows/compute-model-sizes.yaml` (see #674)
> > 
> > 
> > Ah I completely missed that part. Thanks for telling me. Removed the memory property for the moment then. 👍
> 
> I would rather default the memory field to the file size

I think this is misleading

vrothberg:
If we don't have the information about the resource constraints of a model, I wouldn't guess it.  Maybe we can display `NA` (not available) for imported models?

packages/frontend/src/pages/ImportModels.svelte
lstocchi: So the container import page has to be updated. I forgot and i'll create a PR for this on desktop. When working on the load images page i was copying the container import template and Tim commented to remove the divider as it is not used anywhere else https://github.com/containers/podman-desktop/pull/6540#discussion_r1539272375

axel7083: Okey np

lstocchi: Thanks for making me remind about it 😆 -> https://github.com/containers/podman-desktop/pull/6655

jeffmaury:
> > > > > > The size information is missing and I'm wondering how the RAM usage is computed
> > > > > 
> > > > > 
> > > > > The RAM usage is = size of the file as it seems to me that that is the value we use in the catalog, no?
> > > > 
> > > > 
> > > > No, the ram usage does not come from the file size. For the models in the catalog, it is computed with the script in `.github/workflows/compute-model-sizes.yaml` (see #674)
> > > 
> > > 
> > > Ah I completely missed that part. Thanks for telling me. Removed the memory property for the moment then. 👍
> > 
> > 
> > I would rather default the memory field to the file size
> 
> I think this is misleading and does not reflect reality

Both are in the same order of magnitude (file size is a little bit bigger) and as we will check the machine size based on this data, I would rather ask a little bit of additional memory than risking failures

axel7083 (approved):
LGTM ! 🚀 thanks you for your patience and the fixes, great job!

lstocchi:
@jeffmaury set the memory value equals to the file size as you said. Let me know if there is something else to change.
Also rebased.

jeffmaury (approved):
LGTM

Issue #787, opened 2024-04-03
AI Lab does not work with fresh builds of Podman Desktop


slemeur:
Tried to install this: https://github.com/containers/podman-desktop/releases/tag/untagged-bac9090d85d83f7c6b02

And use this tag: ghcr.io/containers/podman-desktop-extension-ai-lab:5a705f2f01109fcf7c65992aaf05153fa3c714a9

<img width="1162" alt="Screenshot 2024-04-03 at 12 09 36" src="https://github.com/containers/podman-desktop-extension-ai-lab/assets/1636769/c06adcce-ac0c-481f-a75f-1e5faf79fd9a">

PR #789, opened 2024-04-03
fix: nighties version are not subject to version verification

axel7083:
### What does this PR do?

The special case of nighties are handled with an ignore warning. We are checking if `major` and `minor` are zeroes, if it does, we ignore the version check.

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/787

### How to test this PR?

- [x] unit tests ensure no regression

### DIFF
--- a/packages/backend/src/studio.spec.ts
+++ b/packages/backend/src/studio.spec.ts
@@ -18,7 +18,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+import { afterEach, beforeEach, expect, test, vi, describe } from 'vitest';
 import { Studio } from './studio';
 import { type ExtensionContext, EventEmitter, version } from '@podman-desktop/api';
 
@@ -37,6 +37,8 @@ const mocks = vi.hoisted(() => ({
   getContainerConnections: vi.fn(),
   postMessage: vi.fn(),
   logErrorMock: vi.fn(),
+  consoleWarnMock: vi.fn(),
+  consoleLogMock: vi.fn(),
 }));
 
 vi.mock('../package.json', () => ({
@@ -95,11 +97,11 @@ vi.mock('@podman-desktop/api', async () => {
 
 /// mock console.log
 const originalConsoleLog = console.log;
-const consoleLogMock = vi.fn();
 
 beforeEach(() => {
   vi.clearAllMocks();
-  console.log = consoleLogMock;
+  console.log = mocks.consoleLogMock;
+  console.warn = mocks.consoleWarnMock;
 
   vi.mocked(EventEmitter).mockReturnValue({
     event: vi.fn(),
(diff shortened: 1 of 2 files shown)


jeffmaury (changes requested):
Lint errors otherwise LGTM

axel7083:
> Lint errors otherwise LGTM

@jeffmaury fixed

packages/backend/src/studio.ts
```diff
@@ -66,18 +66,34 @@ export class Studio {
...
   }
 
+  private checkVersion(): boolean {
+    if (!version) return false;
+
+    const current = coerce(version);
+    if (!current) return false;
+
+    if (current.major === 0 && current.minor === 0) {
+      console.warn('nighties version are not subject to version verification.');
```
deboer-tim, line 76: Nighties are something women wear to bed, please fix typo to nightlies throughout. ;-)

jeffmaury (approved):
LGTM

packages/backend/src/studio.ts
```diff
@@ -66,18 +66,34 @@ export class Studio {
...
+
   public async activate(): Promise<void> {
     console.log('starting AI Lab extension');
     this.telemetry = env.createTelemetryLogger();
 
-    // Ensure version is above minimum podman desktop version supported
-    if (!version || !satisfies(coerce(version), engines['podman-desktop'])) {
+    if (!this.checkVersion()) {
       const min = minVersion(engines['podman-desktop']);
+      const current = version ?? 'unknown';
```
deboer-tim, line 89: I'm confused why the `??` is necessary since the type of version is `string`, not `string | undefined`?

deboer-tim (commented):
Code LGTM and same is working with bootc, just one typo and a question.

packages/backend/src/studio.ts
benoitf: old versions of Podman Desktops we're not providing the version field

deboer-tim: Ah, that makes sense then. But what about `(version as string) = undefined;` in test? It gives me an error locally since version cannot be undefined in current API.

packages/backend/src/studio.ts
axel7083, line 76:
```suggestion
      console.warn('nightlies version are not subject to version verification.');
```

packages/backend/src/studio.spec.ts
```diff
@@ -123,32 +125,55 @@ test('check activate', async () => {
...
-test('check activate next value', async () => {
-  (version as string) = '1.0.1-next';
-  await studio.activate();
+  test('check activate nighties value', async () => {
+    (version as string) = 'v0.0.202404030805-3cb4544';
+    await studio.activate();
 
-  expect(mocks.logErrorMock).not.toHaveBeenCalled();
+    expect(mocks.logErrorMock).not.toHaveBeenCalled();
+    expect(mocks.consoleWarnMock).toHaveBeenCalledWith('nighties version are not subject to version verification.');
```
axel7083, line 170:
```suggestion
    expect(mocks.consoleWarnMock).toHaveBeenCalledWith('nightlies version are not subject to version verification.');
```

packages/backend/src/studio.ts
axel7083: Thanks for raising this !!!

PR #796, opened 2024-04-03
feat: replace simple-git with isomorphic-git

### LINKED ISSUE #390: SimpleGit calling cannot find the git command when Podman Desktop executed from flatpak

feloy:
### What does this PR do?

Replace `simple-git` library (using the git binary) with `isomorphic-git` which is a pure JS implementation of git. 

The PR also changes the `ref` for the recipes, to use the full commit id instead of the short commit id (because isomorphic-git does not support natively doing a checkout when passed a short commit id as ref). This change can be reverted if we want, to keep the original short id, and the expansion of the id can be done every time (by checking it is not a branch/tag name first).

Fixes #390

### How to test this PR?

- with the recipes of the catalog (all are using a commit id as ref)
  - start a recipe, then stop it
  - modify / delete / add files to the repository, 
  - or checkout another commit, branch, etc
  - restart the recipe and check that an alert popup is raised
  - test different buttons of the popup (cancel, continue, reset)
- add new recipes to the catalog, with no ref (using the default branch), or with a ref pointing to a branch or tag and test the same scenarios

### DIFF
--- a/packages/backend/src/managers/gitManager.ts
+++ b/packages/backend/src/managers/gitManager.ts
@@ -204,7 +204,7 @@ export class GitManager {
 
     if (branch) {
       const tracking = await this.getTrackingBranch(directory, branch);
-      if (ref !== undefined && tracking !== `origin/${ref}`) {
+      if (ref && tracking !== `origin/${ref}`) {
         return {
           error: `The local repository is not tracking the right branch. (tracking ${tracking} when expected ${ref})`,
         };


PACKAGING-GUIDE.md
```diff
@@ -49,6 +49,7 @@ A recipe has the following attributes:
 - ```name```: the recipe name
 - ```description```: a detailed description about the recipe
 - ```repository```: the URL where the recipe code can be retrieved
+- ```ref```: an optional ref in the repository to checkout (a branch name, tag name, or commit full id - short commit id won't be recognized). If not defined, tha default branch will be used
```
jeffmaury, line 52:
```suggestion
- ```ref```: an optional ref in the repository to checkout (a branch name, tag name, or commit full id - short commit id won't be recognized). If not defined, the default branch will be used
```

lstocchi (approved):
Tested on linux flatpack and windows. Tried to run recipe having an old/correct repo and it worked perfectly. Nice job!! LGTM 🚀

benoitf (approved):
LGTM

PR #799, opened 2024-04-03
feat: add a Feedback section in the Dashboard

### LINKED ISSUE #62: Feedback mechanism for the extension

jeffmaury:
Fixes #62

### What does this PR do?

Add a new Feedback section in the dashboard page

### Screenshot / video of UI

image

#62 

### How to test this PR?

Go to the dashboard page

### DIFF
--- a/packages/frontend/src/pages/Dashboard.svelte
+++ b/packages/frontend/src/pages/Dashboard.svelte
@@ -1,6 +1,9 @@
 <script lang="ts">
 import NavPage from '/@/lib/NavPage.svelte';
 import { router } from 'tinro';
+import Button from '/@/lib/button/Button.svelte';
+import { faGithub } from '@fortawesome/free-brands-svg-icons';
+import { studioClient } from '/@/utils/client';
 
 const openRecipesCatalog = () => {
   router.goto('/recipes');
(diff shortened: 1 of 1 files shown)


packages/frontend/src/pages/Dashboard.svelte
```diff
@@ -13,9 +16,24 @@ const openModelsPage = () => {
 const openPlaygroundsPage = () => {
   router.goto('/playgrounds');
 };
+
+const openGithub = () => {
+  -studioClient.openURL('https://github.com/containers/podman-desktop-extension-ai-lab');
```
benoitf, line 21: why do we have the `-` ?

benoitf (commented):
it does the job, but  maybe it should be more visible if we want to get feedback

either a different background or separate from the flow or put it on top

packages/frontend/src/pages/Dashboard.svelte
```diff
@@ -71,6 +89,26 @@ const openPlaygroundsPage = () => {
...
+          <a
+            href="{'javascript:void(0);'}"
+            class="underline"
+            role="button"
+            title="Open the issues page"
+            on:click="{openIssuesPage}">here</a>
+        </p>
+        <p>
+          If you want to get more information about this project or exchange with the team, you can start a discussion
+          <a
```
benoitf, line 105: I'm also wondering why you're not using `<Button>` component but then a link with href with JavaScript void

AFAIK we have Link type button

benoitf (approved):

benoitf (commented):
I was thinking 'our Button' not 'button'

jeffmaury:
> I was thinking 'our Button' not 'button'

Trying but got 

image

benoitf:
ok 🤦

benoitf (approved):

axel7083:
I am okey with this PR, but I am wondering if we should have some kind of more direct, flashy way to ask for this. The dashboard being a lot of text, I would not be surprised that people never read and reach the paragraph added. 

In a follow up PR we could create a banner, or something placed at the top with clear and direct access for giving feedback.

Here is a quick mockup 

image

cc @slemeur

PR #812, opened 2024-04-04
chore(tsconfig): enabling strictNullChecks

axel7083:
### What does this PR do?

Enabling the `strictNullChecks` and fixing the places where it raised an error

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/811

### How to test this PR?

- [x] pipeline should be ✅ 

⚠️ Please do try some basic scenarios, this should not change much, but we are never too much careful

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -342,7 +342,6 @@ describe('pullApplication', () => {
       description: '',
       readme: '',
       repository: 'repo',
-      basedir: 'ai-lab.yaml',
     };
     const model: ModelInfo = {
       id: 'model1',
@@ -371,7 +370,6 @@ describe('pullApplication', () => {
       ref: '000000',
       readme: '',
       repository: 'repo',
-      basedir: 'ai-lab.yaml',
     };
     const model: ModelInfo = {
       id: 'model1',
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -278,7 +278,6 @@ describe('pullApplication', () => {
       ref: '000000',
       readme: '',
       repository: 'repo',
-      config: 'ai-lab.yaml',
     };
     const model: ModelInfo = {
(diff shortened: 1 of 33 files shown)


benoitf:
hello, could we postpone these chore post 1.0 ?

axel7083:
> hello, could we postpone these chore post 1.0 ?

I guess we can, but I still think that not having this lead to uncatch errors in our code

packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -342,6 +342,7 @@ describe('pullApplication', () => {
       description: '',
       readme: '',
       repository: 'repo',
+      basedir: 'ai-lab.yaml',
```
axel7083, line 345:
```suggestion
```

packages/backend/src/managers/applicationManager.spec.ts
```diff
@@ -370,6 +371,7 @@ describe('pullApplication', () => {
       ref: '000000',
       readme: '',
       repository: 'repo',
+      basedir: 'ai-lab.yaml',
```
axel7083, line 374:
```suggestion
```

axel7083:
🥳 1.0 has been released, re-opening !

packages/backend/src/studio.ts
```diff
@@ -126,6 +127,7 @@ export class Studio {
         const href = link.match(/href="(.*?)"/);
         if (href) {
           const webviewHref = this.#panel?.webview.asWebviewUri(Uri.joinPath(extensionUri, 'media', href[1]));
+          if (!webviewHref) throw new Error('undefined webviewHref');
```
benoitf, line 130: I think you might need to add more context in the error message for this one and the one line 117 because if I see the error, I don't know for which link for example we tried to find the webviewHref, etc

axel7083, line 130:
```suggestion
          if (!webviewHref) throw new Error('Something went wrong while replacing links with webView Uri links: undefined webviewHref');
```

benoitf (approved):

PR #824, opened 2024-04-05
fix: wrong alignment in playground table header

### LINKED ISSUE #786: Wrong alignement in the header for list of Playground
Screenshot 2024-04-03 at 12 05 04

lstocchi:
### What does this PR do?

fix alignment.
instead of keeping the left alignment and move the action icon a bit more away from the right-end. I think it's nicer to have it centered. Maybe when having more icon we can think about switching back to left

### Screenshot / video of UI

image

it resolves #786 

### How to test this PR?

1. go to the playgrounds page


axel7083 (approved):
LGTM! 🚀

benoitf (approved):

PR #833, opened 2024-04-08
fix: open in VScode URL format is broken on Windows

### LINKED ISSUE #832: Open in VSCode broken on Windows

jeffmaury:
Fixes #832

### What does this PR do?

Fix the VSCode URL which is not valid for Windows

#832

### How to test this PR?

1. Pull a recipe
2. Open it in VSCode

### DIFF
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -367,7 +367,7 @@ export class StudioApiImpl implements StudioAPI {
     const unixPath: string = path.normalize(directory).replace(/[\\/]+/g, '/');
 
     podmanDesktopApi.env
-      .openExternal(podmanDesktopApi.Uri.file(unixPath).with({ scheme: 'vscode', authority: 'file' }))
+      .openExternal(podmanDesktopApi.Uri.file(unixPath).with({ scheme: 'vscode', authority: 'file/' }))
       .catch((err: unknown) => {
         console.error('Something went wrong while trying to open VSCode', err);
       });


packages/backend/src/studio-api-impl.ts
```diff
@@ -367,7 +367,7 @@ export class StudioApiImpl implements StudioAPI {
     const unixPath: string = path.normalize(directory).replace(/[\\/]+/g, '/');
 
     podmanDesktopApi.env
-      .openExternal(podmanDesktopApi.Uri.file(unixPath).with({ scheme: 'vscode', authority: 'file' }))
+      .openExternal(podmanDesktopApi.Uri.file(unixPath).with({ scheme: 'vscode', authority: 'file/' }))
```
axel7083, line 370: @benoitf is this the expected behaviour ? I feel like this is an issue on the podman side ?

I am wondering, if it would be better to have instead something like above to match how URI are working

```
if(!unixPath.startsWith('/')) {
     unixPath = `/${unixPath}`;
}
```

image

benoitf: yes it feels invalid to have a / in the authority

first slash should be part of the unixPath if it requires one

also please add tests for openVSCode method for Windows and macOS

axel7083 (approved):
Thanks for adding the tests! LGTM :rocket:

packages/backend/src/studio-api-impl.spec.ts
```diff
@@ -176,3 +184,25 @@ test('if requestDeleteLocalRepository fails an errorMessage should show up', asy
...
+      },
+    };
+  });
+
+  mocks.openExternalMock.mockResolvedValue(true);
+  await studioApiImpl.openVSCode(folder);
+  expect(mocks.openExternalMock).toHaveBeenCalledWith(
+    expect.objectContaining({ path: expect.stringMatching(/^\//), authority: 'file', scheme: 'vscode' }),
+  );
+});
```
benoitf, line 208: for CI there is no pb as it's running both on Windows and macOS (so path.resolve will work)

but usually unit tests cover both Windows and macOS/Linux cases by testing the two ways

so you can see problems before CI running the test locally either on Windows or on macOS/Linux

like using a test with multiple argument('one with a path obtained from windows) and one from macOS/Linux)
`test.each([...`

benoitf (approved):

Issue #836, opened 2024-04-08
Implement playground telemetry


jeffmaury:
Playground

- Define a playground with a system prompt
- Modified the system prompt
- Change in the parameters
- Send prompt

PR #837, opened 2024-04-08
chore: adding telemetry for playground manager

axel7083:
### What does this PR do?

Adding the following telemetry events

- `playground.delete` 
- `playground.create`
- `playground.system-prompt.create`
- `playground.system-prompt.update`
- `playground.system-prompt.delete`
- `playground.submit`
- `playground.message.complete` (provide a duration data, corresponding to the time between the first and last token provided by the models.)

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/836

### How to test this PR?

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.ts
@@ -43,13 +43,11 @@ export class PlaygroundV2Manager implements Disposable {
 
   deleteConversation(conversationId: string): void {
     const conversation = this.#conversationRegistry.get(conversationId);
-    if (conversation !== undefined) {
-      this.telemetry.logUsage('playground.delete', {
-        totalMessages: conversation.messages.length,
-        modelId: conversation.modelId,
-      });
-      this.#conversationRegistry.deleteConversation(conversationId);
-    }
+    this.telemetry.logUsage('playground.delete', {
+      totalMessages: conversation.messages.length,
+      modelId: conversation.modelId,
+    });
+    this.#conversationRegistry.deleteConversation(conversationId);
   }
 
   async requestCreatePlayground(name: string, model: ModelInfo, systemPrompt?: string): Promise<string> {
@@ -149,7 +147,9 @@ export class PlaygroundV2Manager implements Disposable {
       id: this.#conversationRegistry.getUniqueId(),
       timestamp: Date.now(),
     } as SystemPrompt);
-    this.telemetry.logUsage('playground.system-prompt.create');
+    this.telemetry.logUsage('playground.system-prompt.create', {
+      modelId: this.#conversationRegistry.get(conversationId).modelId,
+    });
   }
 
   /**
@@ -160,11 +160,12 @@ export class PlaygroundV2Manager implements Disposable {
    */
   setSystemPrompt(conversationId: string, content: string | undefined): void {
     const conversation = this.#conversationRegistry.get(conversationId);
-    if (conversation === undefined) throw new Error(`Conversation with id ${conversationId} does not exists.`);
 
     if (content === undefined || content.length === 0) {
       this.#conversationRegistry.removeMessage(conversationId, conversation.messages[0].id);
-      this.telemetry.logUsage('playground.system-prompt.delete');
+      this.telemetry.logUsage('playground.system-prompt.delete', {
+        modelId: conversation.modelId,
+      });
       return;
     }
 
@@ -174,7 +175,9 @@ export class PlaygroundV2Manager implements Disposable {
       this.#conversationRegistry.update(conversationId, conversation.messages[0].id, {
         content,
       });
-      this.telemetry.logUsage('playground.system-prompt.update');
+      this.telemetry.logUsage('playground.system-prompt.update', {
+        modelId: conversation.modelId,
+      });
     } else {
       throw new Error('Cannot change system prompt on started conversation.');
     }
@@ -187,7 +190,6 @@ export class PlaygroundV2Manager implements Disposable {
    */
   async submit(conversationId: string, userInput: string, options?: ModelOptions): Promise<void> {
     const conversation = this.#conversationRegistry.get(conversationId);
-    if (conversation === undefined) throw new Error(`conversation with id ${conversationId} does not exist.`);
 
     const servers = this.inferenceManager.getServers();
     const server = servers.find(s => s.models.map(mi => mi.id).includes(conversation.modelId));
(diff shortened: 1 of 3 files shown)


packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -206,7 +231,11 @@ export class PlaygroundV2Manager implements Disposable {
         });
       })
       .catch((err: unknown) => {
-        console.error('Something went wrong while creating model reponse', err);
+        console.error('Something went wrong while creating model response', err);
+        this.telemetry.logError('playground.error', {
```
benoitf, line 235: Thanks,

FYI could you avoid the usage of telemetry.logError (we've stopped to use that pattern)


prefer

```ts
telemetryData = {};
try {
  const foo = await foo();
  telemetryData.foo =  foo
} catch (error) {
  telemetryData.error = error
}

telemetryLogger.logUsage('myAction', telemetryData);
```

it's because it's easier to map errors with a single event rather than having to match stuff from 2 different events

so if you can send the error as part of the 'playground.submit' event (if any)

axel7083: Oh thanks for the information, I will fix it, and be careful !

axel7083:
> General statement: I feel like the call to logUsage should be done once we've done the call and not as first thing of the method (last call vs first method call)
> 
> Probably need to move for example
> 
> ```
>  this.telemetry.logUsage('playground.create', {
>       hasName: !!name,
>       hasSystemPrompt: !!systemPrompt,
>       modelId: model.id,
>     });
> ```
> 
> at the bottom

Okey for this

> I see two places with `this.telemetry.logUsage('playground.system-prompt.create'`
> 
> I think that we may have later issues, we should have a single entry to createSystemPrompt (and then log only once telemetry from there) so if two methods need to create a system prompt, they call the same method.

We have two way to create a system prompt, from the PlaygroundCreateForm, and manually when the playground is already created, I added a differentiator `onCreation: boolean` to differentiate those two. As we probably want to track if we are adding the system prompt from the form or later.

benoitf:
> We have two way to create a system prompt, from the PlaygroundCreateForm, and manually when the playground is already created, I added a differentiator onCreation: boolean to differentiate those two. 

yes but shouldn't telemetry event shouldn't be sent from the method handling the onCreation parameter ?

I mean rather than calling
```ts
  this.#conversationRegistry.submit(conversationId, {
        role: 'system',
        content,
        timestamp: Date.now(),
      } as SystemPrompt);

 telemetry.logUsage(onCreation, false)
```

and 

```ts
this.#conversationRegistry.submit(conversationId, {
        content: systemPrompt,
        role: 'system',
        id: this.#conversationRegistry.getUniqueId(),
        timestamp: Date.now(),
      } as SystemPrompt);
 this.telemetry.logUsage('playground.system-prompt.create', {
        onCreation: true,
      });
```

have a method

```ts
createSystemPrompt(id, {param}, creation) {
this.#conversationRegistry.submit(conversationId,param);
 this.telemetry.logUsage('playground.system-prompt.create', {
        onCreation: creation,
      });
}

and you call it from your 2 locations instead of duplicating the calls to conversationRegistry.submit()

```

let say someone want to add extra fields on logUsage, we would need to add that at two places while here we have a single place

axel7083:
> have a method createSystemPrompt

I applied the suggestion

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -232,6 +264,9 @@ export class PlaygroundV2Manager implements Disposable {
     }
 
     this.#conversationRegistry.completeMessage(conversationId, messageId);
+    this.telemetry.logUsage('playground.message.complete', {
+      duration: Date.now() - start,
+    });
```
benoitf, line 269: probably a later improvement

but maybe we should log some contextual information on each event.

For example what is the anonymized version of the model name (like its sha), some other useful parameters 
like the number of cpu/memory of the machine, etc

benoitf (approved):

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -64,6 +71,8 @@ export class PlaygroundV2Manager implements Disposable {
         });
       })
       .catch((err: unknown) => {
+        telemetry['error'] = `${String(err)}`;
```
axel7083, line 74:
```suggestion
        telemetry['errorMessage'] = `${String(err)}`;
```

jeffmaury (commented):
I think from a correlation point of view, it would be good to have modelId being part of each event

axel7083:
> I think from a correlation point of view, it would be good to have modelId being part of each event

I added the modelId for the following events

- `playground.delete`
- `playground.message.complete`

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -126,6 +137,21 @@ export class PlaygroundV2Manager implements Disposable {
...
+   * @param content the content of the system prompt
+   */
+  private submitSystemPrompt(conversationId: string, content: string): void {
+    this.#conversationRegistry.submit(conversationId, {
+      content: content,
+      role: 'system',
+      id: this.#conversationRegistry.getUniqueId(),
+      timestamp: Date.now(),
+    } as SystemPrompt);
+    this.telemetry.logUsage('playground.system-prompt.create');
```
jeffmaury, line 152: Why modelId is not being given ?

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -136,20 +162,19 @@ export class PlaygroundV2Manager implements Disposable {
     const conversation = this.#conversationRegistry.get(conversationId);
     if (conversation === undefined) throw new Error(`Conversation with id ${conversationId} does not exists.`);
 
+    if (content === undefined || content.length === 0) {
+      this.#conversationRegistry.removeMessage(conversationId, conversation.messages[0].id);
+      this.telemetry.logUsage('playground.system-prompt.delete');
```
jeffmaury, line 167: Same remark

packages/backend/src/managers/playgroundV2Manager.ts
```diff
@@ -216,13 +251,17 @@ export class PlaygroundV2Manager implements Disposable {
    * @param stream
    */
   private async processStream(conversationId: string, stream: Stream<ChatCompletionChunk>): Promise<void> {
+    const conversation = this.#conversationRegistry.get(conversationId);
+    if (conversation === undefined) throw new Error(`conversation with id ${conversationId} not found.`);
```
jeffmaury, line 255: This error is not being catched by the telemetry would be a good plus

axel7083: I took the liberty to improve the code by moving the logic of always checking `if undefined...` to the conversation registry, in the end simplifying both the PlaygroundManager and the ConversationRegistry

packages/backend/src/managers/playgroundV2Manager.ts
axel7083: Fixed

packages/backend/src/managers/playgroundV2Manager.ts
axel7083: Fixed

jeffmaury (approved):
LGTM

PR #856, opened 2024-04-09
feat: show warning if no machine is running or there are not enough resources (#227)

### LINKED ISSUE #227: Warn users if they are using a podman machine with few resources
It may happen that users want to play with ai-studio but they have a machine with few resources.
We should show a warning somewhere to explain that having less than Y GB and X CPUs could make the recipe run extremely slow and it is recommended to create a new machine with more resources.

- [x] #873
- [x] #874

### LINKED ISSUE #873: Warn users if they are using a podman machine with few resources : inference server

lstocchi:
### What does this PR do?

It shows a warning banner if the running podman machine does not have enough resources for the model selected by the user or if there is no running machine.

This needs https://github.com/containers/podman-desktop/pull/6733 to work

This PR only covers the inference server page to make the review easier. Going to open a new PR for the recipe page once it is merged

### Screenshot / video of UI

This gif shows if you can edit the machine and if there is no running machine
warning_banner_1

This is the message if the edit action is not enabled for the system (like WSL)
image

Fixes #873 

### How to test this PR?

1. if you are on WSL you need to hack the result received from the system as it takes the resources from the host so you have to load the machine a lot to see the warning message
2. on MAC by creating a machine with few resources you should see the banner

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -54,7 +54,7 @@
     "xml-js": "^1.6.11"
   },
   "devDependencies": {
-    "@podman-desktop/api": "^1.9.0",
+    "@podman-desktop/api": "0.0.202404101645-5d46ba5",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^20",
     "@types/postman-collection": "^3.5.10",
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -43,6 +43,8 @@ import type { Language } from 'postman-code-generators';
 import type { ModelOptions } from '@shared/src/models/IModelOptions';
 import type { CancellationTokenRegistry } from './registries/CancellationTokenRegistry';
 import type { LocalModelImportInfo } from '@shared/src/models/ILocalModelInfo';
+import { checkContainerConnectionStatusAndResources, getPodmanConnection } from './utils/podman';
+import type { ContainerConnectionInfo } from '@shared/src/models/IContainerConnectionInfo';
 
 interface PortQuickPickItem extends podmanDesktopApi.QuickPickItem {
   port: number;
@@ -248,6 +250,15 @@ export class StudioApiImpl implements StudioAPI {
     const pod = pods.find(pod => pod.Id === podId);
     if (pod === undefined) throw new Error(`Pod with id ${podId} not found.`);
     return podmanDesktopApi.navigation.navigateToPod(pod.kind, pod.Name, pod.engineId);
+  }
+
+  async navigateToResources(): Promise<void> {
+    return podmanDesktopApi.navigation.navigateToResources();
+  }
+
+  async navigateToEditConnectionProvider(connectionName: string): Promise<void> {
+    const connection = getPodmanConnection(connectionName);
+    return podmanDesktopApi.navigation.navigateToEditProviderContainerConnection(connection);
   }
 
   async getApplicationsState(): Promise<ApplicationState[]> {
@@ -466,5 +477,9 @@ export class StudioApiImpl implements StudioAPI {
 
   copyToClipboard(content: string): Promise<void> {
     return podmanDesktopApi.env.clipboard.writeText(content);
+  }
+  
+  async checkContainerConnectionStatusAndResources(memory: number): Promise<ContainerConnectionInfo> {
+    return checkContainerConnectionStatusAndResources(memory);
   }
 }
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -18,7 +18,7 @@
 
 import type { StudioAPI } from '@shared/src/StudioAPI';
 import type { ApplicationManager } from './managers/applicationManager';
-import type { ModelInfo } from '@shared/src/models/IModelInfo';
+import type { ModelCheckerInfo, ModelInfo } from '@shared/src/models/IModelInfo';
 import * as podmanDesktopApi from '@podman-desktop/api';
 
 import type { CatalogManager } from './managers/catalogManager';
@@ -485,7 +485,7 @@ export class StudioApiImpl implements StudioAPI {
     return podmanDesktopApi.env.clipboard.writeText(content);
   }
   
-  async checkContainerConnectionStatusAndResources(memory: number): Promise<ContainerConnectionInfo> {
-    return checkContainerConnectionStatusAndResources(memory);
+  async checkContainerConnectionStatusAndResources(modelInfo: ModelCheckerInfo): Promise<ContainerConnectionInfo> {
+    return checkContainerConnectionStatusAndResources(modelInfo);
   }
 }
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -484,7 +484,7 @@ export class StudioApiImpl implements StudioAPI {
   copyToClipboard(content: string): Promise<void> {
     return podmanDesktopApi.env.clipboard.writeText(content);
   }
-  
+
   async checkContainerConnectionStatusAndResources(modelInfo: ModelCheckerInfo): Promise<ContainerConnectionInfo> {
     return checkContainerConnectionStatusAndResources(modelInfo);
   }
--- a/packages/backend/src/studio-api-impl.ts
+++ b/packages/backend/src/studio-api-impl.ts
@@ -253,12 +253,18 @@ export class StudioApiImpl implements StudioAPI {
   }
 
   async navigateToResources(): Promise<void> {
-    return podmanDesktopApi.navigation.navigateToResources();
+    // navigateToResources is only vailable from desktop 1.10
+    if (podmanDesktopApi.navigation.navigateToResources) {
+      return podmanDesktopApi.navigation.navigateToResources();
+    }
   }
 
   async navigateToEditConnectionProvider(connectionName: string): Promise<void> {
-    const connection = getPodmanConnection(connectionName);
-    return podmanDesktopApi.navigation.navigateToEditProviderContainerConnection(connection);
+    // navigateToEditProviderContainerConnection is only vailable from desktop 1.10
+    if (podmanDesktopApi.navigation.navigateToEditProviderContainerConnection) {
+      const connection = getPodmanConnection(connectionName);
+      return podmanDesktopApi.navigation.navigateToEditProviderContainerConnection(connection);
+    }
   }
 
   async getApplicationsState(): Promise<ApplicationState[]> {
--- a/packages/backend/src/utils/podman.ts
+++ b/packages/backend/src/utils/podman.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 import type { ProviderContainerConnection } from '@podman-desktop/api';
-import { configuration, containerEngine, env, process, provider } from '@podman-desktop/api';
+import { configuration, containerEngine, env, navigation, process, provider } from '@podman-desktop/api';
 import type { ContainerConnectionInfo } from '@shared/src/models/IContainerConnectionInfo';
 
 export const MIN_CPUS_VALUE = 10;
@@ -143,9 +143,13 @@ export async function checkContainerConnectionStatusAndResources(
     console.log(String(e));
   }
 
+  // starting from podman desktop 1.10 we have the navigate functions
+  const hasNavigateFunction = !!navigation.navigateToResources;
+
   if (!connection) {
     return {
       status: 'no-machine',
+      canRedirect: hasNavigateFunction,
     };
   }
 
@@ -153,6 +157,7 @@ export async function checkContainerConnectionStatusAndResources(
   if (!engineInfo) {
     return {
       status: 'no-machine',
+      canRedirect: hasNavigateFunction,
     };
   }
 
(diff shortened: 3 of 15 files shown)


packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.svelte
```diff
@@ -0,0 +1,94 @@
...
+      }
+    } else {
+      machineCurrentStateDescription += `${filesize(connectionInfo.memoryIdle, { base: 2 })} of memory available`;
+      machinePreferredStateDescription += `${filesize(connectionInfo.memoryExpected, { base: 2 })} of memory`;
+    }
+
+    const machineName = `${connectionInfo.name.includes('Podman Machine') ? connectionInfo.name : `Podman Machine ${connectionInfo.name}`}`;
+    description = `Your ${machineName} has ${machineCurrentStateDescription}.`;
+
+    if (connectionInfo?.canEdit) {
```
axel7083, line 49: Would `upgradeRecommended` be more representative than `canEdit` ? I feel like canEdit would mean if we are able or not to edit the podman machine, rather that should be edit for better performance

packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.svelte
```diff
@@ -0,0 +1,94 @@
+<script lang="ts">
```
axel7083, line 1: Could we have some tests ?

packages/backend/src/studio-api-impl.ts
```diff
@@ -235,6 +237,15 @@ export class StudioApiImpl implements StudioAPI {
     return podmanDesktopApi.navigation.navigateToPod(pod.kind, pod.Name, pod.engineId);
   }
 
+  async navigateToResources(): Promise<void> {
+    return podmanDesktopApi.navigation.navigateToResources();
+  }
+
+  async navigateToEditConnectionProvider(connectionName: string): Promise<void> {
+    const connection = getPodmanConnection(connectionName);
+    return podmanDesktopApi.navigation.navigateToEditProviderContainerConnection(connection);
```
axel7083, line 246: This API would not be available in version prior to 1.10.0, since we are supporting version from 1.8.0 we need to have a mechanism to not show the redirect button I guess ?

packages/shared/src/models/IContainerConnectionInfo.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export type ContainerConnectionInfo =
+  | RunningContainerConnection
+  | NotEnoughResourcesContainerConnection
+  | NoContainerConnection;
+
+export type ContainerConnectionInfoStatus = 'running' | 'no-machine' | 'not-enough-resources';
```
axel7083, line 24: `not-enough-resources` seems a bit rash, maybe `low-resources` ?

axel7083: For specific models, like embedding which requires almost no ram, a few GB of RAM are enough

packages/backend/src/utils/podman.ts
```diff
@@ -118,3 +132,48 @@ export async function isQEMUMachine(): Promise<boolean> {
...
+  }
+
+  const engineInfo = await containerEngine.info(`${connection.providerId}.${connection.connection.name}`);
+  if (!engineInfo) {
+    return {
+      status: 'no-machine',
+    };
+  }
+
+  const hasCpus = engineInfo.cpus && engineInfo.cpus >= MIN_CPUS_VALUE;
```
axel7083, line 159: How do handle the case where the podman machine already has the maximum than the host, but still fell bellow the MIN you choice ?

packages/frontend/src/lib/ContentDetailsLayout.svelte
```diff
@@ -8,33 +8,38 @@ const toggle = () => {
 };
 </script>
 
-<div class="grid w-full overflow-y-auto lg:grid-cols-[1fr_auto] max-lg:grid-cols-[auto]">
-  <div class="p-5 inline-grid">
-    <slot name="content" />
+<div class="flex flex-col w-full overflow-y-auto">
```
axel7083, line 11: You seems to be changing the layout, could you make it in another PR to ease the review and tests ?

packages/shared/src/StudioAPI.ts
```diff
@@ -164,4 +167,10 @@ export abstract class StudioAPI {
    * @param tokenId the id of the CancellationToken to cancel
    */
   abstract requestCancelToken(tokenId: number): Promise<void>;
+
+  /**
+   * Check if the running podman machine is running and has enough resources to execute task
+   * @param memory amount of memory that must be idle
+   */
+  abstract checkContainerConnectionStatusAndResources(memory: number): Promise<ContainerConnectionInfo>;
```
axel7083, line 175: 👍 thanks for the doc!

packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.svelte
```diff
@@ -0,0 +1,94 @@
...
+function updateTitleDescription(connectionInfo: ContainerConnectionInfo) {
+  if (connectionInfo.status === 'no-machine') {
+    title = 'No Podman machine is running';
+    description = 'Please start a Podman Machine before proceeding further.';
+    actionName = 'Start now';
+    return;
+  }
+
+  if (connectionInfo.status === 'not-enough-resources') {
+    title = 'Upgrade your Podman machine for best AI performance';
```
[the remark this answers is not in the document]
lstocchi: I just copied the title from Mo's mockup. 
I'm ok with changing it and with both of your suggestions. Maybe @vrothberg can give some help here

packages/backend/src/studio-api-impl.ts
lstocchi: Yes, i'm going to hide the button. Still in progress

packages/backend/src/utils/podman.ts
lstocchi: I think we can discuss about the minimum value but as we're not preventing the user from working even if its system does not have the minimum resources, we should be ok with it. If you have a machine with 6cpus and we suggest to have a machine with 10, you just keep doing your work ignoring it. If then, you see any performance issue, you know what could be the problem. Maybe we can add a setting to hide the warnings? 🤷‍♂️

packages/frontend/src/lib/ContentDetailsLayout.svelte
lstocchi: Sure, np. I'm going to show the warning on the recipe page in another PR

packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.svelte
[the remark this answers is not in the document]
jeffmaury: Upgrade seems to me related to software update, would prefer edit or update

packages/backend/src/utils/podman.ts
```diff
@@ -16,7 +16,10 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 import type { ProviderContainerConnection } from '@podman-desktop/api';
-import { configuration, env, process, provider } from '@podman-desktop/api';
+import { configuration, containerEngine, env, navigation, process, provider } from '@podman-desktop/api';
+import type { ContainerConnectionInfo } from '@shared/src/models/IContainerConnectionInfo';
+
+export const MIN_CPUS_VALUE = 10;
```
jeffmaury, line 22: Wondering if this is not quite a high value

lstocchi: I just copied what it was on the mockup. Which value should i set? 4-6-8?

jeffmaury:
Another general comment: here we are comparing the amout of free memory to the memory of the model but I think a little bit is required to run container(s). So I would apply a factor to the model memory before comparing to the free memory: what about:

- +10% for inference server
- + 25% for recipe (as we have several containers)

packages/backend/src/utils/podman.ts
jeffmaury: I would choose 4 but let me give a try

lstocchi:
> Another general comment: here we are comparing the amout of free memory to the memory of the model but I think a little bit is required to run container(s). So I would apply a factor to the model memory before comparing to the free memory: what about:
> 
> * +10% for inference server
> * * 25% for recipe (as we have several containers)

25% seems a bit high to me. If you're using a model of 10GB it's like 2.5Gb more. Maybe we should have a threshold. Like 1GB. So if 25% is higher than 1GB we stick to 1GB, otherwise we use the 25%??

packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  const banner = screen.getByLabelText('Container connection info banner');
+  expect(banner).toBeInTheDocument();
+  const titleDiv = screen.getByLabelText('title');
+  expect(titleDiv).toBeInTheDocument();
+  expect(titleDiv.textContent).equals('Upgrade your Podman machine for best AI performance');
+  const descriptionDiv = screen.getByLabelText('description');
+  expect(descriptionDiv).toBeInTheDocument();
+  expect(descriptionDiv.textContent).equals(
+    `Your Podman Machine has ${connectionInfo.cpus} vCPUs and ${filesize(connectionInfo.memoryIdle, { base: 2 })} of memory available. We recommend upgrading your Podman machine with at least ${connectionInfo.cpusExpected} vCPUs and ${filesize(connectionInfo.memoryExpected, { base: 2 })} of memory for better AI performance.`,
```
vrothberg, line 123:
```suggestion
    `Your Podman Machine has ${connectionInfo.cpus} vCPUs and ${filesize(connectionInfo.memoryIdle, { base: 2 })} of memory available. We recommend upgrading your Podman Machine to at least ${connectionInfo.cpusExpected} vCPUs and ${filesize(connectionInfo.memoryExpected, { base: 2 })} of memory for better AI performance.`,
```

Should we also add link to the settings menu?  I guess that most users don't know how to upgrade the resources.

lstocchi: There is already the button on the right (when we can show it), isn't that enough?

The different scenario are:
1. there is no podman machine running -> there is a `Start now` button that redirects the user to the resources page where they can see all podman machines available and start/create one
2. there is a podman machine but it does not have the resources we recommend ->
  2.1 if on mac -> we show the 'Update now' button that redirects the user to the page to edit the podman machine resources inside podman desktop
  2.2 if on windows -> there is no button as WSL uses the resources from the host and there is no reason to update the podman machine

lstocchi: Edit done 👍 I also replaced `Upgrade` with `Update` as Jeff asked for the change

packages/backend/src/utils/podman.ts
lstocchi: Edited with 4 for the moment, gonna give it a try on the mac to see if it runs fine

lstocchi:
> Another general comment: here we are comparing the amout of free memory to the memory of the model but I think a little bit is required to run container(s). So I would apply a factor to the model memory before comparing to the free memory: what about:
> 
> * +10% for inference server
> * * 25% for recipe (as we have several containers)

used these values for the moment. We can update them later if we see some issue

packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.svelte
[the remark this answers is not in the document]
lstocchi: I used the last sentence proposed by Axel and replace Upgrade with Update
`Update your Podman machine to improve performance`

packages/backend/package.json
```diff
@@ -54,7 +54,7 @@
     "xml-js": "^1.6.11"
   },
   "devDependencies": {
-    "@podman-desktop/api": "^1.9.0",
+    "@podman-desktop/api": "0.0.202404101645-5d46ba5",
```
axel7083, line 57: We are always updating it, can't we just use `next` @benoitf ?

packages/frontend/src/lib/notification/ContainerConnectionStatusInfo.spec.ts
vrothberg: I totally missed the button for some reason.  Looks great, thank you!

lstocchi:
> Do we need a similar warning when starting an AI App? Currently, I don't get one when starting the ChatBot.

Not in this PR. I'm gonna create a new one when this is merged. But yes, there will be a similar warning when starting an AI app

vrothberg (commented):
LGTM

lstocchi:
I used the values as we do in desktop for the warning banner and this is the result 
(the text and button are a sample, that text won't appear when the update button is visible)

image

jeffmaury:
One comment as I setup a small Podman machine so got the warning, updated the machine, so the warning disappear, then the model service is created, but once the service is created and running the warning is displayed again. Sounds strange but not false

axel7083 (commented):
Nice ! I really like with the warning instead of the error! Some remarks

- I see the banner inside the inference server creation form but not in the recipe page 
- I am using latest version of podman, should I see the upgrade button ? Because I am not

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/806f2250-1243-422a-80d8-531a6c16aa0c

jeffmaury:
> Nice ! I really like with the warning instead of the error! Some remarks
> 
>     * I see the banner inside the inference server creation form but not in the recipe page
> 
>     * I am using latest version of podman, should I see the upgrade button ? Because I am not
> 
> 
> warning-msg.mp4

- There's a follow up PR for the recipe page (based ion your requirement for review :))- 
- Windows WSL does not support edit

Issue #858, opened 2024-04-09
Add syntax highlighting to client code in service details


ckyrouac:
It would be nice to have basic syntax highlighting of the client code in the service details page to make the code easier to read.

![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/2886111/551bad4a-fafd-43b4-b7fa-eec32f4f5c83)

PR #863, opened 2024-04-10
feat: using svelte highlight library for code snippets

axel7083:
### What does this PR do?

Using [svelte-highlight](https://www.npmjs.com/package/svelte-highlight) library to perform synthaxe highlight for the code snippet. This library is also very easy to use out of the box.

> Why not using Monaco ? Because it is an heavy library, with too much feature for just a read only code snippet highlight.

### Screenshot / video of UI

| Before | After |
| --- | --- |
| image | image |
| image | image |

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/858

### How to test this PR?

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -36,6 +36,7 @@
     "postcss-load-config": "^5.0.3",
     "svelte": "4.2.12",
     "svelte-fa": "^4.0.2",
+    "svelte-highlight": "^7.6.0",
     "svelte-markdown": "^0.4.1",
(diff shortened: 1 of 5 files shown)


benoitf:
I think that before bringing a new library, we need to step back and see if we're not already using other libraries, if it's the best choice, etc. if each extension comes with its own library to do something it might become challenging.
(for example docusaurus is using prism, Desktop is using Monaco)

Not saying it's not the good choice, but it requires formal discussions.

axel7083:
Closing as require more discussion

PR #868, opened 2024-04-10
fix: adding chatformat to use for inference servers

axel7083:
### What does this PR do?

This PR is adding the optional `chatformat` to the model object.

Need https://github.com/containers/podman-desktop-extension-ai-lab-playground-images/pull/11 to be merged, and a new version published of the image to be working.

Fixing the playgrounds for the following models
- Merlinite
- Cerebrum
- dragon-mistral

### Screenshot / video of UI

image

image

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/825

### How to test this PR?

### DIFF
--- a/packages/backend/src/utils/inferenceUtils.ts
+++ b/packages/backend/src/utils/inferenceUtils.ts
@@ -34,7 +34,7 @@ export const SECOND: number = 1_000_000_000;
 export const LABEL_INFERENCE_SERVER: string = 'ai-lab-inference-server';
 
 export const INFERENCE_SERVER_IMAGE =
-  'ghcr.io/containers/podman-desktop-extension-ai-lab-playground-images/ai-lab-playground-chat:0.2.0';
+  'ghcr.io/containers/podman-desktop-extension-ai-lab-playground-images/ai-lab-playground-chat:0.3.0';
 
 /**
  * Return container connection provider
--- a/packages/backend/src/utils/inferenceUtils.ts
+++ b/packages/backend/src/utils/inferenceUtils.ts
@@ -116,7 +116,7 @@ export function generateContainerCreateOptions(
   }
 
   const envs: string[] = [`MODEL_PATH=/models/${modelInfo.file.file}`, 'HOST=0.0.0.0', 'PORT=8000'];
-  if(modelInfo.chatformat) {
+  if (modelInfo.chatformat) {
     envs.push(`CHAT_FORMAT=${modelInfo.chatformat}`);
   }
 
--- a/packages/backend/src/utils/inferenceUtils.ts
+++ b/packages/backend/src/utils/inferenceUtils.ts
@@ -117,10 +117,12 @@ export function generateContainerCreateOptions(
 
   const envs: string[] = [`MODEL_PATH=/models/${modelInfo.file.file}`, 'HOST=0.0.0.0', 'PORT=8000'];
   if (modelInfo.properties) {
-    envs.push(...Object.entries(modelInfo.properties).map(([key, value]) => {
-      const formattedKey = key.replace(/[A-Z]/g, m => `_${m}`).toUpperCase();
-      return `MODEL_${formattedKey}=${value}`;
-    }));
+    envs.push(
+      ...Object.entries(modelInfo.properties).map(([key, value]) => {
+        const formattedKey = key.replace(/[A-Z]/g, m => `_${m}`).toUpperCase();
+        return `MODEL_${formattedKey}=${value}`;
+      }),
+    );
   }
 
   return {
--- a/packages/backend/src/utils/inferenceUtils.ts
+++ b/packages/backend/src/utils/inferenceUtils.ts
@@ -118,7 +118,8 @@ export function generateContainerCreateOptions(
   const envs: string[] = [`MODEL_PATH=/models/${modelInfo.file.file}`, 'HOST=0.0.0.0', 'PORT=8000'];
   if (modelInfo.properties) {
     Object.entries(modelInfo.properties).forEach(([key, value]) => {
-      envs.push(`${key.toUpperCase()}=${value}`);
+      const formatedKey = key.replace(/[A-Z]/g, m => `_${m}`);
+      envs.push(`${formatedKey}=${value}`);
     });
   }
 
--- a/packages/backend/src/utils/inferenceUtils.ts
+++ b/packages/backend/src/utils/inferenceUtils.ts
@@ -115,6 +115,11 @@ export function generateContainerCreateOptions(
     throw new Error('The model info file provided is undefined');
   }
 
+  const envs: string[] = [`MODEL_PATH=/models/${modelInfo.file.file}`, 'HOST=0.0.0.0', 'PORT=8000'];
+  if(modelInfo.chatformat) {
+    envs.push(`CHAT_FORMAT=${modelInfo.chatformat}`);
+  }
+
   return {
     Image: imageInfo.Id,
     Detach: true,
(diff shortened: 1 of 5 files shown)


packages/backend/src/utils/inferenceUtils.ts
```diff
@@ -115,6 +115,11 @@ export function generateContainerCreateOptions(
     throw new Error('The model info file provided is undefined');
   }
 
+  const envs: string[] = [`MODEL_PATH=/models/${modelInfo.file.file}`, 'HOST=0.0.0.0', 'PORT=8000'];
```
benoitf, line 118: missing a unit test to see if env variable contains CHAT_FORMAT if modelInfo has or not the chatformat attribute

packages/shared/src/models/IModelInfo.ts
```diff
@@ -29,4 +29,9 @@ export interface ModelInfo {
   file?: LocalModelInfo;
   state?: 'deleting';
   memory: number;
+  /**
+   * the chat format to use
+   * @example llama-2
+   */
+  chatformat?: 'openchat' | 'llama-2' | string
```
benoitf, line 36: the `| string` is making the other one useless

axel7083: The `| string` is here to allow arbritrary value, while showing the possible existing supported values

I took inspiration from the OpenAI lib

image

packages/backend/src/utils/inferenceUtils.ts
axel7083: Forgot to push the commit 😪

jeffmaury:
As a general comment I think this is too specific for chat models. We already have the plan to add a type attribute to models (chat, vision,...) so I think it would be better to rename the field to something more general purpose (format ?). And the env variable should be renamed accordingly

jeffmaury:
Was thinking intensively about this during the night: found a more generic way to handle that.
Basically we want to pass model attributes to the containers so let's add an attributes field to the model catalog and this will be a record of values. When launching the container, pass each value of this record as a env variable named MODEL_key where key is the uppercase transcription eg if the key is chatformat then the env var would be MODEL_CHATFORMAT.
This would requires a small change on the recipes but allows for better evolution in the future if new attributes are added.
WDYT ?

axel7083:
> MODEL_

I like this solution but I am not sure about the necessity of the `MODEL_` prefix ?

benoitf:
```json
"properties" : {
  "chatFormat": "openchat"
}
```

jeffmaury:
> > MODEL_
> 
> I like this solution but I am not sure about the necessity of the `MODEL_` prefix ?

The MODEL_ prefix just to standardize on what is provider by AI Lab as we already provider MODEL_PATH and as it's related to the model

jeffmaury:
> > > MODEL_
> > 
> > 
> > I like this solution but I am not sure about the necessity of the `MODEL_` prefix ?
> 
> The MODEL_ prefix just to standardize on what is provided by AI Lab as we already provide MODEL_PATH and as it's related to the model

axel7083:
> ```json
> "properties" : {
>   "chatFormat": "openchat"
> }
> ```
> 
> ?

We are using the `labels` wording in several places, might want to stick with that ?

https://github.com/containers/podman-desktop-extension-ai-lab/blob/1882d0c6e1fc30d90a1f65b190d721dc93459255/packages/shared/src/models/ITask.ts#L27

https://github.com/containers/podman-desktop-extension-ai-lab/blob/5dd413eb2bb84f6ae5ce8317f4f59f133496463a/packages/shared/src/models/ILocalRepository.ts#L4

https://github.com/containers/podman-desktop-extension-ai-lab/blob/039b7ca839dd8a45a8d8d338e869d2eb197ef986/packages/shared/src/models/InferenceServerConfig.ts#L38

jeffmaury:
> > ```json
> > "properties" : {
> >   "chatFormat": "openchat"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> We are using the `labels` wording in several places, might want to stick with that ?
> 
> https://github.com/containers/podman-desktop-extension-ai-lab/blob/1882d0c6e1fc30d90a1f65b190d721dc93459255/packages/shared/src/models/ITask.ts#L27
> 
> https://github.com/containers/podman-desktop-extension-ai-lab/blob/5dd413eb2bb84f6ae5ce8317f4f59f133496463a/packages/shared/src/models/ILocalRepository.ts#L4
> 
> https://github.com/containers/podman-desktop-extension-ai-lab/blob/039b7ca839dd8a45a8d8d338e869d2eb197ef986/packages/shared/src/models/InferenceServerConfig.ts#L38

labels are intended for display I think this is more internal here so I would be more for attributes or properties

axel7083:
> labels are intended for display I think this is more internal here so I would be more for attributes or properties
Go for properties then !

The prefix would not work with https://github.com/containers/podman-desktop-extension-ai-lab-playground-images/pull/11 for now, might want to not having it for now ?

axel7083:
> ```json
> "properties" : {
>   "chatFormat": "openchat"
> }
> ```

`chat_format` is easier to transform to `CHAT_FORMAT` than `chatFormat`, we jsut have to use uppercase() for the first one

benoitf:
I think that I expect camelCase in a json file also about easier I'm not sure it would be much harder anyway

```ts
'chatFormat'.replace(/[A-Z]/g, m => `_${m}`).toUpperCase() => CHAT_FORMAT
```

axel7083:
> I think that I expect camelCase in a json file also about easier I'm not sure it would be much harder anyway

No problem then

packages/backend/src/utils/inferenceUtils.ts
```diff
@@ -115,6 +115,14 @@ export function generateContainerCreateOptions(
...
+  if (modelInfo.properties) {
+    Object.entries(modelInfo.properties).forEach(([key, value]) => {
+      const formatedKey = key.replace(/[A-Z]/g, m => `_${m}`);
+      envs.push(`${formatedKey}=${value}`);
+    });
```
benoitf, line 123: looks like formatedKey has a typo (missing two t) but it's not really the matter
AFAIK it's missing the toUppercase
FYI you can have an array from entries, so no need to have a forEach loop as you can from transformation have an array and then you can destructure the array with `...`

so something like (not tested)
```ts
const additionalEnvs = Array.from(Object.entries(modelInfo.attributes)).map((e) => `${e[0].replace(/[A-Z]/g, m => `_${m}`).toUpperCase()}=${e[1]}`)
envs.push(...additionalEnvs);
```

axel7083:
````ts
const additionalEnvs = Array.from(Object.entries(modelInfo.attributes)).map((e) => `${e[0].replace(/[A-Z]/g, m => `_${m}`).toUpperCase()}=${e[1]}`)
envs.push(...additionalEnvs);
````

I am thinking that is hard to read and understand for an external reader

benoitf: well you can extract key and value and assemble them in the map function but it's still using stream pattern and not loop pattern.

jeffmaury:
> > labels are intended for display I think this is more internal here so I would be more for attributes or properties
> > Go for properties then !
> 
> The prefix would not work with [containers/podman-desktop-extension-ai-lab-playground-images#11](https://github.com/containers/podman-desktop-extension-ai-lab-playground-images/pull/11) for now, might want to not having it for now ?

I would favor (and not delay) the prefix is the ai-recipes team is ok with this change, starting the discussion on their channel

packages/backend/src/utils/inferenceUtils.ts
benoitf: thanks

benoitf:
so here we need to update https://github.com/containers/podman-desktop-extension-ai-lab-playground-images/pull/11/files to use MODEL_ env var

jeffmaury (approved):
LGTM. Recipe repo has been updated so we should be good

benoitf:
before merging, https://github.com/containers/podman-desktop-extension-ai-lab/blob/1f4223af76651322091ebd26809211d6d1167d12/packages/backend/src/utils/inferenceUtils.ts#L37 need to be updated and a new release should happen in the https://github.com/containers/podman-desktop-extension-ai-lab-playground-images repo

benoitf (approved):

jeffmaury (approved):
LGTM

PR #902, opened 2024-04-16
fix: cannot delete the model with operation not permitted

### LINKED ISSUE #730: Cannot delete the model with operation not permitted

jeffmaury:
Fixes #730

### What does this PR do?

Fixes error spawn when deleting a just downloaded model on Windows

### Screenshot / video of UI

N/A

#730 

### How to test this PR?

1. Download a model
2. Once it is downloaded, delete it

### DIFF
--- a/packages/backend/src/utils/downloader.ts
+++ b/packages/backend/src/utils/downloader.ts
@@ -87,10 +87,6 @@ export class Downloader {
   private followRedirects(url: string, callback: (message: { ok?: boolean; error?: string }) => void): void {
     const tmpFile = `${this.target}.tmp`;
 
-    const stream = createWriteStream(tmpFile, {
-      signal: this.abortSignal,
-    });
-
     let totalFileSize = 0;
     let progress = 0;
     let previousProgressValue = -1;
@@ -105,6 +101,10 @@ export class Downloader {
       if (totalFileSize === 0 && resp.headers['content-length']) {
         totalFileSize = parseFloat(resp.headers['content-length']);
       }
+
+      const stream = createWriteStream(tmpFile, {
+        signal: this.abortSignal,
+      });
 
       // Capture potential errors
       resp.on('error', (err: Error) => {


axel7083 (commented):
What was the root of the issue ? I am not sure to follow why it fixes the problem to just move it inside the on callback ?

jeffmaury:
> What was the root of the issue ? I am not sure to follow why it fixes the problem to just move it inside the on callback ?

The file was opened twice: the first time when the redirect response was received (we opened the file but to do use the handle so it stays open) and a second time after the redirection is processed. So at then end the second handle is closed but not the first one. The path delay the creation of the handle when it's really needed

axel7083:
> > What was the root of the issue ? I am not sure to follow why it fixes the problem to just move it inside the on callback ?
> 
> The file was opened twice: the first time when the redirect response was received (we opened the file but to do use the handle so it stays open) and a second time after the redirection is processed. So at then end the second handle is closed but not the first one. The fix delay the creation of the handle when it's really needed

But, this is only valid for the `.tmp` file, as when the download is finish we rename it. The deletion issue was on the renamed file ?

jeffmaury:
> > > What was the root of the issue ? I am not sure to follow why it fixes the problem to just move it inside the on callback ?
> > 
> > 
> > The file was opened twice: the first time when the redirect response was received (we opened the file but to do use the handle so it stays open) and a second time after the redirection is processed. So at then end the second handle is closed but not the first one. The fix delay the creation of the handle when it's really needed
> 
> But, this is only valid for the `.tmp` file, as when the download is finish we rename it. The deletion issue was on the renamed file ?

Yes but it fixes the issue so I think Windows is now smarter than it was and it moves the open handles to the renamed file

lstocchi (approved):
Tried to delete a model after having started/stopped a recipe and an inference server. Works fine 🚀

axel7083 (approved):

PR #904, opened 2024-04-16
feat: provides model properties to model service while running an AI App

### LINKED ISSUE #903: Provides model properties to model service while running an AI App
They are provided only for the inference server

jeffmaury:
Fixes #903

### What does this PR do?

Allow to pass model properties as environment variables when starting an AI App like it is done for an inference server

### Screenshot / video of UI

N/A

#903

### How to test this PR?

1. Start a recipe with a model with properties
2. Enter the model server container
3. Verify the environment variable is set

### DIFF
--- a/packages/backend/src/utils/modelsUtils.ts
+++ b/packages/backend/src/utils/modelsUtils.ts
@@ -71,3 +71,14 @@ export async function deleteRemoteModel(machine: string, modelInfo: ModelInfo): 
     console.error('Something went wrong while trying to stat remote model path', err);
   }
 }
+
+export function addModelPropertiesToEnvironment(modelInfo: ModelInfo, envs: string[]) {
+  if (modelInfo.properties) {
+    envs.push(
+      ...Object.entries(modelInfo.properties).map(([key, value]) => {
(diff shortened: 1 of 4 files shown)


packages/backend/src/utils/modelsUtils.ts
```diff
@@ -71,3 +71,14 @@ export async function deleteRemoteModel(machine: string, modelInfo: ModelInfo):
     console.error('Something went wrong while trying to stat remote model path', err);
   }
 }
+
+export function addModelPropertiesToEnvironment(modelInfo: ModelInfo, envs: string[]) {
```
axel7083, line 75: i would prefer a method that does not change the parameter it receive, and have something which return the environements variable.

So instead of having
```
addModelPropertiesToEnvironment(modelInfo, envs);
```

I would rather have

```
envs.push(...getModelPropertiesForEnvironment(modelInfo));
```

lstocchi (approved):
Tested and works as expected

axel7083 (approved):
Thanks for the change 🚀 ! LGTM

PR #946, opened 2024-04-22
fix: playground improvements for keyboard usage

### LINKED ISSUE #839: Playground improvements for keyboard usage
When I click on a playground I'm landing to a page

1. The prompt is at the bottom but is not focused so if I try to type something, nothing happens (I expect to have the blinking cursor)
1. If I use `ENTER` key after typing something, it adds a newline ? But usually on chat app, it's sending the message right away. (For a newline I use most of the time `MAJ` + `ENTER` key

I still have an issue with the prompt being collapsed to the left side and using the same color than the left navbar and not being like detached

currently it is like:
image

I expect to see something like:

image

or even more like chat apps where we have the 'send message' icon like

image

jeffmaury:
Fixes #839

### What does this PR do?

Gives focus to the prompt and handle Enter key

### Screenshot / video of UI

playground_enter

#839 

### How to test this PR?

1. Create a playground
2. Enter text + enter => should send the request

### DIFF
--- a/packages/frontend/src/pages/Playground.spec.ts
+++ b/packages/frontend/src/pages/Playground.spec.ts
@@ -348,6 +348,51 @@ test('sending prompt should disable the send button', async () => {
   });
 });
 
+test('sending prompt not using button should disable the send button', async () => {
+  vi.mocked(studioClient.getCatalog).mockResolvedValue({
+    models: [
+      {
+        id: 'model-1',
+        name: 'Model 1',
+      },
+    ] as ModelInfo[],
+    recipes: [],
+    categories: [],
+  });
+  vi.mocked(studioClient.submitPlaygroundMessage).mockResolvedValue();
+  const customConversations = writable<Conversation[]>([
+    {
+      id: 'playground-1',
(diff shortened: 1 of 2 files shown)


benoitf:
to me it fixes the keyboard issue but the layout is still not the expected one

I expect the button being part of the widget and being centered and not attached to the left

image

benoitf:
on my side the prompt is the same color than the background

image

lstocchi (commented):
In my case, when i click Enter, the message is sent correctly but then there is an active new line in the textarea that prevents the plaholder to be seen.

lstocchi:
@benoitf Jeff asked me to fix this as he's on PTO. If it looks good for you we can merge

lstocchi:
> In my case, when i click Enter, the message is sent correctly but then there is an active new line in the textarea that prevents the plaholder to be seen.

fixed

benoitf:
@lstocchi yes, probably need to update the screenshot/gif of the PR body

lstocchi:
@benoitf updated
Sorry, i always forgot 😵‍💫

PR #954, opened 2024-04-22
chore: no implicit any linter rule

axel7083:
### What does this PR do?

Adding `noImplicitAny` linter rule

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/953

### How to test this PR?

Pipeline should be ✅

### DIFF
--- a/packages/backend/src/models/AIConfig.ts
+++ b/packages/backend/src/models/AIConfig.ts
@@ -52,11 +52,25 @@ export function assertString(value: unknown): string {
 export function parseYamlFile(filepath: string, defaultArch: string): AIConfig {
   const raw: string = fs.readFileSync(filepath, 'utf-8');
 
-  const aiLabConfig = jsYaml.load(raw);
-  const application = aiLabConfig?.['application'];
-  if (!application) throw new Error('AIConfig has bad formatting.');
+  const aiLabConfig: unknown = jsYaml.load(raw);
+  if (!aiLabConfig || typeof aiLabConfig !== 'object') {
+    throw new Error('malformed configuration file.');
+  }
 
-  const containers: unknown[] = application['containers'] ?? [];
+  if (!('application' in aiLabConfig)) {
+    throw new Error('AIConfig has bad formatting: missing application property');
+  }
+
+  const application: unknown = aiLabConfig['application'];
+  if (!application || typeof application !== 'object' || !('containers' in application)) {
+    throw new Error('AIConfig has bad formatting: application does not have valid container property');
+  }
+
+  if (!Array.isArray(application['containers'])) {
+    throw new Error('AIConfig has bad formatting: containers property must be an array.');
+  }
+
+  const containers: unknown[] = application['containers'];
 
   return {
     application: {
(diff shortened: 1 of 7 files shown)


packages/backend/src/models/AIConfig.ts
```diff
@@ -70,15 +84,35 @@ export function parseYamlFile(filepath: string, defaultArch: string): AIConfig {
...
+          architectures.push(defaultArch);
+        } else if (Array.isArray(container['arch']) && container['arch'].every(arch => typeof arch === 'string')) {
+          architectures.push(...container['arch']);
+        } else if (typeof container['arch'] === 'string') {
+          architectures.push(container['arch']);
+        } else {
+          throw new Error('malformed arch property');
+        }
+
+        let contanerfile: string | undefined = undefined;
```
benoitf, line 98: there is a typo on containerfile

packages/backend/src/models/AIConfig.ts
```diff
@@ -52,11 +52,25 @@ export function assertString(value: unknown): string {
...
-  const aiLabConfig = jsYaml.load(raw);
-  const application = aiLabConfig?.['application'];
-  if (!application) throw new Error('AIConfig has bad formatting.');
+  const aiLabConfig: unknown = jsYaml.load(raw);
+  if (!aiLabConfig || typeof aiLabConfig !== 'object') {
+    throw new Error('malformed configuration file.');
+  }
 
-  const containers: unknown[] = application['containers'] ?? [];
+  if (!('application' in aiLabConfig)) {
```
benoitf, line 60: I think it will help if you introduce a type for the aiLabConfig object

and then do use the type guards

like 

if (!isAiLabConfig(obj)) {
 throw new Error
} 

and you do the check using the https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

packages/backend/src/models/AIConfig.ts
```diff
@@ -70,15 +84,35 @@ export function parseYamlFile(filepath: string, defaultArch: string): AIConfig {
...
+          containerfile: contanerfile,
           contextdir: contextdir,
-          name: assertString(container['name']),
-          gpu_env: Array.isArray(container['gpu-env']) ? container['gpu-env'] : [],
-          ports: Array.isArray(container['ports']) ? container['ports'] : [],
-          image: isString(container['image']) ? container['image'] : undefined,
+          name: container['name'],
+          gpu_env: 'gpu-env' in container && Array.isArray(container['gpu-env']) ? container['gpu-env'] : [],
+          ports: 'ports' in container && Array.isArray(container['ports']) ? container['ports'] : [],
+          image: 'image' in container && isString(container['image']) ? container['image'] : undefined,
```
benoitf, line 115: not sure it's covered by tests ?

packages/backend/src/models/AIConfig.ts
axel7083: The `parseYamlFile` will be an all in one, as we already have a type `AIConfig` but it does not match the file, and need to be transformed to fill with default values, map properties to other name (E.g. `gpu-env` -> `gpu_env`) etc.

packages/backend/src/models/AIConfig.ts
axel7083: Yeah there was not test for failing cases, only a valid config. I added all the tests needed

packages/backend/src/models/AIConfig.ts
axel7083: thx

benoitf (approved):

Issue #857, opened 2024-04-09
Clicking delete on a running AI app doesn't provide feedback


ckyrouac:
When clicking the Delete Icon in the AI Apps table, it can take some time for the app to be deleted. During this time, there is no indication the app is currently being deleted. On my resource starved machine, the app entry in the table took around 10s to eventually disappear. During that time, I wasn't sure if the delete request worked or not, and almost clicked it a second time. A spinner on the delete icon, or another indication that the delete operation is in progress would be helpful.

axel7083:
Related issue https://github.com/containers/podman-desktop-extension-ai-lab/issues/547

Issue #942, opened 2024-04-22
Service model status is not updated when starting it after it was shut down


cbr7:
It seems that the state icons for service model do not get updated to Running state when the service model is started up after it was shut down. The icons are in a continues starting/loading cycle.

Steps:
1. Create a service model.
2. Shut down the service model create in previous step.
3. Try to start back up the service model.
4. Notice that the status icons are in permanent starting/loading state.

To note that it seems this is just a graphical issue, if the user tries to make a curl query from the CLI the service model will reply as normal even though in PD it shows as starting/loading.

https://github.com/containers/podman-desktop-extension-ai-lab/assets/106525396/abc71d1c-8dec-4704-87df-5c0c1e6152a7

axel7083:
We are using the podman health check, if you can replicates please copy/paste the container inspect result (from container details page).

And the logs of podman desktop also if possible, it might also be a problem from the frontend of the extension, you can see those by 
- right click on the extension ai lab icon of the navigation bar of Podman Desktop (bellow the kubernetes icons)
- Open devtools

axel7083:
I was able to reproduce

ScrewTSW:
Issue was not reproduced in AI Lab 1.1.0 or 1.1.1, closing.

PR #550, opened 2024-03-14
feat(inference): more statuses (`stopping` | `deleting` | `starting`)

axel7083:
### What does this PR do?

Adding more statuses to the inference server. This allow to better handle the transition state and properly display feedback to the user when he clicks on buttons ! 

- [x] Require https://github.com/projectatomic/ai-studio/pull/548 to be merged (need rebase)
 
### Screenshot / video of UI

https://github.com/projectatomic/ai-studio/assets/42176370/1107c540-6d0f-4de4-a401-e37250721f24

### What issues does this PR fix or reference?

Fixes https://github.com/projectatomic/ai-studio/issues/547

### How to test this PR?

- [x] unit tests has been provided

Issue #547, opened 2024-03-14
feat(InferenceServer): Add more status value `deleting` | `stopping` | `starting` for better UI


axel7083:

axel7083:
Fixed by https://github.com/containers/podman-desktop-extension-ai-lab/pull/980

PR #980, opened 2024-04-24
feat(inference): adding transition statuses

axel7083:
### What does this PR do?

Adding transition statuses to the Inference Servers. This work has started with https://github.com/containers/podman-desktop-extension-ai-lab/pull/550 but was delayed to other priorities.

This allows to improve the user experience.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/ef72faa4-737f-4aaa-8c5c-7238123e61b2

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/857 https://github.com/containers/podman-desktop-extension-ai-lab/issues/547 https://github.com/containers/podman-desktop-extension-ai-lab/issues/942

### How to test this PR?

You can test it manually by reproducing the videos above.

### DIFF
--- a/packages/backend/src/managers/inference/inferenceManager.spec.ts
+++ b/packages/backend/src/managers/inference/inferenceManager.spec.ts
@@ -729,6 +729,7 @@ describe('transition statuses', () => {
     const inferenceManager = await getInitializedInferenceManager();
     await inferenceManager.startInferenceServer('dummyId');
 
+    // first status must be set to starting
     expect(webviewMock.postMessage).toHaveBeenCalledWith({
       id: Messages.MSG_INFERENCE_SERVERS_UPDATE,
       body: [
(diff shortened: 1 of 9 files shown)


vrothberg (commented):
The transition works from running->stopped but not the other way around from stopped->running.  Can we add the same there?

I also noticed that the icon doesn't turn green once a service transi
Screenshot 2024-04-24 at 17 48 14
tioned from stopped->running:

axel7083:
> The transition works from running->stopped but not the other way around from stopped->running. Can we add the same there?

Thanks for catching that, the transition was not correct for the UI. Here  are the statuses table

| screenshots | status |
| --- | --- |
| image | `healthy` |
| image | `stopped` |
| image | `running` with `undefined` health check or `running` with `starting` health check |
| image | `starting` or `deleting` or `stopping` |
| image | 'error` or health check `unhealthy`
 
And after the fix the start of an inference look like the following

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/1c279586-9e82-4ea0-87de-ea91a58282d0

> I also noticed that the icon doesn't turn green once a service transitioned from stopped->running

We have a small difference here, the service will only turn green when the health check is `healthy`. The running state is not the condition for it to become green.

vrothberg (approved):
LGTM, thanks for the update 🏁

PR #1020, opened 2024-04-30
Remove system prompt from playground creation page

### LINKED ISSUE #660: Remove system prompt from playground creation page
We would like to improve the experience on creating a playground and defining the system prompt.

**First time user**
We are asking the user to define something that is not yet known. The change it will do to the way the model interact will not be understood AND it will not be editable "on the go"

--> The system prompt should be treated as an advanced setting in the creation of the playground environment 
Screenshot 2024-03-25 at 11 16 45

Right now:
- The user does not what's mandatory from what's optional
- The user will be intrigued by the "system prompt", while it's not something needed for the first time experience -> adding more cognitive load

**User interacting with the model**
As a user I've interacting with the model and I'm writing a lot of messages. I will lost the system prompt and need to scroll to the top to see what was configured. 
While the system prompt is a message in the conversation, it's also a setting that impacts the entire conversation. As a result, it would be easier if the user was able to see at any point in time in the conversation, without doing back and forth with the beginning of the conversation.

**User iterating customizing the use of the model**
As a user, I want to be able to adjust and iterate into the definition of how my model must interact. It's a multi-stage purposes.

- User will start without defining a system prompt
- User will adjust the settings and add a system prompt, replay the discussion to observe how it change the behavior of the model
- User will continue to adjust the system prompt and would like to be able to compare the behavior
It's going to be an iterative process, but here, because of this UX, the user is forced to each time get back to the creation of the playground environment.

We would need to introduce a capability to edit the system prompt and let the user decide either:
- to create a new playground environment --> so the user has a fresh session where it's possible to start from blank
- to replay the current conversation but with a different system prompt. 

That also means that there should be a way to "edit" the system prompt directly from the playground page.

feloy:
### What does this PR do?

Remove the system prompt input from playground creation page. The system prompt is still settable from the playground page.

### Screenshot / video of UI

no-system-prompt

Fix #660 

### How to test this PR?

Create a playground and check the system prompt can be set after creation time

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.spec.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.spec.ts
@@ -81,7 +81,7 @@ test('submit should throw an error if the server is stopped', async () => {
     } as unknown as InferenceServer,
   ]);
   const manager = new PlaygroundV2Manager(webviewMock, inferenceManagerMock, taskRegistryMock, telemetryMock);
-  await manager.createPlayground('playground 1', { id: 'model1' } as ModelInfo, '', 'tracking-1');
+  await manager.createPlayground('playground 1', { id: 'model1' } as ModelInfo, 'tracking-1');
 
   vi.mocked(inferenceManagerMock.getServers).mockReturnValue([
     {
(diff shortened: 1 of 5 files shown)


axel7083 (approved):
LGTM! Thanks for cleaning the api also !

PR #1043, opened 2024-05-03
resolve tag before to compare with current commit

### LINKED ISSUE #1038: check git repository when ref is a tag fails
When the `ref` of a recipe is a tag, the check of the repository _integrity_ fails, as the `ref` is compared with the commit id and not the tag name 

git-check-fails

feloy:
### What does this PR do?

When starting a recipe, resolve tag (if a tag) before to compare with current commit

Fix #1038 

### How to test this PR?

- Start a recipe, stop it, and start it again. No error should be raised.
- Change sources, then restart recipe. A message should indicate the repository is not clean

### DIFF
--- a/packages/backend/src/managers/gitManager.spec.ts
+++ b/packages/backend/src/managers/gitManager.spec.ts
@@ -291,6 +291,7 @@ describe('isRepositoryUpToDate', () => {
         url: 'repo',
       },
     ]);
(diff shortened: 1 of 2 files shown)


benoitf (approved):
checked, without the fix = warning

with the fix, no dialog and I'm able to restart the recipe

axel7083 (approved):
LGTM, works on windows!

PR #1069, opened 2024-05-08
feat: allows to set a custom folder where to download models

### LINKED ISSUE #51: Option to configure the path where models are stored
In the preferences, we should have an option to configure the path of where the models will be downloaded

lstocchi:
### What does this PR do?

It allows to set a custom folder where to download models.
Right now, the settings it is just a textbox where the user has to copy/paste the path. It will need the PR https://github.com/containers/podman-desktop/pull/7135 to be merged to switch to a file component

### Screenshot / video of UI

N/A

it resolves #51 

### How to test this PR?

1. update the models dir by using the new setting. Restart the extension and check it uses the new folder

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -11,6 +11,17 @@
   },
   "main": "./dist/extension.js",
   "contributes": {
+    "configuration": {
+      "title": "AI Lab",
+      "properties": {
+        "ai-lab.models.path": {
+          "type": "string",
+          "format": "folder",
+          "default": "",
+          "description": "Custom path where to download models. Note: The extension must be restarted for changes to take effect. (Default is blank)"
+        }
+      }
+    },
     "icons": {
       "brain-icon": {
         "description": "Brain icon",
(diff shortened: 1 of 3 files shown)


packages/backend/package.json
```diff
@@ -11,6 +11,17 @@
...
   "main": "./dist/extension.js",
   "contributes": {
+    "configuration": {
+      "title": "AI Lab",
+      "properties": {
+        "ai-lab.models.path": {
+          "type": "string",
+          "format": "folder",
+          "default": "",
+          "description": "Custom path where to download models. Note: The extension must be restarted for changes to take effect. (Default is blank)"
```
feloy, line 21: It's not very user friendly to need to restart ht eextension after the user changes the setting. The AI Lab extension could use the `onDidChangeConfiguration` to detect when the setting is changed, and update the `#modelsDir` at this moment. Or is there something else that needs to be initialized at init time, and cannot be updated at this moment?

lstocchi: No there should be no problem at init time, we would just need to dispose the old watchers i guess. My thought was to behave similarly to the podman binary path setting which is only read at activation. As i don't think this setting will be used frequently.

feloy (approved):

axel7083 (approved):
LGTM

PR #1078, opened 2024-05-10
feat: add checksum verification for downloaded models

### LINKED ISSUE #977: add checksum verification for downloaded models
it looks like when huggingface.co is not working, we download files thinking it worked well but in fact it downloaded a corrupted file

as huggingface.co is providing the checksums for the files, checksum should be added to the ai.json file in the models definition https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/backend/src/assets/ai.json#L110

and we should verify the checksum once downloaded

how to get checksum, just click on the file on huggingface.co
example on https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/blob/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf

you see
```
Git LFS Details
SHA256: 14466f9d658bf4a79f96c3f3f22759707c291cac4e62fea625e80c7d32169991
```

lstocchi:
### What does this PR do?

It display an error if the downloaded model has a different sha than the expected value

### Screenshot / video of UI

On the models page
error_sha_model

On the recipe page (it still shows the old error mesage with the double `Something went wrong` but it has been cleaned)
error_sha_recipe

it resolves #977

### How to test this PR?

1. change the sha value of a model in ai.json and download it. You should see an error

### DIFF
--- a/PACKAGING-GUIDE.md
+++ b/PACKAGING-GUIDE.md
@@ -38,7 +38,7 @@ A model has the following attributes:
 - ```license```: the license under which the model is available
 - ```url```: the URL used to download the model
 - ```memory```: the memory footprint of the model in bytes, as computed by the workflow `.github/workflows/compute-model-sizes.yaml`
-- ```sha256```: the SHA-256 checksum to be used to verify the downloaded model is identical to the original
+- ```sha256```: the SHA-256 checksum to be used to verify the downloaded model is identical to the original. It is optional and it must be HEX encoded  
 
 #### Recipes
 


lstocchi:
Rebased. Ready to be reviewed.

packages/backend/src/assets/ai.json
```diff
@@ -123,7 +123,8 @@
       "memory": 4080218931,
       "properties": {
         "chatFormat": "openchat"
-      }
+      },
+      "sha": "6adeaad8c048b35ea54562c55e454cc32c63118a32c7b8152cf706b290611487"
```
jeffmaury, line 127: Maybe rename as sha256 to better explain the algorithm used

feloy (approved):
works as expected (on mac)

jeffmaury (changes requested):
PACKAGING-GUIDE.MD should be updated

lstocchi:
@jeffmaury updated

packages/frontend/src/lib/Tooltip.svelte
```diff
@@ -58,14 +58,15 @@ export let bottom = false;
...
+export let tipClass = '';
 </script>
 
 <div class="relative inline-block">
   <span class="group tooltip-slot {$$props.class}">
     <slot name="content" />
   </span>
   <div
-    class="whitespace-nowrap absolute tooltip opacity-0 inline-block transition-opacity duration-150 ease-in-out pointer-events-none z-[10]"
+    class="whitespace-nowrap absolute tooltip opacity-0 inline-block transition-opacity duration-150 ease-in-out pointer-events-none z-[10] {tipClass}"
```
axel7083, line 69: Missing rebase ?

lstocchi: Nice catch. Removed. Thanks!!

lstocchi: I actually rebased but i forgot to remove that part

packages/backend/src/utils/downloader.ts
```diff
@@ -150,6 +160,16 @@ export class Downloader {
           return;
         }
 
+        if (checkSum) {
+          const actualSha = checkSum.digest('hex');
+          if (this.sha256 !== actualSha) {
+            callback({
```
axel7083, line 166: is the file deleted if it does not match ?

packages/frontend/src/lib/ErrorMessage.svelte
```diff
@@ -5,17 +5,24 @@ import Tooltip from './Tooltip.svelte';
 
 export let error: string;
 export let icon = false;
+export let tooltipPosition: 'top' | 'topLeft' | 'bottomLeft' = 'top';
```
axel7083, line 8: Why do you change the ErrorMessage here ?

packages/frontend/src/models/ModelInfoUI.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ModelInfo } from '@shared/src/models/IModelInfo';
+
+export interface ModelInfoUI extends ModelInfo {
+  actionError?: string;
```
axel7083, line 22: I am not sure to understand why you created an `actionError` here ? The error is not stored in the Task ?

lstocchi: Yes, it is taken from there. By using the object we can keep the logic in the Models and just pass the result to the ModelColumnActions to display it. Otherwise we have to subscribe to the tasks inside the modelcolumnactions

packages/backend/src/utils/downloader.ts
lstocchi: I think it is not atm. Maybe i should 👍

packages/frontend/src/lib/ErrorMessage.svelte
lstocchi: TBH i didn't think it was in the UI package, i'll reset it

packages/frontend/src/models/ModelInfoUI.ts
axel7083: I think we should not create a Modelinfo, as we consider the file corrupted if the sha is not matching. Therefore it should stop at the Download level, in a similar maner than a connection RESET or something failing the download task

packages/backend/src/utils/downloader.ts
axel7083: yes, I think we should consider the sha as part of the download process, meaning it is simply failed, if the sha is not matching.

packages/frontend/src/models/ModelInfoUI.ts
axel7083: I mean, I think we do not need to have this error icon on the model row. I mean, when we have an error, we could keep the error at the top.

When downloading a recipe, if we put the error in the task the user will see it, so we could just have the same for the the models page?

lstocchi: Ah ok, got it. I'll give it a try

lstocchi:
Moved the error displayed on the top as suggested by @axel7083 . TBH i preferred the error on the row as it was in line with desktop but as we rely on the top banner for showing the download in progress it makes sense to show it there

yarn.lock
```diff
@@ -4277,16 +4277,7 @@ stop-iteration-iterator@^1.0.0:
   dependencies:
```
axel7083, line 4277: Why the yarn.lock is changed ?

axel7083 (commented):
Except the yarn.lock change LGTM, great job

yarn.lock
lstocchi: It looks to me an automatic optimization.
It merged the field that were equals.

lstocchi: I guess bc i added a dependency to test the sha calculation and then i remove it so i rebuilt the yarn.lock

lstocchi: Reverted

axel7083 (approved):
Thanks for all changes, great work! 🚀

jeffmaury (approved):
LGTM. Maybe in the documentation we may indicate that the field is optional and should be HEX encoded

PR #1088, opened 2024-05-14
Do not check resources on Linux

### LINKED ISSUE #1022: No Podman machine running banner should not be displayed on Linux
We currently get the first running connection to detect if there is a podman machine running but on linux this is not mandatory.

image

feloy:

Fixes #1022 

### How to test this PR?

On Linux, Go to a Recipe page and check that no warning about podman machine is displayed


lstocchi (approved):
Codewise LGTM and thanks for fixing it!!

axel7083 (approved):
LGTM!

Issue #1134, opened 2024-05-29
Importing a models on first install does not refresh the catalog


axel7083:
### Bug description

When importing a models for the first time, the models is not shown in the page.

### Operating system

Windows 11

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

- Start podman desktop with a fresh install of AI Lab (delete `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab` or rename)
- Import a model directly

The UI does not refresh, the new imported model is not visible.

PR #1135, opened 2024-05-29
fix: json watcher parent problem

axel7083:
### What does this PR do?

The file system watcher[^1] used by Podman Desktop has a design flaw, it does not detect changes when targeting a file in a folder than does not exist.

In our scenario, on fresh install, we were activating the file system watcher on `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/user-catalog.json` however on fresh install the `redhat.ai-lab` folder does not exists.

Therefore, when we import our first model, and `redhat.ai-lab` directory is created, then `user-catalog.json` created, the file system watcher does not detect it.

To fix this, inside our JsonWatcher class, we simply need to run a `mkdir` to ensure the parent directory of our target exists. 

[^1]: https://github.com/paulmillr/chokidar

Reference https://github.com/paulmillr/chokidar/issues/1278

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1134

### How to test this PR?

- [x]: Unit tests has been provided

#### Manually

- Delete `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab` folder
- Start Podam Desktop with AI Lab
- Import model

See how it now appears in the catalog, as it is detected by the file watcher now

### DIFF
--- a/packages/backend/src/utils/JsonWatcher.ts
+++ b/packages/backend/src/utils/JsonWatcher.ts
@@ -16,7 +16,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 import { type Disposable, type FileSystemWatcher, fs, EventEmitter, type Event } from '@podman-desktop/api';
-import { promises, existsSync } from 'node:fs';
+import { promises, existsSync, mkdirSync } from 'node:fs';
+import path from 'node:path';
 
 export class JsonWatcher<T> implements Disposable {
   #fileSystemWatcher: FileSystemWatcher | undefined;
@@ -31,6 +32,10 @@ export class JsonWatcher<T> implements Disposable {
 
   init(): void {
     try {
+      // we create the parent directory of the
+      mkdirSync(path.dirname(this.path), { recursive: true });
+
+      // create file system watcher
       this.#fileSystemWatcher = fs.createFileSystemWatcher(this.path);
       // Setup listeners
       this.#fileSystemWatcher.onDidChange(this.onDidChange.bind(this));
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

feloy:
Good catch 👍 💯 It is also the reason for https://github.com/containers/podman-desktop/issues/6406, I'll make a fix there too

packages/backend/src/utils/JsonWatcher.ts
```diff
@@ -43,14 +48,17 @@ export class JsonWatcher<T> implements Disposable {
   }
 
   private onDidCreate(): void {
+    console.log('JsonWatcher onDidCreate', this.path);
```
feloy, line 51: to remove?

feloy (approved):

packages/backend/src/utils/JsonWatcher.ts
```diff
@@ -31,6 +32,10 @@ export class JsonWatcher<T> implements Disposable {
 
   init(): void {
     try {
+      // we create the parent directory of the
```
axel7083, line 35:
```suggestion
      // we create the parent directory of the watched content
      // if the parent directory does not exists, the watcher is not initialized properly 
```

packages/backend/src/utils/JsonWatcher.ts
axel7083, line 51:
```suggestion
```

packages/backend/src/utils/JsonWatcher.ts
```diff
@@ -43,14 +48,17 @@ export class JsonWatcher<T> implements Disposable {
   }
 
   private onDidCreate(): void {
+    console.log('JsonWatcher onDidCreate', this.path);
     this.requestUpdate();
   }
 
   private onDidDelete(): void {
+    console.log('JsonWatcher onDidDelete', this.path);
```
axel7083, line 56:
```suggestion
```

packages/backend/src/utils/JsonWatcher.ts
axel7083: Yes! Done thx

benoitf (commented):
I would still enhance the podman desktop watcher

can be merged but need to enhance the behavior of podman desktop core

if watchers are applied on unexisting folders, podman desktop should keep the watchers until the parent folders are added

PR #1141, opened 2024-05-31
feat(win): better gpu detection

axel7083:
### What does this PR do?

Remove the old GPUManager code detecting the Nvidia GPU using `nvidia-smi` and replace it with a generic worker system.

This PR is adding the `WinGPUDetector` worker.  This worker is responsible of detecting the GPUs available on windows.

### How

We need two information, the name and the VRAM available for each GPU, sadly windows is crap[^1][^2][^3], and they don't fix their code[^1], so we have to read the registry[^4] to access the exact value of the VRAM for each GPU detected.

> Reading the registry could feel scary but we are simply following the official Microsoft documentation here, with long term support values. Therefore not too scary :)

Other solution do not provide accurate value[^3].

Related to https://github.com/containers/podman-desktop-internal/issues/270

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1144

### How to test this PR?

- [x] Unit tests has been added

#### Manually

Adding the following in the studio.ts file (activate function)
```js
const gpuManager = new GPUManager(this.#panel?.webview);
gpuManager.collectGPUs().then((gpus) => {
  console.log('Collected GPUs', gpus);
}).catch((err: unknown) => {
  console.error('Something went wrong while tryng to collect GPUs', err);
});
```

Example output 

```
[ai-lab] Collected GPUs [ { model: 'NVIDIA GeForce GTX 1060 6GB', vram: 6442450944 } ]
```

[^1]: Msinfo32.exe reports an unexpected value for the Display Adapter RAM when the graphics adapter has 2 GB or more of dedicated video memory https://learn.microsoft.com/en-us/troubleshoot/windows-client/setup-upgrade-and-drivers/msinfo32-report-wrong-display-adapter-ram
[^2]: Windows is crap https://superuser.com/questions/1461858/fetch-correct-vram-for-gpu-via-command-line-on-windows
[^3]: Windows is crap https://stackoverflow.com/questions/68274009/wmi-win32-videocontroller-ram-4gb-limit
[^4]: Windows registry values https://learn.microsoft.com/fr-fr/windows-hardware/drivers/install/system-defined-device-setup-classes-available-to-vendors

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -55,13 +55,13 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
-    "fast-xml-parser": "^4.4.0",
     "isomorphic-git": "^1.25.10",
     "mustache": "^4.2.0",
     "openai": "^4.47.2",
     "postman-code-generators": "^1.10.1",
     "postman-collection": "^4.4.0",
     "semver": "^7.6.2",
+    "winreg": "^1.2.5",
     "xml-js": "^1.6.11"
   },
   "devDependencies": {
@@ -69,6 +69,7 @@
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^20",
     "@types/postman-collection": "^3.5.10",
+    "@types/winreg": "^1.2.36",
     "vitest": "^1.6.0",
     "@types/mustache": "^4.2.5"
   }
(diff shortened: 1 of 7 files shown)


lstocchi (approved):
Tested and my output is `model: "Intel(R) UHD Graphics\u0000" vram: undefined` which is in line with what i have since my gpu does not have dedicated ram.

LGTM

Issue #1112, opened 2024-05-22
InferenceServer should support multiple backend 


axel7083:
### Is your enhancement related to a problem? Please describe

In today implementation, the inference server will always use the playground container image, this is not suitable for whisper models for example.

### Describe the solution you'd like

After https://github.com/containers/podman-desktop-extension-ai-lab/issues/1111 we should make the InferenceServer use the backend field to select the right container to run the models.

PR #1161, opened 2024-06-05
feat(inference): introducing InferenceProviders

axel7083:
### What does this PR do?

Introducing `InferenceProvider` interface, which is an abstraction class made to create InferenceServers. In todays implementation we do not make distinctions between `backends` (llamacpp, whispercpp etc.)

This is the first step in abstracting the inference providers, to ease the customization of inference server with new provider in the future (whispercpp, llamacpp-cuda, ollama etc.)

### Documentation

image

[](https://mermaid.live/edit#pako:eNp1kcFugzAMhl_Fyrl9AQ69bNPEAQnRKxcvMSwaOMxxkKqq774gSqtuXU5O_P3-f8VnY4MjU5hI34nY0qvHXnBsGfJBq0EgRZL1PqGot35CVigrwAgldySLrkLG_ilXNw8g1BJm70igod5HlVPLq2rx2e8Ph7Iqci_HifoihEo37ZFk3jzKamPfOCYhmHHwDmzgzvdJUH3gB7JuCngnvSfZgsQrVjd3d03C_5M362UeMX4M5P7Sv2GLwwATSRdkhMAQaSCrz5Wr1uzMSDKid3lD5-WtNfpJI7WmyKVD-WpNy5fMYdJwPLE1hUqinUmTyz933aYpOhwiXX4AV1unEw)

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1112

### How to test this PR?

#### Manually (recommended)

- Create an inference server
- Start chatbot recipe

### DIFF
--- a/packages/backend/src/managers/inference/inferenceManager.spec.ts
+++ b/packages/backend/src/managers/inference/inferenceManager.spec.ts
@@ -223,7 +223,7 @@ describe('Create Inference Server', () => {
         modelsInfo: [],
         port: 8888,
       }),
-    ).rejects.toThrowError('o enabled provider could be found.');
+    ).rejects.toThrowError('no enabled provider could be found.');
   });
 
   test('inference provider provided should use get from InferenceProviderRegistry', async () => {
--- a/packages/backend/src/managers/inference/inferenceManager.spec.ts
+++ b/packages/backend/src/managers/inference/inferenceManager.spec.ts
@@ -17,42 +17,35 @@
  ***********************************************************************/
 import {
   containerEngine,
-  provider,
   type Webview,
   type TelemetryLogger,
-  type ImageInfo,
   type ContainerInfo,
   type ContainerInspectInfo,
-  type ProviderContainerConnection,
 } from '@podman-desktop/api';
 import type { ContainerRegistry } from '../../registries/ContainerRegistry';
 import type { PodmanConnection } from '../podmanConnection';
 import { beforeEach, expect, describe, test, vi } from 'vitest';
 import { InferenceManager } from './inferenceManager';
 import type { ModelsManager } from '../modelsManager';
-import { LABEL_INFERENCE_SERVER, INFERENCE_SERVER_IMAGE } from '../../utils/inferenceUtils';
+import { LABEL_INFERENCE_SERVER } from '../../utils/inferenceUtils';
 import type { InferenceServerConfig } from '@shared/src/models/InferenceServerConfig';
 import type { TaskRegistry } from '../../registries/TaskRegistry';
 import { Messages } from '@shared/Messages';
+import type { InferenceProviderRegistry } from '../../registries/InferenceProviderRegistry';
+import type { InferenceProvider } from '../../workers/provider/InferenceProvider';
 
 vi.mock('@podman-desktop/api', async () => {
   return {
     containerEngine: {
       startContainer: vi.fn(),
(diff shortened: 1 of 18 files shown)


feloy:
In a near future, we will want to run inference servers in Kubernetes clusters too (OpenShift AI typically). I cannot see anything blocking doing this with this architecture, but just to be sure you have this scenario in mind

axel7083:
> In a near future, we will want to run inference servers in Kubernetes clusters too (OpenShift AI typically). I cannot see anything blocking doing this with this architecture, but just to be sure you have this scenario in mind

Thanks @feloy for this feedback. I was not having this in mind, but I think it would make it easier than our current architecture, as we could create a `KubernetesInferenceProvider` responsible of creating the pod in the Kubernetes cluster.

lstocchi (approved):
Codewise LGTM. Also tested and works fine. Nice job!!

packages/backend/src/managers/inference/inferenceManager.spec.ts
```diff
@@ -233,119 +212,59 @@ describe('init Inference Manager', () => {
...
-    ).rejects.toThrowError('image unknown not found.');
-
-    expect(containerEngine.listImages).toHaveBeenCalled();
+      inferenceManager.createInferenceServer({
+        inferenceProvider: undefined,
+        labels: {},
+        modelsInfo: [],
+        port: 8888,
+      }),
+    ).rejects.toThrowError('o enabled provider could be found.');
```
jeffmaury, line 226: The error message looks strange to me

jeffmaury (commented):
Dependency on Podman Desktop 1.11 does not seems an absolute requirement and could be relaxed with few changes so I would delay merge as @slemeur approval should be required

packages/backend/src/managers/inference/inferenceManager.spec.ts
axel7083: Fixed

axel7083:
> Dependency on Podman Desktop 1.11 does not seems an absolute requirement and could be relaxed with few changes so I would delay merge as @slemeur approval should be required

I revert to `@podman-desktop/api` 1.10.3. As mention there is not absolute requirement.

packages/backend/src/workers/provider/InferenceProvider.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export type BetterContainerCreateResult = ContainerCreateResult & { engineId: string };
+
+export abstract class InferenceProvider
+  implements IWorker<InferenceServerConfig, BetterContainerCreateResult>, Disposable
+{
+  protected constructor(private taskRegistry: TaskRegistry) {}
+
+  abstract name: string;
+  abstract enabled(): boolean;
```
jeffmaury, line 39: Not sure what is the purpose of this ? May need to be async for future Kubernetes provider

jeffmaury (approved):

packages/backend/src/workers/provider/InferenceProvider.ts
axel7083: Because we are extending the IWorker interface, which aims to provide a quick way of determining if a worker is enabled or not.

In the case of a kubernetes connection, the worker is not responsible of maintaining the connection with the cluster, we would have to have a component/registry keeping track of this information, and making it available in a sync maner.

Issue #1166, opened 2024-06-06
bug(LocalRepositoryRegistry): local repositories are not registered when the extension start


axel7083:
### Bug description

LocalRepository is only updated when recipes are started, this is an issue, as we cannot keep track of which one is already cloned, before running the recipe.

### Operating system

Windows 11

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

- Start a recipe
- Restart AI Lab extension
- Go to recipe, you do not see it available locally.

axel7083:
This is required to achieve design in https://github.com/containers/podman-desktop-extension-ai-lab/issues/1126

Issue #822, opened 2024-04-05
API/CLI for ai-lab


slemeur:
It would be nice to have api to interact with ai-lab, so other tools can connect to ai-lab and pull, start models from APIs.
Having a CLI wrapper around those apis would also be nice.

slemeur:
Example use case: 
- Enable AI Lab to be a provider for continue (https://continue.dev/docs/model-setup/select-provider)
- Quarkus Integration

slemeur:
Initial step should be to provide compatibility to enable compatibility with the following APIs: https://github.com/ollama/ollama/blob/main/docs/api.md

PR #1175, opened 2024-06-06
chore: improve resource disposal

axel7083:
### What does this PR do?

Improve the `studio.ts` file, to improve the way we dispose resources (managers).

Moreover, making more classes implements the disposable, to improve consistency in our designs 

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1166

### How to test this PR?

### DIFF
--- a/packages/backend/src/managers/applicationManager.spec.ts
+++ b/packages/backend/src/managers/applicationManager.spec.ts
@@ -1001,7 +1001,7 @@ describe('pod detection', async () => {
     );
   });
 
-  test('adoptRunningApplications updates the app state with the found pod', async () => {
+  test('init updates the app state with the found pod', async () => {
     vi.mocked(podManager.getPodsWithLabels).mockResolvedValue([
       {
         Labels: {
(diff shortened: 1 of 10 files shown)


jeffmaury (commented):
The way is is implemented is incompatible with the future REST server (see 
#822)
Why lazy initialization is not handled by the component itself when a request is processed (ie modelManager when getModels and so on is called)

axel7083:
> Why lazy initialization is not handled by the component itself when a request is processed (ie modelManager when getModels and so on is called)

It can be, it would not be hard to change in the future, this PR is also cleanup, and documentation. We can make the change later for https://github.com/containers/podman-desktop-extension-ai-lab/issues/822

jeffmaury:
> > Why lazy initialization is not handled by the component itself when a request is processed (ie modelManager when getModels and so on is called)
> 
> It can be, it would not be hard to change in the future, this PR is also cleanup, and documentation. We can make the change later for #822
> 
> Moreover, as for now, we avoid duplicating a lot of code inside each component, because if we want each of them to be initialized when used, with the _dynamic initialization_ you are talking about @jeffmaury they would all need to add some initialization check before each call (E.g. `if(!isInitiliazied()) init()`) ?

As we have the use case in the near future, I would prefer refactoring this PR now rather than later

axel7083:
> As we have the use case in the near future, I would prefer refactoring this PR now rather than later

Alright, I updated the title of the Pull request to reflect what it does

jeffmaury (approved):
LGTM

Issue #1111, opened 2024-05-22
epic(Recipes): change relation between models and recipes


axel7083:
### Epic domain

Today, when importing a new models, it cannot be used in recipes. This issue come from the fact that we are hardcoding the compatible models in the recipe definition.

## Introducing `backend` field

- [x] #1181
- [x] #1192
- [x] #1195

## Recipes should have a `recommended` field

- [ ] Recipes could recommend the best suitable models (plural)
- [ ] Recipe details page highlight the recommended model(s)

## Catalog migration

- [ ] Catalog should be versioned 
- [ ] We should handle catalog migration

axel7083:
Closing as completed

PR #1180, opened 2024-06-07
docs: discussion recipe models relation

axel7083:
See docs/proposals/recipe-models-relation.md

Related to https://github.com/containers/podman-desktop-extension-ai-lab/issues/1111

### DIFF
--- a/docs/proposals/recipe-models-relation.md
+++ b/docs/proposals/recipe-models-relation.md
@@ -64,3 +64,7 @@ Should we add a property to providers to group them ? (tags, labels, implementat
 ### Personal
 
 I am having difficulty choosing the right wording to define them.
+
+An easy solution might be to add a `type: string` property to InferenceProvider.
+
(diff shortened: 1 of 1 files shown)


jeffmaury:
My 2 cents:

- the cuda and other variants shoud be handled by the inferenceprovider based on the environment
- inference providers should be registered according to the backend they support
- I don't see the use case for llama.cpp and ollama as they are the same
- I'm +1 for backend property

axel7083:
> My 2 cents:
> 
> * the cuda and other variants shoud be handled by the inferenceprovider based on the environment

Okey with this proposition

> * I don't see the use case for llama.cpp and ollama as they are the same

When I say llama.cpp I am thinking about our own image build by the recipe team. Because they probably do not need the same arguments, environment variables etc. They should be consider as two separate provider.

> * I'm +1 for backend property

Good, then https://github.com/containers/podman-desktop-extension-ai-lab/pull/1186

axel7083:
Closing as resolved

PR #1196, opened 2024-06-11
feature: revamp model import page

axel7083:
### What does this PR do?

To support the new `backend` property, we need to revamp the Model import page. Originally @lstocchi implement it to support importing multiple models at the same time, however with the growing number of properties, we discussed and thought it would be simpler to have a dedicated form page, to import a model. 

This PR revamp the model import page to add allow the user to define the backend used.

### Screenshot / video of UI

image

image

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/68ac2f48-e425-466f-92ea-02c946d3216f

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1195

### How to test this PR?

#### Manually

- Go to models page, click on `Import`
- Drag and drop or open the file explorer
- Change the name or not
- Select the backend

Open the `user-catalog.json` and see the `backend` property properly set.


lstocchi (commented):
I really like the new UI, great work!!

I noticed two things while playing with it.
1) if you click on the drag and drop field, it opens up 2 dialogs
2) If you drop a model and leave everything as it is then you're not able to start the inference server. I think that we show llamacpp as the default backend but it is not actually populated. So i need to change item into the dropdownlist and then go back to llamacpp to make it actually pick it

jeffmaury (approved):
LGTM. Only issue is the double dialog being displayed

axel7083:
File explorer opening two time will be fixed by https://github.com/containers/podman-desktop-extension-ai-lab/pull/1207

Issue #1126, opened 2024-05-28
Recipes details page is inconsistent and is hard to use


axel7083:
### Is your enhancement related to a problem? Please describe

The details page depends on the model selected, leading to problems in the UI when switching models, this is a weird behaviour, leading to potential problem when multiple models could be selected.

### Describe the solution you'd like

Improve the way we show running recipes

### Describe alternatives you've considered

Creating a recipe should be done in a similar way than creating a Playground and InferenceServer.

### Additional context

Model selected with matching pod
![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/0866d5db-8f66-4894-8c2f-d2241c07d1da)

Model selected without matching pod
![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/c0d933e6-c539-442e-8c46-6164e56edfdb)

ekidneyrh:
We discussed this during the UX call yesterday. 

Problem is, when looking to run an AI app, the model has to be selected from the long unsorted list first. This can make it hard to find the right model. It also follows a pattern that's not seen anywhere else in pd.

We could have it similar to how a model is created. Going through Model Services > New Model Service and then a form page. 

![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/87311656/d093428c-4f15-435b-91e2-065282278107)

Recipes catalog would benefit from a revamp. The categories and how it's displayed hasn't been touched in a few months and would benefit from another iteration. Maybe we could have it similar to the catalog under models with the tabs at the top for sorting. 

![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/87311656/f540f356-77f3-4857-bdce-dc468ee6c214)

Do we need this page that divides up the different recipes? Was its original purpose that there are so many different recipes, they needed to be sorted like this? Or is it more so to have the page that describes what each category is when you click into a card? 

![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/87311656/88231a6c-6e13-4915-b442-45863f9fe7fc)

![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/87311656/d4aedb35-da62-467b-9993-203a647fbcbb)

axel7083:
> Do we need this page that divides up the different recipes? Was its original purpose that there are so many different recipes, they needed to be sorted like this? Or is it more so to have the page that describes what each category is when you click into a card?

[...]

PR #1199, opened 2024-06-11
feat: revamp recipes list page

axel7083:
### What does this PR do?

Following one of the mockup in https://github.com/containers/podman-desktop-extension-ai-lab/issues/1126, improving the recipes card.

#### Notable change

- Adding an icon showing the status (cloned or not) similar to the one in extension catalog
- Adding the path in the card 

### Screenshot / video of UI

image

image

image

Fixes partially https://github.com/containers/podman-desktop-extension-ai-lab/issues/1126

### How to test this PR?

### DIFF
--- a/packages/backend/src/managers/applicationManager.ts
+++ b/packages/backend/src/managers/applicationManager.ts
@@ -129,6 +129,29 @@ export class ApplicationManager extends Publisher<ApplicationState[]> implements
     }
   }
 
+  public async cloneApplication(recipe: Recipe, labels?: { [key: string]: string }): Promise<void> {
+    const localFolder = path.join(this.appUserDirectory, recipe.id);
+
+    // clone the recipe repository on the local folder
+    const gitCloneInfo: GitCloneInfo = {
+      repository: recipe.repository,
+      ref: recipe.ref,
+      targetDirectory: localFolder,
+    };
+    await this.doCheckout(gitCloneInfo, {
+      ...labels,
+      'recipe-id': recipe.id,
+    });
+
+    this.localRepositories.register({
+      path: gitCloneInfo.targetDirectory,
+      sourcePath: path.join(gitCloneInfo.targetDirectory, recipe.basedir ?? ''),
+      labels: {
+        'recipe-id': recipe.id,
+      },
+    });
+  }
+
   /**
    * This method will execute the following tasks
    * - git clone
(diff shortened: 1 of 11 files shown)


benoitf (approved):
for the path inside the card I don't know if we should truncate the middle of the path so we see the beginning and the end of the path but it can also be a bad idea !

axel7083:
> for the path inside the card I don't know if we should truncate the middle of the path so we see the beginning and the end of the path but it can also be a bad idea !

We will probably iterate a bit on this with Emma when she will be back ! I keep your remark in mind, I also feel a bit weird with the path.

jeffmaury:
> for the path inside the card I don't know if we should truncate the middle of the path so we see the beginning and the end of the path but it can also be a bad idea !

Yes can't we have the full path as tooltip ?

jeffmaury (approved):
LGTM comment wise

feloy:
I find the recipe status button/indicator disturbing. It's ok when the recipe is downloadable, but when it is downloaded, the button is still active and does nothing. Also, a tooltip would help understand the meaning

axel7083:
⚠️ NOT INCLUDED IN THIS PR

I had a little fun with finding a nice way of allowing the user to see the full path, and I though: what if the text automatically scroll when hovering

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/99d1eb8a-459e-4f64-b122-45385a131284

if this idea please some people, I could make another PR to include it, what do you think @jeffmaury @benoitf @deboer-tim

jeffmaury:
> ⚠️ NOT INCLUDED IN THIS PR
> 
> I had a little fun with finding a nice way of allowing the user to see the full path, and I though: what if the text automatically scroll when hovering
> text-hover-scroll.mp4
> 
> if this idea please some people, I could make another PR to include it, what do you think @jeffmaury @benoitf @deboer-tim
> 
> > I can control the speed of the animation, I am also thinking we could add `cursor-copy` and make it copy when the user click on it

Seems to me that it is more common to have a UI glitch (tooltip) that shows the full path when a specific UI action is taken (hover,...). Here the con that I see is that the full path is not displayed entirely but that's still an improvement

feloy:
> > ⚠️ NOT INCLUDED IN THIS PR
> > I had a little fun with finding a nice way of allowing the user to see the full path, and I though: what if the text automatically scroll when hovering
> > text-hover-scroll.mp4
> > if this idea please some people, I could make another PR to include it, what do you think @jeffmaury @benoitf @deboer-tim
> > > I can control the speed of the animation, I am also thinking we could add `cursor-copy` and make it copy when the user click on it
> 
> Seems to me that it is more common to have a UI glitch (tooltip) that shows the full path when a specific UI action is taken (hover,...). Here the con that I see is that the full path is not displayed entirely but that's still an improvement

Also, one important thing for the user I think is to be able to copy the path string. If the path is not displayed completely, the copy is not effective. The tooltip won't help either, as its content cannot be copied. Should we add a copy button?

axel7083:
Merging for now this PR and I will open an issue for follow up

PR #1210, opened 2024-06-12
feat: make recipes uses models based on backend property

axel7083:
### What does this PR do?

Make the list of models displayed in the recipe details page based on the backend.

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/3fce9fad-fb34-4005-9759-74bf8755c0da

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1192

### How to test this PR?

- [x] unit tests has been adapted


packages/frontend/src/lib/RecipeDetails.svelte
```diff
@@ -135,7 +135,7 @@ const deleteLocalClone = () => {
       <div class="bg-charcoal-900 min-w-[200px] grow flex flex-col p-2 rounded-md space-y-3">
         <div class="flex justify-between items-center">
           <span class="text-sm" aria-label="model-selected">{model?.name}</span>
-          {#if recipe?.models?.[0] === model.id}
+          {#if recipe?.recommended?.[0] === model.id}
```
jeffmaury, line 138: There maybe several recommended models now

packages/frontend/src/pages/Recipe.svelte
```diff
@@ -15,21 +15,31 @@ import ContainerConnectionStatusInfo from '../lib/notification/ContainerConnecti
...
 export let recipeId: string;
 
 // The recipe model provided
 $: recipe = $catalog.recipes.find(r => r.id === recipeId);
 $: categories = $catalog.categories;
+
+// model selected to start the recipe
 let selectedModelId: string;
-$: selectedModelId = recipe?.models?.[0] ?? '';
+$: selectedModelId = recipe?.recommended && recipe.recommended.length > 0 ? recipe?.recommended?.[0] : '';
```
jeffmaury, line 29: Can't be shortened ?

packages/frontend/src/pages/Recipe.svelte
```diff
@@ -15,21 +15,31 @@ import ContainerConnectionStatusInfo from '../lib/notification/ContainerConnecti
...
 let connectionInfo: ContainerConnectionInfo | undefined;
 $: if ($modelsInfo && selectedModelId) {
   checkContainerConnectionStatus($modelsInfo, selectedModelId, 'recipe')
     .then(value => (connectionInfo = value))
     .catch((e: unknown) => console.log(String(e)));
 }
 
+let models: ModelInfo[];
+$: models = $catalog.models.filter(
+  model => (model.backend ?? InferenceType.NONE) === (recipe?.backend ?? InferenceType.NONE),
```
jeffmaury, line 40: Wondering if we should not also consider recommanded as a way to associate a model with a recipe (old style) eg recipe without backend

jeffmaury (changes requested):

packages/frontend/src/pages/Recipe.svelte
axel7083: Oh yeah, that's very smart ! For the object detection that would works

packages/frontend/src/pages/Recipe.svelte
axel7083: Not really, because the previous code had a flaw, it was not taking the possibility than there is not models. Let's say we create a recipe without recommended, then we have an error, this line ensure no error will be raised.

packages/frontend/src/pages/Recipe.svelte
axel7083: I would delay this in another PR, probably https://github.com/containers/podman-desktop-extension-ai-lab/issues/1126 as this would probably need some UI/UX because we would still want user to be able to do their own choice I guess

axel7083:
@jeffmaury I adapted the code to handle multiple recommended models. Here is a demo of what it would look like

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/351297a7-0e75-4deb-8fa2-e5bc69f3745f

packages/frontend/src/models/RecipeModelInfo.ts
```diff
@@ -18,6 +18,7 @@
 
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 
+// todo :remove
```
jeffmaury, line 21: Can't we remove it now ?

jeffmaury (approved):
LGTM

packages/frontend/src/models/RecipeModelInfo.ts
axel7083: Sadly no, there are some weird dependency on it yet that I can't remove

https://github.com/containers/podman-desktop-extension-ai-lab/blob/df409dd5a854bfc288768c0e37459a2da157ceb5/packages/frontend/src/lib/table/model/ModelColumnRecipeSelection.svelte#L6

Moreover there are some work being done in the Table component, which will require some work to be done to fix it. I would prefer keep the todo remove as this need to get addressed when the Table component will be updated

PR #1227, opened 2024-06-17
feat: dedicated page for starting recipe

axel7083:
## What does this PR do?

For the recipe details revamp, this PR improve the consistency of the application, by making a dedicated page for starting a recipe. This new page is similar to the `Service` and `Playground` creation page.

## Notable change

- Removing the `RecipeModels` page
- Removing the applications statuses from the RecipeDetails page
> Recipe are like container image, they need to be appart 
- Removing the double models table (which were inconsistent)
- Adding `svelte-select` library, as it is wonderful to work with customizable, nice looking select.

## Future work

- We need to refine the wording: `pulling application`, `starting application`, application, recipe etc. This is very difficult to make some clear statement of what is what.
- Prevent/alert user when trying to start multiple time the same recipe
- Using the podman-desktop api, we need to be able to come back to the start recipe page

## Screenshot / video of UI

**Full workflow**

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/3d5bb168-ef7e-46b5-93aa-87d3fa80ce25

**Error handling**

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/a5ab9de3-c2a2-49ab-b8a2-269ba9f3fdb8

## What issues does this PR fix or reference?

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1126

## How to test this PR?

- [x] Unit tests has been provided ✅

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,10 +59,6 @@
       "tests/*"
     ]
   },
-  "resolutions": {
-    "string-width": "^4.2.0",
-    "wrap-ansi": "^7.0.0"
-  },
   "dependencies": {
     "js-yaml": "^4.1.0"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -241,6 +241,26 @@
   resolved "https://registry.yarnpkg.com/@faker-js/faker/-/faker-5.5.3.tgz#18e3af6b8eae7984072bbeb0c0858474d7c4cefe"
   integrity sha512-R11tGE6yIFwqpaIqcfkcg7AICXzFg14+5h5v0TfF/9+RMDL6jhzCy/pxHVOfbALGdtVYdt6JdR21tuxEgl34dw==
 
+"@floating-ui/core@^1.5.0", "@floating-ui/core@^1.6.0":
+  version "1.6.4"
+  resolved "https://registry.yarnpkg.com/@floating-ui/core/-/core-1.6.4.tgz#0140cf5091c8dee602bff9da5ab330840ff91df6"
+  integrity sha512-a4IowK4QkXl4SCWTGUR0INAfEOX3wtsYw3rKK5InQEHMGObkR8Xk44qYQD9P4r6HHw0iIfK6GUKECmY8sTkqRA==
+  dependencies:
+    "@floating-ui/utils" "^0.2.4"
+
+"@floating-ui/dom@^1.5.3":
+  version "1.6.7"
+  resolved "https://registry.yarnpkg.com/@floating-ui/dom/-/dom-1.6.7.tgz#85d22f731fcc5b209db504478fb1df5116a83015"
+  integrity sha512-wmVfPG5o2xnKDU4jx/m4w5qva9FWHcnZ8BvzEe90D/RpwsJaTAVYPEPdQ8sbr/N8zZTAHlZUTQdqg8ZUbzHmng==
+  dependencies:
+    "@floating-ui/core" "^1.6.0"
+    "@floating-ui/utils" "^0.2.4"
+
+"@floating-ui/utils@^0.2.4":
+  version "0.2.4"
+  resolved "https://registry.yarnpkg.com/@floating-ui/utils/-/utils-0.2.4.tgz#1d459cee5031893a08a0e064c406ad2130cced7c"
+  integrity sha512-dWO2pw8hhi+WrXq1YJy2yCuWoL20PddgGaqTgVe4cOS9Q6qklXCiA1tJEqX6BEwRNSCP84/afac9hd4MS+zEUA==
+
 "@fortawesome/fontawesome-common-types@6.5.2":
   version "6.5.2"
   resolved "https://registry.yarnpkg.com/@fortawesome/fontawesome-common-types/-/fontawesome-common-types-6.5.2.tgz#eaf2f5699f73cef198454ebc0c414e3688898179"
@@ -1111,6 +1131,11 @@ ansi-styles@^5.0.0:
   resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-5.2.0.tgz#07449690ad45777d1924ac2abb2fc8895dba836b"
   integrity sha512-Cxwpt2SfTzTtXcfOlzGEee8O+c+MmUgGrNiBcXnuWxuFJHe6a5Hz7qwhwe5OgaSYI0IJvkLqWX1ASG+cJOkEiA==
 
+ansi-styles@^6.1.0:
+  version "6.2.1"
+  resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-6.2.1.tgz#0e62320cf99c21afff3b3012192546aacbfb05c5"
+  integrity sha512-bN798gFfQX+viw3R7yrGWRqnrN2oRkEkUjjl4JNn4E8GxxbjtG3FbrEIIY3l8/hrwUwIeCZvi4QuOTP4MErVug==
+
 any-promise@^1.0.0:
   version "1.3.0"
   resolved "https://registry.yarnpkg.com/any-promise/-/any-promise-1.3.0.tgz#abc6afeedcea52e809cdc0376aed3ce39635d17f"
@@ -1717,6 +1742,11 @@ dom-accessibility-api@^0.6.3:
   resolved "https://registry.yarnpkg.com/dom-accessibility-api/-/dom-accessibility-api-0.6.3.tgz#993e925cc1d73f2c662e7d75dd5a5445259a8fd8"
   integrity sha512-7ZgogeTnjuHbo+ct10G9Ffp0mif17idi0IyWNVA/wcwcm7NPOD/WEHVP3n7n3MhXqxoIYm8d6MuZohYWIZ4T3w==
 
+eastasianwidth@^0.2.0:
+  version "0.2.0"
+  resolved "https://registry.yarnpkg.com/eastasianwidth/-/eastasianwidth-0.2.0.tgz#696ce2ec0aa0e6ea93a397ffcf24aa7840c827cb"
+  integrity sha512-I88TYZWc9XiYHRQ4/3c5rjjfgkjhLyW2luGIheGERbNQ6OY7yTybanSpDXZa8y7VUP9YmDcYa+eyq4ca7iLqWA==
+
 electron-to-chromium@^1.4.820:
   version "1.4.828"
   resolved "https://registry.yarnpkg.com/electron-to-chromium/-/electron-to-chromium-1.4.828.tgz#a1ee8cd8847448b2898d3f2d9db02113f9c5b35c"
@@ -1735,6 +1765,11 @@ emoji-regex@^8.0.0:
   version "8.0.0"
   resolved "https://registry.yarnpkg.com/emoji-regex/-/emoji-regex-8.0.0.tgz#e818fd69ce5ccfcb404594f842963bf53164cc37"
   integrity sha512-MSjYzcWNOA0ewAHpz0MxpYFvwg6yjy1NG3xteoqz644VCo/RPgnr1/GGt+ic3iJTzQ8Eu3TdM14SawnVUmGE6A==
+
+emoji-regex@^9.2.2:
+  version "9.2.2"
+  resolved "https://registry.yarnpkg.com/emoji-regex/-/emoji-regex-9.2.2.tgz#840c8803b0d8047f4ff0cf963176b32d4ef3ed72"
+  integrity sha512-L18DaJsXSUk2+42pv8mLs5jJT2hqFkFE4j21wOmgbUqsZ2hL72NsUU785g9RXgo3s0ZNgVl42TiHp3ZtOv/Vyg==
 
 end-of-stream@^1.1.0:
   version "1.4.4"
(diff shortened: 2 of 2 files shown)


slemeur:
Nice work on this, great improvement.

A couple of feedback:
- The first inputbox is not necessarily clear to the end user. While for the choice of the model, we have a tooltip that explains what to do. It's not clear that I'll be doing a clone and I can choose where to make this clone. Having a tooltip or a description label could nicely improve the experience and better explain the experience.

- The other point is more about the way we show the status; I acknowledge we are doing it this way in other screens of podman desktop - but with so many items appearing during the process, it looks strange that they are added at the top of the screen, before the form, which cause the screen to show a lot of "moving effects".

axel7083:
> * The first inputbox is not necessarily clear to the end user.

It is not an inputbox, it is a card showing the selected recipe. This page is not design to navigate from the recipes list, but from a recipe page. it is here to inform the user which recipe he is going to start.

> While for the choice of the model, we have a tooltip that explains what to do. It's not clear that I'll be doing a clone and I can choose where to make this clone. Having a tooltip or a description label could nicely improve the experience and better explain the experience.

There are two things here:  the cloning aspect (1) and the directory used (2).
- (1) I can add similar information than for the model under the recipe name, to indicate it will be clone, I can also show from where (git repo). 
- (2) We do not support setting custom directory for now.

But all of this are improvements, I would rather have dedicated issues later to work on, as this is already a large change. 

> * The other point is more about the way we show the status; I acknowledge we are doing it this way in other screens of podman desktop - but with so many items appearing during the process, it looks strange that they are added at the top of the screen, before the form, which cause the screen to show a lot of "moving effects".

I would rather keep focus on functionality in this PR, and iterate later on potential improvement. I have a `Future work` on this PR, we can have a meeting later so discuss those issues more in details. But for now I think it is better to stick with the current design used in other pages.

lstocchi (commented):
If i start a recipe and then i leave the page there is no way for me to know which is the status of the starting process or i'm missing something?

axel7083:
> * Using the podman-desktop api, we need to be able to come back to the start recipe page

@lstocchi as it is written in `Future work` section _Using the podman-desktop api, we need to be able to come back to the start recipe page_ By this I mean, it would be possible to use the task manager of podman desktop to add a link to this page.

lstocchi:
> > * Using the podman-desktop api, we need to be able to come back to the start recipe page
> 
> @lstocchi as it is written in `Future work` section _Using the podman-desktop api, we need to be able to come back to the start recipe page_ By this I mean, it would be possible to use the task manager of podman desktop to add a link to this page.
> 
> So for now, yes it is not possible if you leave the page, you we can add that support in a future PR

For me it's a step back for what we have now and i would prefer to have that as well.

feloy:
When the user restarts an app (IMO it is a very common operation, as the user will modify the sources to adapt the sample app, and will want to rebuild/restart the app), this should be done from the "Running AI Apps" page, but the steps (and the potential errors) are not visible anymore.

axel7083:
> When the user restarts an app (IMO it is a very common operation, as the user will modify the sources to adapt the sample app, and will want to rebuild/restart the app), this should be done from the "Running AI Apps" page, but the steps (and the potential errors) are not visible anymore.

We can probably in another PR redirect to the StartRecipe page when we want to rebuild the recipe

axel7083:
@lstocchi I was hopping I could set a goto link in the task manager of podman desktop, however it is not yet supported. The best I can do for now is to save the trackingId in the router queries, therefore allowing to restore the page status when leaving the webview.

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/8734a2bd-c785-4403-ade1-d468a714d92e

I opened https://github.com/containers/podman-desktop/issues/7709 to enhance Tasks to let extension define custom actions.

axel7083:
@lstocchi @feloy PR rebased ✅

lstocchi (commented):
Tested again. My concern still remain. If i go back to the ai-lab after changing page i do not see the building status.

no_status

I would say to wait for @jeffmaury, @benoitf or @slemeur . If they're ok to merge it, np for me. The process to start a recipe works as before, so no complains.

lstocchi:
Tried again and i think i didn't built with the last commit last time bc now it works fine

lstocchi (approved):
Still some concern about the missing way to retrieve the status when navigating in ai-lab but i appreciate the change to retrieve it when moving out/in ai-lab. Thank you!

The creation works as expected

feloy:
> > When the user restarts an app (IMO it is a very common operation, as the user will modify the sources to adapt the sample app, and will want to rebuild/restart the app), this should be done from the "Running AI Apps" page, but the steps (and the potential errors) are not visible anymore.
> 
> We could probably in another PR redirect to the StartRecipe page when we want to rebuild the recipe

I probably worked too much on `odo` to be objective, but it seems to me that we are decreasing a lot the experience when iterating the inner loop. Even by being redirected to the StartRecipe page, that will be multiple clicks, instead of a single click in the previous UX.

I would prefer that @slemeur or @jeffmaury validates the change

jeffmaury (changes requested):
Seems UI is inconsistent to me: when starting a recipe, I have to explicitely select a model but when I click on the select box, it seems a default model is selected

axel7083:
> but it seems to me that we are decreasing a lot the experience

The current user experience is terrible for the following reason
- Selecting a model in a different page than the starting button
- The RecipeModelDetails not being the same table as the one in the Models page
- The tasks only appear if you select the right models: people get lost
- This will not scale with a lot of models 

> that will be multiple clicks, instead of a single click

This is bad, because user lost controls of what they can do, having a single magic button doing everything prevent the user of doing basic actions. Maybe he want just to clone the recipe ?

#### Personal Opinion

- We improve the consistency of our design in AI Lab by having this page
- We simplify our codebase by cleaning the RecipeDetails code
- We make a dedicated page where we can **easily** redirect in the future using the Podman-desktop tasks action (when it will be supported)

feloy:
> > but it seems to me that we are decreasing a lot the experience
> 
> The current user experience is terrible for the following reason
> 
> * Selecting a model in a different page than the starting button
> * The RecipeModelDetails not being the same table as the one in the Models page
> * The tasks only appear if you select the right models: people get lost
> * This will not scale with a lot of models
> 
> > that will be multiple clicks, instead of a single click
> 
> This is bad, because user lost controls of what they can do, having a single magic button doing everything prevent the user of doing basic actions. Maybe he want just to clone the recipe ?
> 
> #### Personal Opinion
> * We improve the consistency of our design in AI Lab by having this page
> * We simplify our codebase by cleaning the RecipeDetails code
> * We make a dedicated page where we can **easily** redirect in the future using the Podman-desktop tasks action (when it will be supported)

Yes, I agree that the experience for **starting** a recipe (which happens once) is much better with your changes, but to the detriment of the experience for **restarting** the recipe every time the user changes the source (which can happen a lot of time if the developer want to develop an app using AI Lab).

axel7083:
> > > but it seems to me that we are decreasing a lot the experience
> > 
> > 
> > The current user experience is terrible for the following reason
> > 
> > * Selecting a model in a different page than the starting button
> > * The RecipeModelDetails not being the same table as the one in the Models page
> > * The tasks only appear if you select the right models: people get lost
> > * This will not scale with a lot of models
> > 
> > > that will be multiple clicks, instead of a single click
> > 
> > 
> > This is bad, because user lost controls of what they can do, having a single magic button doing everything prevent the user of doing basic actions. Maybe he want just to clone the recipe ?
> > #### Personal Opinion
> > 
> > * We improve the consistency of our design in AI Lab by having this page
> > * We simplify our codebase by cleaning the RecipeDetails code
> > * We make a dedicated page where we can **easily** redirect in the future using the Podman-desktop tasks action (when it will be supported)
> 
> Yes, I agree that the experience for **starting** a recipe (which happens once) is much better with your changes, but to the detriment of the experience for **restarting** the recipe every time the user changes the source (which can happen a lot of time if the developer wants to develop an app using AI Lab).

This can be done from the Running page ?

@jeffmaury  I added a default value (took from recommended) for the select. I also added unit tests for this change.

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/079d9069-a82b-4635-b368-0f83e4be6fd2

feloy:
> This can be done from the Running page ?

Yes, but what I explain in
https://github.com/containers/podman-desktop-extension-ai-lab/pull/1227#issuecomment-2175946785
is that from this page, it is not possible anymore to see the steps, and more importantly the potential errors, that the user were seeing from the Recipe page.

axel7083:
> > This can be done from the Running page ?
> 
> Yes, but what I explain in [#1227 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/1227#issuecomment-2175946785) is that from this page, it is not possible anymore to see the steps, and more importantly the potential errors, that the user were seeing from the Recipe page.

IMO this is a temporary issue, since we said we wanted to get a details page for the running recipes, but I agree in the current situation this is not ideal.

We may want to expose the building process inside the recipe details page, we would for examples have a section with building logic (list image being build, and some actions) what do you think ?

feloy:
> > > This can be done from the Running page ?
> > 
> > 
> > Yes, but what I explain in [#1227 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/1227#issuecomment-2175946785) is that from this page, it is not possible anymore to see the steps, and more importantly the potential errors, that the user were seeing from the Recipe page.
> 
> IMO this is a temporary issue, since we said we wanted to get a details page for the running recipes, but I agree in the current situation this is not ideal.
> 
> We may want to expose the building process inside the recipe details page, we would for examples have a section with building logic (list image being build, and some actions) what do you think ?
> 
> ### Edit
> I made a quick mockup of what I would imagine
> 
> image
> 
> I still think the building process should be separate from the running process, treating recipes as container file, their _built_ step would be the image created.
> 
> Today we have no places where we could redirect the std of the building process, we could add a `Logs` or `Build log` or anything, to the Recipes details page, and place inside the logs of the building process, which would correspond to the tasks developer would be interested to explore when editing the source code.
> 
> i also added a `Inspect` tab, to see the recipe information, as it is IMO very important

Yes, displaying the steps with the logs in a dedicated page seems a very good idea. I can imagine something similar to CI/CD pages displaying the steps of a pipeline, with their logs.

We have to take care to not mix up again a recipe with a couple (recipe, model).

jeffmaury:
Seems to me the state of the recipe is lost as I was able to start several time the same recipe and if I go back to the recipe catalog before the recipe is fully started seems the processus is cancelled

axel7083:
> Seems to me the state of the recipe is lost as I was able to start several time the same recipe

@jeffmaury the goal, would be to have the link back to the creation task, and being to cancel it through the tasks in Podman Desktop.

image

However, currently this is not possible, and would be addressed by https://github.com/containers/podman-desktop/issues/7709

> if I go back to the recipe catalog before the recipe is fully started seems the processus is cancelled

See the recipes as container images, if you go in podman desktop and start an image, go back to the image, you don't see the container, but it is still running. 

## Comited solution

In the same way downloading models display at the top of the catalog I added a task banner cc @lstocchi @jeffmaury 

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/d9dcd9af-14e8-404d-af53-83de4b5ce1c8

axel7083:
rebased

axel7083:
rebased

slemeur:
The page to start the recipe should be getting a title associated to the recipe the user is about to start:
Screenshot 2024-07-09 at 10 37 56

There should be indication about what are the information displayed and choices the user can do. The labels are not explicit enough.
Screenshot 2024-07-09 at 10 53 44

The experience is also making the flow very clutered compared to what we have at the moment. This is not a flow anymore that a user can just "TRY" with a single click experience. With the proposal the user, new to this world, now has to go to the list of recipes, go to the details of starting a recipe, once started, clicking on a button "open running apps", get into a new screen that is displaying detailed information. For me we are requiring the user to go into too many steps and things are disconnected. At the end, the user will not make the connection between all the things he/she went through. 
At the end, when I'm coming back to the screen of the recipe, I have no clue the app is already running, no way to find it back. 

Screenshot 2024-07-10 at 10 20 12

One of the benefit, is the "start" button which is becoming more prominent and consistent with other screens of Podman Desktop. 

In order to move forward with this, I'd recommend:
- We keep the "start" button in the main header of the recipe. 
- We keep the right sidepanel of the recipe, just removing the "start button"
- We introduce the page to start the recipe, but the end of the flow should redirect to the details of the recipe, the user will see the status of the recipe from the sidepanel.

axel7083:
@jeffmaury my proposal is to merge this PR, then I will open a PR for https://github.com/containers/podman-desktop-extension-ai-lab/issues/1360 and the design we discussed to display the running application

packages/frontend/src/lib/progress/TasksBanner.svelte
```diff
@@ -0,0 +1,29 @@
...
+  return tasks.subscribe(value => {
+    loadingTasks = value.filter(
+      task => task.state !== 'success' && labels.every(label => label in (task.labels ?? {})),
+    );
+  });
+});
+</script>
+
+{#if loadingTasks.length > 0}
+  <Card classes="bg-charcoal-800 mt-4 mx-5">
```
jeffmaury, line 23: Light mode ?

jeffmaury (commented):
I see a lot of non themes enabled colors is light mode being delayed then I think we should get an issue created

package.json
```diff
@@ -55,6 +55,10 @@
       "tests/*"
     ]
   },
+  "resolutions": {
```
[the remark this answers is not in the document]
axel7083: Same issue as bootc raised by charlie https://github.com/containers/podman-desktop-extension-bootc/pull/611

packages/frontend/src/lib/progress/TasksBanner.svelte
axel7083: Opened issue at https://github.com/containers/podman-desktop-extension-ai-lab/issues/1372

jeffmaury (approved):
LGTM

PR #1230, opened 2024-06-18
Use FormPage for Import Model page

### LINKED ISSUE #1229: use FormPage for Import Model page

feloy:
### What does this PR do?

Updtae the "Import Model" page to use the dedicated FormPage component instead of NavPage

Also change a few colors to support light mode (except colors in the form itself)

Fixes #1229  

### How to test this PR?

Navigate to Catalog > Import

### DIFF
--- a/packages/frontend/src/App.svelte
+++ b/packages/frontend/src/App.svelte
@@ -34,7 +34,7 @@ onMount(() => {
 </script>
 
 <Route path="/*" isAppMounted="{isMounted}" let:meta>
-  <main class="flex flex-col w-screen h-screen overflow-hidden bg-charcoal-700">
+  <main class="flex flex-col w-screen h-screen overflow-hidden bg-[var(--pd-content-bg)]">
     <div class="flex flex-row w-full h-full overflow-hidden">
       <Navigation meta="{meta}" />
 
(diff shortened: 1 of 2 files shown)


axel7083 (changes requested):
One small issue 

**Before**

| dark | light |
| --- | --- |
| image | image |

**After**

| dark | light |
| --- | --- |
| image | image |

We have a clear improvement on the light mode (nice job), but we loose the separation between the secondary navigation and the content on both theme

packages/frontend/src/App.svelte
```diff
@@ -34,7 +34,7 @@ onMount(() => {
 </script>
 
 <Route path="/*" isAppMounted="{isMounted}" let:meta>
-  <main class="flex flex-col w-screen h-screen overflow-hidden bg-charcoal-700">
+  <main class="flex flex-col w-screen h-screen overflow-hidden bg-[var(--pd-content-bg)]">
```
axel7083, line 37: I am wondering if we should not use `var(--pd-invert-content-bg)` as in the preferences, it is what is used however this would make the card of the form hidden :/

feloy:
> One small issue
> 
> [...]
> We have a clear improvement on the light mode (nice job), but we loose the separation between the secondary navigation and the content on both theme

Do you mean the vertical shadow between the 2 zones? If so, this issue tracks the problem: https://github.com/containers/podman-desktop/issues/7549

axel7083:
> Do you mean the vertical shadow between the 2 zones? If so, this issue tracks the problem:

Sorry if I was not clear enough, the fact that the main and secondary navigation are both the same color:

image

feloy:
> > Do you mean the vertical shadow between the 2 zones? If so, this issue tracks the problem:
> 
> Sorry if I was not clear enough, the fact that the main and secondary navigation are both the same color:
> 
> image

I double-checked, and for me the 2 colors were already before really the same (#343434), but the shadow in the middle makes one think they are different

PR #1238, opened 2024-06-19
Use latest ui-svelte version including DetailsPage

feloy:
### What does this PR do?

Use latest ui-svelte version including DetailsPage

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -12,7 +12,7 @@
     "watch": "vite --mode development build -w"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


axel7083:
I think there was a change to the Table component, that need to be addressed 

https://github.com/containers/podman-desktop-extension-ai-lab/blob/e19aec76f6b4bd54a4b3398b7738acfa6c3978c7/packages/frontend/src/pages/RecipeModels.svelte#L14-L19

I don't remember for sure, but I think @deboer-tim did some change to the way this information was stored ?

axel7083 (approved):
LGTM !

PR #1255, opened 2024-06-20
Light mode for model import page

### LINKED ISSUE #1254: chore: light mode `Model import` page

feloy:
### What does this PR do?

Light mode for model import page

select depends on https://github.com/containers/podman-desktop/issues/6199

Fixes #1254

### DIFF
--- a/packages/frontend/src/pages/ImportModel.svelte
+++ b/packages/frontend/src/pages/ImportModel.svelte
@@ -3,7 +3,7 @@ import { faCircleInfo, faFileImport } from '@fortawesome/free-solid-svg-icons';
 import { studioClient } from '../utils/client';
 import { Uri } from '@shared/src/uri/Uri';
 import type { LocalModelImportInfo } from '@shared/src/models/ILocalModelInfo';
-import { Button, ErrorMessage, FormPage, Tooltip } from '@podman-desktop/ui-svelte';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/pages/ImportModel.svelte
```diff
@@ -132,37 +133,36 @@ export function goToUpPage(): void {
...
           {:else}
             <!-- showing path -->
-            <label for="path" class="w-full block mb-2 text-sm font-bold text-gray-400">Path</label>
-            <input
-              class="flex w-full flex-col grow p-2 outline-none text-sm bg-transparent rounded-sm text-gray-700 placeholder-gray-700 border-charcoal-100 border-b mr-2"
-              bind:value="{localModel.path}"
-              name="path"
-              aria-label="model path"
-              readonly="{true}" />
+            <label for="path" class="w-full block mb-2 text-sm font-bold text-[var(--pd-content-card-header-text)]"
```
lstocchi, line 142: should it use text-[var(--pd-label-text)] ? The same for all labels. It is too light otherwise imo
image

lstocchi: Ops, there was an error while building and something got wrong. This is what i actually see
image

lstocchi (approved):
I wonder if we should use text-[var(--pd-label-text)] for labels but the result is ok like this as well

axel7083 (approved):

Issue #1083, opened 2024-05-13
After 10-20 questions the chatbot app is no longer working


benoitf:
### Is your enhancement related to a problem? Please describe

ask many questions like
"What is Podman Desktop" (waiting each time to get the answer)

After that you get an error or application hangs and consume all the CPU

### Describe the solution you'd like

No error or being properly asked to restart the app

benoitf:
I had this error sometimes but here I got

```
Llama.generate: prefix-match hit
```
in the logs

```
Llama.generate: prefix-match hit
INFO:     127.0.0.1:40230 - "GET / HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:40232 - "GET / HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:50622 - "GET / HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:50638 - "GET / HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:40224 - "POST /v1/chat/completions HTTP/1.1" 200 OK
INFO:     127.0.0.1:57906 - "GET / HTTP/1.1" 404 Not Found
```

and then all CPU is eaten but I never get an answer

benoitf:
but yes at summit I faced the `Requested tokens exceed context window` error

MichaelClifford:
OK, I think we can fix the context window issue. But I'm not sure about the "all-cpu-no-answer" issue. What hardware are you using?

benoitf:
forgot to hit send, hardware is Apple M2 Max
but I think I only reached once the 'no reply'

I think also https://github.com/containers/podman-desktop-extension-ai-lab/issues/1082 is more important

MichaelClifford:
@benoitf please see https://github.com/containers/ai-lab-recipes/pull/495

lstocchi:
We should show an error message if the user exceed the token limit

Issue #1021, opened 2024-04-30
Silent failure when text is too long when using the playground


lstocchi:
If you ask a question that has a very long content, it never replies back. If you go to the server logs you can see that it failed but nothing has been displayed to the user 

```
ValueError: Requested tokens (4237) exceed context window of 2048
INFO:     10.88.0.1:37666 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
Exception: Requested tokens (4237) exceed context window of 2048
Traceback (most recent call last):
  File "/usr/local/lib64/python3.9/site-packages/llama_cpp/server/errors.py", line 171, in custom_route_handler
    response = await original_route_handler(request)
  File "/usr/local/lib/python3.9/site-packages/fastapi/routing.py", line 278, in app
    raw_response = await run_endpoint_function(
  File "/usr/local/lib/python3.9/site-packages/fastapi/routing.py", line 191, in run_endpoint_function
    return await dependant.call(**values)
  File "/usr/local/lib64/python3.9/site-packages/llama_cpp/server/app.py", line 462, in create_chat_completion
    ] = await run_in_threadpool(llama.create_chat_completion, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/starlette/concurrency.py", line 42, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "/usr/local/lib/python3.9/site-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
  File "/usr/local/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 2144, in run_sync_in_worker_thread
    return await future
  File "/usr/local/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 851, in run
    result = context.run(func, *args)
  File "/usr/local/lib64/python3.9/site-packages/llama_cpp/llama.py", line 1657, in create_chat_completion
    return handler(
  File "/usr/local/lib64/python3.9/site-packages/llama_cpp/llama_chat_format.py", line 599, in chat_completion_handler
    completion_or_chunks = llama.create_completion(
  File "/usr/local/lib64/python3.9/site-packages/llama_cpp/llama.py", line 1493, in create_completion
    completion: Completion = next(completion_or_chunks)  # type: ignore
  File "/usr/local/lib64/python3.9/site-packages/llama_cpp/llama.py", line 972, in _create_completion
[...]

PR #1341, opened 2024-07-05
fix: display error message from chat conversation

lstocchi:
### What does this PR do?

This PR adds capabilities to display error messages during a chat. So far we just printed the errors in the console, but they are useful to guide the user. 

I extended the messages that a conversation includes. So far, we just had ChatMessages, but we could also have ErrorMessages. The only thing i'm not sure about is how to display them, so i chose to use what we have now (just show the newest error message if it's the last message of the conversation), but it could be very easy to display them as part of the conversation so when you scroll up you can see the error you received for an old input. Waiting to hear your preference.

### Screenshot / video of UI

image

it resolves https://github.com/containers/podman-desktop-extension-ai-lab/issues/1083 and https://github.com/containers/podman-desktop-extension-ai-lab/issues/1021

### How to test this PR?

1. open the playground
2. add a very long text ( > 2k chars), it should error (i used mistral model)

### DIFF
--- a/packages/shared/src/models/IPlaygroundMessage.ts
+++ b/packages/shared/src/models/IPlaygroundMessage.ts
@@ -27,9 +27,9 @@ export interface ErrorMessage extends Message {
   message: string;
 }
 
-export interface ChatMessage extends Message {  
+export interface ChatMessage extends Message {
   role: 'system' | 'user' | 'assistant';
-  content?: string;  
+  content?: string;
 }
 
 export interface AssistantChat extends ChatMessage {
--- a/packages/shared/src/models/IPlaygroundMessage.ts
+++ b/packages/shared/src/models/IPlaygroundMessage.ts
@@ -24,7 +24,7 @@ export interface Message {
 }
 
 export interface ErrorMessage extends Message {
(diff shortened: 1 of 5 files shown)


packages/shared/src/models/IPlaygroundMessage.ts
```diff
@@ -18,11 +18,18 @@
...
 import type { ModelOptions } from './IModelOptions';
 
-export interface ChatMessage {
+export interface Message {
   id: string;
+  timestamp: number;
+}
+
+export interface ErrorMessage extends Message {
+  message: string;
```
axel7083, line 27: IMO this is not explanatory enough, might have something like `error: string` instead ?

lstocchi: updated 👍

axel7083 (approved):

jeffmaury (changes requested):
Seems a unit test needs to be fixed

lstocchi:
@jeffmaury are you good with this?

jeffmaury (approved):
LGTM

Issue #1344, opened 2024-07-08
feat(Models): adding inspect tab with metadata information


axel7083:
### Is your feature request related to a problem? Please describe

Models in `gguf` format exposes a lot of metadata

### Describe the solution you'd like

We should add a tab, `Inspect` in the model detail page listing those metadata

PR #1280, opened 2024-06-24
feat(Inference): supporting gpu layers

axel7083:
### What does this PR do?

Add an `experimental GPU` preference. When enabled, it will uses the matching container image for the VMType used

- On WSL and with a NVIDIA GPU 
- On Libkrun => llamacpp mac vulkan 

### Screenshot / video of UI

New option `Experimental GPU` in podman desktop preferences

image

### Side by Side playground comparison 

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/6a8c5308-dc96-40bc-9200-210bd1fe9477

### What issues does this PR fix or reference?

Need https://github.com/containers/podman-desktop-extension-ai-lab/pull/1278 to be merged
Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1279

### How to test this PR?

**Require to be on windows with a WSL podman machine and a NVIDIA GPU**

- Turn on the experimental flag in the preference
- Set a value for the gpu offloading (E.g. 20)
- Start the inference server
- Go to playground, notice the speed

PR #1345, opened 2024-07-08
feat: showing models metadata

axel7083:
### What does this PR do?

Adding a Tab in the model detail page for GGUF models.

If the model is already downloaded, using it, otherwise using download url to fetch metadata.

This is possible thanks to the amazing library [@huggingface/gguf](https://www.npmjs.com/package/@huggingface/gguf)

### Screenshot / video of UI

https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/c7449021-e428-4202-bbc1-bd693afa0344

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1344

### How to test this PR?

### DIFF
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -34,7 +34,8 @@ import { deleteRemoteModel, getLocalModelFile, isModelUploaded } from '../utils/
 import { getFirstRunningMachineName } from '../utils/podman';
 import type { CancellationTokenRegistry } from '../registries/CancellationTokenRegistry';
 import { hasValidSha } from '../utils/sha';
-import { gguf, GGUFParseOutput } from '@huggingface/gguf';
+import type { GGUFParseOutput } from '@huggingface/gguf';
+import { gguf } from '@huggingface/gguf';
 
 export class ModelsManager implements Disposable {
   #models: Map<string, ModelInfo>;
@@ -430,16 +431,16 @@ export class ModelsManager implements Disposable {
 
     const before = performance.now();
     const data: Record<string, unknown> = {
-      'model-id': model.url?modelId:'imported', // filter imported models
+      'model-id': model.url ? modelId : 'imported', // filter imported models
     };
 
     try {
       let result: GGUFParseOutput<{ strict: false }>;
-      if(this.isModelOnDisk(modelId)) {
+      if (this.isModelOnDisk(modelId)) {
         const modelPath = path.normalize(getLocalModelFile(model));
         console.debug(`[getModelMetadata] reading model ${modelPath}`);
         result = await gguf(modelPath, { allowLocalFile: true });
-      } else if(model.url) {
+      } else if (model.url) {
         result = await gguf(model.url);
       } else {
         throw new Error('cannot get model metadata');
--- a/packages/backend/src/managers/modelsManager.ts
+++ b/packages/backend/src/managers/modelsManager.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 
 import type { LocalModelInfo } from '@shared/src/models/ILocalModelInfo';
-import fs from 'fs';
+import fs from 'node:fs';
 import * as path from 'node:path';
 import { type Webview, fs as apiFs, type Disposable, env } from '@podman-desktop/api';
 import { Messages } from '@shared/Messages';
@@ -34,6 +34,7 @@ import { deleteRemoteModel, getLocalModelFile, isModelUploaded } from '../utils/
 import { getFirstRunningMachineName } from '../utils/podman';
 import type { CancellationTokenRegistry } from '../registries/CancellationTokenRegistry';
 import { hasValidSha } from '../utils/sha';
+import { gguf, GGUFParseOutput } from '@huggingface/gguf';
 
 export class ModelsManager implements Disposable {
   #models: Map<string, ModelInfo>;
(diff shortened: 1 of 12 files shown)


packages/backend/src/managers/modelsManager.ts
```diff
@@ -422,4 +424,36 @@ export class ModelsManager implements Disposable {
...
+      if (this.isModelOnDisk(modelId)) {
+        const modelPath = path.normalize(getLocalModelFile(model));
+        console.debug(`[getModelMetadata] reading model ${modelPath}`);
+        result = await gguf(modelPath, { allowLocalFile: true });
+      } else if (model.url) {
+        result = await gguf(model.url);
+      } else {
+        throw new Error('cannot get model metadata');
+      }
+      console.log('result', result);
```
lstocchi, line 448: to be removed

packages/backend/src/managers/modelsManager.ts
```diff
@@ -422,4 +424,36 @@ export class ModelsManager implements Disposable {
...
+      return result.metadata;
+    } catch (err: unknown) {
+      data['error'] = err;
+      console.error(err);
+      throw err;
+    } finally {
+      data['duration'] = performance.now() - before;
+      this.telemetry.logUsage('get-metadata');
```
lstocchi, line 456:
```suggestion
      this.telemetry.logUsage('get-metadata', data);
```

packages/frontend/src/pages/ModelInspect.svelte
```diff
@@ -0,0 +1,81 @@
...
+
+onMount(() => {
+  return modelsMetadata.subscribe(metadatas => {
+    const records = metadatas[modelId];
+    if (records) {
+      items = Object.entries(records).map(([key, value]) => ({
+        name: key,
+        value: String(value),
+      }));
+      console.log(`found ${items.length} items`);
```
lstocchi, line 50: to be removed

packages/frontend/src/pages/ModelInspect.svelte
```diff
@@ -0,0 +1,81 @@
...
+  }),
+  new TableColumn<Item, string>('value', {
+    renderMapping: item => String(item.value),
+    renderer: TableSimpleColumn,
+  }),
+];
+const row = new TableRow<Item>({});
+
+onMount(() => {
+  return modelsMetadata.subscribe(metadatas => {
```
lstocchi, line 43: Why subscribing to a store if the only place where it is written it's here? Would it be simpler to have something like
```
studioClient
        .getModelMetadata(modelId)
        .then(result => {
        items = .....
```

without creating/subscribing to a store which is never updated elsewhere.

lstocchi (commented):
I'm a bit puzzled about the UI and some of the values displayed. TBH i'm not able to evaluate which are the values that can be useful but when they are very long they are cut, so not very readable.

E.g.
image

Then i think that using the UI that we use on the k8s resources would present the values in a much more clear way
image

packages/frontend/src/pages/ModelInspect.svelte
axel7083: If you leave the page and come back, this is made for caching it

axel7083:
> I'm a bit puzzled about the UI and some of the values displayed. 

I was not sure, I am open to suggestion, but we should keep in mind that those are metadatas. The most useful are probably the one related to the head count, which can be used as suggested by https://github.com/containers/podman-desktop-extension-ai-lab/pull/1280#issuecomment-2196310464

> Then i think that using the UI that we use on the k8s resources would present the values in a much more clear way 

@lstocchi I agree, but that would require to hardcode the keys to decide in advance where to place them, however those information are metadatas, and we are not sure for most of them where they will be. We might want to have it discuss in the UX call what do you think ?

packages/frontend/src/pages/ModelInspect.svelte
lstocchi: ah ok, please add a comment as a reminder

lstocchi:
> I was not sure, I am open to suggestion, but we should keep in mind that those are metadatas. The most useful are probably the one related to the head count, which can be used as suggested by [#1280 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/1280#issuecomment-2196310464)

> @lstocchi I agree, but that would require to hardcode the keys to decide in advance where to place them, however those information are metadatas, and we are not sure for most of them where they will be. We might want to have it discuss in the UX call what do you think ?

Sure. 
Also @slemeur and @jeffmaury can give some suggestions

axel7083:
> > I was not sure, I am open to suggestion, but we should keep in mind that those are metadatas. The most useful are probably the one related to the head count, which can be used as suggested by [#1280 (comment)](https://github.com/containers/podman-desktop-extension-ai-lab/pull/1280#issuecomment-2196310464)
> 
> > @lstocchi I agree, but that would require to hardcode the keys to decide in advance where to place them, however those information are metadatas, and we are not sure for most of them where they will be. We might want to have it discuss in the UX call what do you think ?
> 
> Sure. Also @slemeur and @jeffmaury can give some suggestions

I will convert it to draft then, however this is a requirement for https://github.com/containers/podman-desktop-extension-ai-lab/pull/1280 (which is my critical for this sprint)

slemeur:
I do think the table is not looking consistent with the other screens. 
If we want to quickly move forward on this, I'd recommend the format we have on the cards for the kubecontexts, which would just require to adjust the sizes.
Screenshot 2024-07-09 at 15 44 32

That being said, I agree with @lstocchi and would prefer we take the same approach as for the kube objects.
> but that would require to hardcode the keys to decide in advance where to place them, however those information are metadatas, and we are not sure for most of them where they will be. We might want to have it discuss in the UX call what do you think ?

I think you might take the first word/token of the metadata as the category for each sections. Example `tokenizer.dldld.dldld` would create a section `tokenizer` and then you'd have sub tables. 
The one with only one word/token, would go into "Globals" or something like that.

axel7083:
Closing for now

Issue #59, opened 2024-01-16
Version the catalog format


slemeur:

axel7083:
Some consideration we might want to have while writing the catalog specifications is how we specify the models information.

We probably need to find a way, since the models is provided by the catalog, to have more informations to get back to source, than just the download url.

## Models registry

### Hugging face

For example looking at hugging faces hub api to get models details we get 

```json
    {
        "_id": "621ffdc136468d709f17debc",
        "id": "microsoft/BiomedNLP-BiomedBERT-base-uncased-abstract-fulltext",
        "likes": 143,
        "private": false,
        "downloads": 92815,
        "tags": [
            "transformers",
            "pytorch",
            "jax",
            "bert",
            "fill-mask",
            "exbert",
            "en",
            "arxiv:2007.15779",
            "license:mit",
            "autotrain_compatible",
            "endpoints_compatible",
            "region:us"
        ],
        "pipeline_tag": "fill-mask",
        "library_name": "transformers",
        "createdAt": "2022-03-02T23:29:05.000Z",
        "modelId": "microsoft/BiomedNLP-BiomedBERT-base-uncased-abstract-fulltext"
    },
```

### MLflow

> MLflow is an open source platform to manage the ML lifecycle, including experimentation, reproducibility, deployment, and a central model registry.

They are providing model informations in a similar way than hugging face see [RegisteredModels](https://mlflow.org/docs/latest/rest-api.html#registeredmodel)

lstocchi:
I don't know if i misunderstood it yesterday but i thought we were going to add the version to the ai-studio.yaml file? 🤔 So if in future we add any breaking changes we can check the versioning and we know how to parse it?

But, by reading the title, it says catalog so i'm a bit confused.

lstocchi:
@axel7083 maybe it's more related to https://github.com/projectatomic/studio-extension/issues/45

axel7083:
> I don't know if i misunderstood it yesterday but i thought we were going to add the version to the ai-studio.yaml file? 🤔 So if in future we add any breaking changes we can check the versioning and we know how to parse it?

@lstocchi  I think this is the issue related to the ai-studio.yaml https://github.com/projectatomic/studio-extension/issues/19 

> @axel7083 maybe it's more related to #45

The thing is, we need to know where the models came from for https://github.com/projectatomic/studio-extension/issues/45, I cannot get metadata if I do not know where the models came from. 

[...]

PR #1381, opened 2024-07-16
feature: adding versioning to the catalog

axel7083:
### What does this PR do?

This PR set a version to the catalog, current `1.0`. When undefined or different, the user-catalog will not be loaded and a warning will be logged.

To ensure the user-catalog respect the format, the `CatalogUtils` file has been created, which will properly check for all the fields.

### Screenshot / video of UI

https://github.com/user-attachments/assets/744cc19d-bbea-4b46-a814-1da62b540684

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/59

### How to test this PR?

Reproduce video above

### DIFF
--- a/PACKAGING-GUIDE.md
+++ b/PACKAGING-GUIDE.md
@@ -14,6 +14,10 @@ Any change done to this file will also be automatically loaded by AI Lab.
 The catalog file has three main elements: categories, models and recipes. Each of these elements is
 represented in the JSON file as an array.
 
+The catalog is `versioned`. Current version can be found in [ai.json](https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/backend/src/assets/ai.json#L2).
+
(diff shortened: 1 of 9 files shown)


lstocchi (commented):
Any reason why we do not show any warning to the user so that it is visible without opening the console?

axel7083:
> Any reason why we do not show any warning to the user so that it is visible without opening the console?

We could @lstocchi. This is not the goal of this PR, I can open a follow up issue

jeffmaury (approved):
LGTM. I'm wondering if we should not issue a notification or report on UI that user catalog is invalid

axel7083:
> LGTM. I'm wondering if we should not issue a notification or report on UI that user catalog is invalid

Same remark as Luca, I will open an issue and add it to the backlog

axel7083:
Opened https://github.com/containers/podman-desktop-extension-ai-lab/issues/1394

lstocchi (approved):

PR #1391, opened 2024-07-17
feat: start REST server

feloy:
### What does this PR do?

[DRAFT]


api/openapi.yaml
```diff
@@ -0,0 +1,1026 @@
+openapi: 3.1.0
```
axel7083, line 1: Why do we need this file ?

feloy: I would like to use a library like https://github.com/cdimascio/express-openapi-validator to validate the requests/responses based on the openapi spec (the file comes from https://github.com/ollama/ollama/pull/5040)

axel7083: I would suggest another approach. We should write our own `ai-lab-spec.yaml` file with the `open api` specification and use open source tool like [openapi-generator.tech](https://openapi-generator.tech/) or [express-openapi](https://www.npmjs.com/package/express-openapi) to directly generate the server.

## Why ?

This will ensure type safety of our implementation matching the open api spec. Thus allowing users (E.g. quarkus) to generate a client using tools, to directly be able to uses our API.

## Personal opinion

This is an important architectural choice, and should be discussed before using/choosing any framework.

feloy: As discussed with @jeffmaury and @slemeur , we would like to implement the Ollama api, thus the use of this specific openapi spec and not a ai-lab specific one.

I already looked at https://openapi-generator.tech/, but there is only one node server generator, in beta, which writes Javascript (not Typescript) code, which seems not optimal. I will have a look at https://www.npmjs.com/package/express-openapi.

I completely agree with your constraints in `Why?`, which would also be respected with the https://github.com/cdimascio/express-openapi-validator tool and the openapi spec provided by ollama. The only difference is that the creation of the server and the definition of the routes would be to be written by ourselves, and the types would be generated by https://github.com/openapi-ts/openapi-typescript. Type safety would be ensured by https://github.com/cdimascio/express-openapi-validator in real-time. In any case, the clients can use the ollama openapi spec to generate their client code. And because it is ollama compatible, the client is most probably already generated somewhere.

jeffmaury (commented):
The issue mentioned using defaut port 10434 (ollama - 1000) and if busy allocate a random port and store the value in a preference so that external scripts/tools have a way to get the port value

Not sure we should use the Ollama API spec file as this as we will implement incrementally and expose a Swagger endpoint

feloy:
> The issue mentioned using defaut port 10434 (ollama - 1000) and if busy allocate a random port and store the value in a preference so that external scripts/tools have a way to get the port value

> 
> Not sure we should use the Ollama API spec file as this as we will implement incrementally and expose a Swagger endpoint

This PR is a very very early draft, just to share the code importing `express`, and fix the problem with vite not importing it correctly.

api/openapi.yaml
```diff
@@ -0,0 +1,1026 @@
+openapi: 3.0.0
```
axel7083, line 1: Moving to `packages/backend/src/assets` ?

axel7083: I think the openapi file should not contain in the `main` any api that is not supported.

feloy: I'm not sure. `assets/ai.json` at build time and not copied to the user's disk. For `openapi.json`, it is different, as it needs to be copied to the disk, and read at runtime.

feloy: > I think the openapi file should not contain in the `main` any api that is not supported.

OK, I'll remove the unimplemented endpoints and related definitions

packages/backend/vite.config.js
```diff
@@ -50,6 +50,8 @@ const config = {
       external: [
         '@podman-desktop/api',
         '@huggingface/gguf',
+        'express',
```
axel7083, line 53: This turn out not to be working out in production build, as it completely exclude the package and make it not being packed in the output.

axel7083 (changes requested):

feloy:
When running Podman Desktop with the extension (using `yarn watch --extension-folder path/to/extension` or by installing it with a distribution image), this error occurs:

ai-lab-ext-error

The problem is reproducible with other extensions as well (tested with kube-context extension).

axel7083:
Do do you have some explanation on how you solved the issue ? 

- I see that you added a new package `supertest` why ? to bypass the `request` import ? Maybe @dgolovin have an opinion on this, (I think we are overwriting the request or http package, may be linked to https://github.com/containers/podman-desktop/issues/7833)

You are copying the full node_modules in the containerfile, `COPY node_modules/ /extension/node_modules` not sure about that, when we are suppose to have the modules bundles properly by rollout (cc @benoitf )

I see that you added a new step in the release workflow `Prepare node_modules for prod` why ?

feloy:
> Do do you have some explanation on how you solved the issue ?
> 
> * I see that you added a new package `supertest` why ? to bypass the `request` import ? Maybe @dgolovin have an opinion on this, (I think we are overwriting the request or http package, may be linked to [extensions using npm package with dependency on `http` leads to crash podman-desktop#7833](https://github.com/containers/podman-desktop/issues/7833))
> 

I have added the supertest package to write the tests in apiServer.spec.ts

> You are copying the full node_modules in the containerfile, `COPY node_modules/ /extension/node_modules` not sure about that, when we are suppose to have the modules bundled properly by rollout (cc @benoitf )
> 
> I see that you added a new step in the release workflow `Prepare node_modules for prod` why ?

Yes, this is a proposal from Florent to copy the node_modules files to the image if we cannot bundle express and express-openapi-validator properly (and for now, I cannot find a way to do it). The step `Prepare node_modules for prod` is to limit the content of node_modules to the production packages before copying them to the image.

packages/backend/src/managers/apiServer.ts
```diff
@@ -0,0 +1,158 @@
...
+import { existsSync } from 'fs';
+import { getFreeRandomPort } from '../utils/ports';
+import * as podmanDesktopApi from '@podman-desktop/api';
+import { readFile } from 'fs/promises';
+
+const DEFAULT_PORT = 10434;
+const SHOW_API_INFO_COMMAND = 'ai-lab.show-api-info';
+
+export class ApiServer implements Disposable {
+  listener?: Server;
```
axel7083, line 18: could be private ?

feloy: It was first, but it needs to be accessed from the unit tests

axel7083: I would rather have it private, have a `protected getListener` and have a TestApiServer exposing public the getListener.

This approach is often used in podman-desktop

https://github.com/containers/podman-desktop/blob/0b77cf2d6b70fb559abd5b4a1891a195a5d6950a/extensions/podman/scripts/podman-download.spec.ts#L59

https://github.com/containers/podman-desktop/blob/7d997640c368bb62037bed8ed337c732d78b3484/packages/main/src/tray-animate-icon.spec.ts#L27

https://github.com/containers/podman-desktop/blob/0698e4d2fd1f3fcb64bfc8d66a96003545b6ee89/packages/main/src/plugin/webview/webview-registry.spec.ts#L56

axel7083 (approved):
Work on my machine, the endpoints are working as expected.

I think we said in the last meeting that the port should be configurable in the preference, and the port should not be random, it should fail if the port is not free.

feloy:
> Work on my machine, the endpoints are working as expected.
> 
> I think we said in the last meeting that the port should be configurable in the preference, and the port should not be random, it should fail if the port is not free.

@jeffmaury what do we want to do about the port, random or configurable?

jeffmaury:
> > Work on my machine, the endpoints are working as expected.
> > I think we said in the last meeting that the port should be configurable in the preference, and the port should not be random, it should fail if the port is not free.
> 
> @jeffmaury what do we want to do about the port, random or configurable?

My proposition was to use a default value (from settings) if binds fails generate random value and update settings

feloy:
> > > Work on my machine, the endpoints are working as expected.
> > > I think we said in the last meeting that the port should be configurable in the preference, and the port should not be random, it should fail if the port is not free.
> > 
> > 
> > @jeffmaury what do we want to do about the port, random or configurable?
> 
> My proposition was to use a default value (from settings) if binds fails generate random value and update settings

OK, I have created a follow-up issue #1426

benoitf:
nice PR, the only thing that is worrying me is that we include all the production dependencies while we would need only express dependencies.
It means that it'll increase the size of the extension a lot and on the filesystem for all people installing it

for now the image size is 44.1 MB
```
podman images | grep ai-lab
ghcr.io/containers/podman-desktop-extension-ai-lab  c6afaa0ed63601394bff38301acc8698db477e63  9afe74e53326  14 hours ago  44.1 MB
```

feloy:
> nice PR, the only thing that is worrying me is that we include all the production dependencies while we would need only express dependencies. It means that it'll increase the size of the extension a lot and on the filesystem for all people installing it
> 
> for now the image size is 44.1 MB
> 
> ```
> podman images | grep ai-lab
> ghcr.io/containers/podman-desktop-extension-ai-lab  c6afaa0ed63601394bff38301acc8698db477e63  9afe74e53326  14 hours ago  44.1 MB
> ```

I have added a commit to move other dependencies to devDependencies, which reduces the image size to 21MB.

feloy:
I have tested other solutions to replace `express` and `express-openapi-validator`, which cannot be bundled with `vite`:
- `fastify`: a mandatory dependency `pino` cannot be bundled
- `restana`: the http server/router can be bundled, but the module to handle OpenAPI validation (`restana-swagger-validator`) cannot be bundled

feloy:
With the latest commit (removing sub-dependencies of express/express-openapi-validator already bundled into dist/extension.js), only 6MB are added into node_modules, which makes the image grow from 44MB to 50MB.

jeffmaury:
Replaced by #1544

PR #1393, opened 2024-07-17
chore: moving dev deps to dev

axel7083:
### What does this PR do?

Some deps are in the dev dependencies when they should not


lstocchi (approved):
LGTM, i wonder why we never face any issue

benoitf:
> LGTM, i wonder why we never face any issue

in fact as we're using a bundler (vite) at the end there is mostly one .js produced embedding the libs

PR #1396, opened 2024-07-17
feat(backend): split application manager

axel7083:
### What does this PR do?

The application manager is too big, and is doing too much thing, without talking about the `ApplicationManager.spec.ts` which is more than `1k` lines.

This PR simplify the ApplicationManager by adding a RecipeManager.

Here a break down of responsibilities

### Application Manager

- `create | delete | start | remove` pods
- create containers
- adopt pods

### Recipe Manager 

- clone recipes
- build recipe container images
- read `ai-lab configuration`

### How to test this PR?

- [x] lot of tests has been written :)


feloy (approved):
Tested by doing different operations on AI Apps, works as expected

Thanks for the refactoring

PR #1397, opened 2024-07-17
fix: remove gguf from external rollup

axel7083:
### What does this PR do?

Remove from the rollup external the gguf library: it lead to `@huggingface/gguf` not found when starting the extension in production.


axel7083:
Tested with `quay.io/rh-ee-astefani/demo-ai-lab:local-20240717173049` which is a locally built extension

Issue #1213, opened 2024-06-13
Adding explore api button in InferenceServer


axel7083:
### Is your enhancement related to a problem? Please describe

The inference server we are using, exposes a `/docs` 

![image](https://github.com/containers/podman-desktop-extension-ai-lab/assets/42176370/cc953d6a-6fbe-486d-a76f-8a1e39c71e08)

### Describe the solution you'd like

we should have `Explorer API` button which open the browser to this page

### Additional context

cc @slemeur

PR #1412, opened 2024-07-22
feat: makes inference server details link clickable

axel7083:
### What does this PR do?

When the service is running, makes the url link clickable. It will open the default browser 

### Screenshot / video of UI

image

https://github.com/user-attachments/assets/fc160c27-cbaa-4682-8165-1d21f2d38300

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1213

### How to test this PR?

**manually**

Go to inference service details, click on the link

### DIFF
--- a/packages/frontend/src/pages/InferenceServerDetails.svelte
+++ b/packages/frontend/src/pages/InferenceServerDetails.svelte
@@ -201,7 +201,7 @@ export function goToUpPage(): void {
                     <button
                       on:click={() =>
                         service && studioClient.openURL(`http://localhost:${service.connection.port}/docs`)}
-                      class="bg-charcoal-600 rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center underline">
+                      class="bg-[var(--pd-label-bg)] text-[var(--pd-label-text)] rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center underline">
                       http://localhost:{service.connection.port}/docs
                       <Fa class="ml-2" icon={faArrowUpRightFromSquare} />
                     </button>
--- a/packages/frontend/src/pages/InferenceServerDetails.svelte
+++ b/packages/frontend/src/pages/InferenceServerDetails.svelte
@@ -199,12 +199,11 @@ export function goToUpPage(): void {
                 <div class="flex flex-row gap-4">
                   {#if service.status === 'running' && service.type === InferenceType.LLAMA_CPP}
                     <button
-                      on:click="{() =>
-                        service &&
-                        studioClient.openURL(`http://localhost:${service.connection.port}/docs`)}"
+                      on:click={() =>
+                        service && studioClient.openURL(`http://localhost:${service.connection.port}/docs`)}
                       class="bg-charcoal-600 rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center underline">
                       http://localhost:{service.connection.port}/docs
-                      <Fa class="ml-2" icon="{faArrowUpRightFromSquare}" />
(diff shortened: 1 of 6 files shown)


lstocchi (approved):
LGTM

packages/frontend/src/pages/InferenceServerDetails.svelte
```diff
@@ -192,10 +199,15 @@ export function goToUpPage(): void {
...
                 <div class="flex flex-row gap-4">
-                  <div
-                    class="bg-[var(--pd-label-bg)] text-[var(--pd-label-text)] rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center">
-                    http://localhost:{service.connection.port}/v1
-                  </div>
+                  {#if service.status === 'running' && service.type === InferenceType.LLAMA_CPP}
+                    <button
+                      on:click={() =>
+                        service && studioClient.openURL(`http://localhost:${service.connection.port}/docs`)}
+                      class="bg-charcoal-600 rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center underline">
```
lstocchi, line 206: Maybe using dynamic bg for light mode?

axel7083, line 206:
```suggestion
                      class="bg-[var(--pd-label-bg)] text-[var(--pd-label-text)] rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center underline">
```

packages/frontend/src/pages/InferenceServerDetails.svelte
```diff
@@ -192,10 +199,15 @@ export function goToUpPage(): void {
               <div class="bg-[var(--pd-content-card-bg)] rounded-md w-full px-4 pt-2 pb-4 mt-2">
                 <span class="text-sm text-[var(--pd-content-card-text)]">Server</span>
                 <div class="flex flex-row gap-4">
-                  <div
-                    class="bg-[var(--pd-label-bg)] text-[var(--pd-label-text)] rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center">
-                    http://localhost:{service.connection.port}/v1
-                  </div>
+                  {#if service.status === 'running' && service.type === InferenceType.LLAMA_CPP}
```
jeffmaury, line 202: I think that the value of the URL should be computed by the Inference server implementation (llama_cpp,...) as the /docs suffix is not generic. And return null will condition the presence of the button without llama_cpp specific code here

slemeur:
I'm worry this change is making the use of the inference server with other tool (application code, other LLMs tools) harder. 
As a developer, I'd like to be able to very quickly copy the base URL of the inference server so I can paste it in my code. If I need documentation on the API end points, then, I'd like to have a direct access to the API documentation. 

As a result, I'd be in favor of having a different button to access the documentation and let the user still be able to "copy" the URL of the inference server.

packages/frontend/src/pages/InferenceServerDetails.svelte
axel7083: I don't see an easy way of doing this, as the LlamaCpp Inference provider is a worker, not a manager that can be reused

jeffmaury: Maybe the connection field should be expanded to contain not only the port

axel7083:
Thanks to https://github.com/containers/podman-desktop-extension-ai-lab/pull/1443 inference server have a labels property, which is used to have an optional `api` or `docs` label.

packages/frontend/src/pages/InferenceServerDetails.svelte
axel7083: The value of the field is now the responsability of the Inference Provider!

PR #1414, opened 2024-07-22
feat(ui): create SelectModel component

axel7083:
### What does this PR do?

We are offering a Select form in `CreateInference`, `CreatePlayground` and `Start Recipe` page, let's create a single component, easier to style for light mode.

> :warning: I did not change the styling made to the component that were used for StartRecipe. 

### Screenshot / video of UI

image

image

image

N/A

### How to test this PR?

### DIFF
--- a/packages/frontend/src/lib/ModelSelect.svelte
+++ b/packages/frontend/src/lib/ModelSelect.svelte
@@ -15,15 +15,11 @@ export let recommended: string[] | undefined = undefined;
  */
 export let models: ModelInfo[];
 
-let sorted: ModelInfo[];
-
-/**
- * Put the recommended models at the top
- */
-$: {
-  sorted = models.toSorted((a, b) => {
-    return (recommended?.includes(a.id) ? 0 : 1) - (recommended?.includes(b.id) ? 0 : 1);
-  });
+function getModelSortingScore(modelInfo: ModelInfo): number {
+  let score: number = 0;
+  if (modelInfo.file) score -= 1;
+  if (recommended?.includes(modelInfo.id)) score -= 1;
+  return score;
 }
 
 /**
@@ -63,7 +59,9 @@ $: {
   --border-focused="var(--pd-input-field-focused-bg)"
   placeholder="Select model to use"
   class="!bg-[var(--pd-content-bg)] !text-[var(--pd-content-card-text)]"
-  items={sorted.map(model => ({ ...model, value: model.id, label: model.name }))}
+  items={models
+    .toSorted((a, b) => getModelSortingScore(a) - getModelSortingScore(b))
+    .map(model => ({ ...model, value: model.id, label: model.name }))}
   showChevron>
   <div slot="item" let:item>
     <div class="flex items-center">
--- a/packages/frontend/src/lib/ModelSelect.svelte
+++ b/packages/frontend/src/lib/ModelSelect.svelte
@@ -1,0 +1,73 @@
+<script lang="ts">
+import { faCheckCircle, faDownload } from '@fortawesome/free-solid-svg-icons';
+import Select from 'svelte-select';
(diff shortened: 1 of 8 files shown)


jeffmaury (commented):
Can we get recommanded models to be listed first ?

axel7083:
> Can we get recommanded models to be listed first ?

@jeffmaury no problem, I added it

image

packages/frontend/src/lib/ModelSelect.svelte
```diff
@@ -0,0 +1,84 @@
...
+
+let sorted: ModelInfo[];
+
+/**
+ * Put the recommended models at the top
+ */
+$: {
+  sorted = models.toSorted((a, b) => {
+    return (recommended?.includes(a.id) ? 0 : 1) - (recommended?.includes(b.id) ? 0 : 1);
+  });
```
benoitf, line 26: should downloaded models be sorted first as well ? I guess that if I have downloaded a model I don't want to be listed at the end.

packages/frontend/src/pages/CreateService.spec.ts
```diff
@@ -92,6 +92,8 @@ beforeEach(() => {
   mocks.getInferenceServersMock.mockReturnValue([
     { container: { containerId: 'dummyContainerId' } } as InferenceServer,
   ]);
+
+  window.HTMLElement.prototype.scrollIntoView = function () {};
```
benoitf, line 96: should it be an arrow function ?

packages/frontend/src/lib/ModelSelect.svelte
```diff
@@ -0,0 +1,84 @@
...
+export let disabled: boolean = false;
+/**
+ * Recommended model ids
+ */
+export let recommended: string[] | undefined = undefined;
+
+/**
+ * List of models
+ */
+export let models: ModelInfo[];
```
benoitf, line 16: for new components, should we leverage svelte5 or not ?

like
let {disabled, recommended....... = $props();

and use runes rather than `$:`

benoitf (approved):

packages/frontend/src/lib/ModelSelect.svelte
axel7083: I am looking at the experimental documentation, and I don't find a nice a way to type the `$props();` 

https://component-party.dev/?f=svelte4,svelte5#component-composition.props

packages/frontend/src/pages/CreateService.spec.ts
axel7083: Yes, I made it a `vi.fn` at the end

packages/frontend/src/lib/ModelSelect.svelte
axel7083: Good idea

axel7083:
After applying @benoitf remark, the models are sorted as followed 

1. recommended downloaded models
2. recommended not downloaded models
3. downloaded models
4. not downloaded models

image

jeffmaury (approved):
LGTM.

However, I see a warning when running `yarn build`:
```

[0] vite v5.3.4 building for production...
[0] transforming...
[0] 15:12:48 [vite-plugin-svelte] C:/Users/Jeff/work/src/github.com/containers/ai-lab/node_modules/svelte-select/Select.svelte:828:12 Self-closing HTML tags for non-void elements are ambiguous — use `<select ...></select>` rather than `<select ... />`
[0] 826:     {#if required && (!value || value.length === 0)}
[0] 827:         <slot name="required" {value}>
[0] 828:             <select class="required" required tabindex="-1" aria-hidden="true" />
[0]                                                                                       ^
[0] 829:         </slot>
[0] 830:     {/if}
[1] ✓ 3022 modules transformed.
```

Issue #1089, opened 2024-05-15
Model Playground  - Stop output button


myfear:
### Is your enhancement related to a problem? Please describe

I am always frustrated when the model generates a long answer that I see is going into the wrong direction and I want to stop the generation of the response.

### Describe the solution you'd like

I would love a "Stop Generation" button in the shape of a "STOP" (square) button to be available.

### Describe alternatives you've considered

Some other icon or text would work too. 

### Additional context

The output continuously scrolls down to the last generated line from the model. I can not scroll upwards while the response is triggering in.

PR #1429, opened 2024-07-24
feature: adding stop button playground

axel7083:
### What does this PR do?

Adding a Stop button to the playground when the LLM is responding. 

### Screenshot / video of UI

https://github.com/user-attachments/assets/f833fe0c-c175-4a7b-9df4-8b45a0022de6

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1089

### How to test this PR?

**Manually**

- Create playground
- Submit message
- Stop message
- Again

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.spec.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.spec.ts
@@ -27,6 +27,7 @@ import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import type { TaskRegistry } from '../registries/TaskRegistry';
 import type { Task, TaskState } from '@shared/src/models/ITask';
 import type { ChatMessage, ErrorMessage } from '@shared/src/models/IPlaygroundMessage';
+import type { CancellationTokenRegistry } from '../registries/CancellationTokenRegistry';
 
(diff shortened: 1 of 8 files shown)


axel7083:
> Also, should we indicate somewhere that the response has been aborted ?

The server does not precise if it stop because the LLM provide a stop token or because he received a abort signal. We could incude this info in the Message interface from the client side, but I would rather not for now.

ekidneyrh:
Yah we should have the buttons follow the same format as those on the details page

image

There's not a format decided upon for the outline / secondary version of it yet, but you could make two separate buttons side by side, one for send and one for stop?

axel7083:
I updated the design and the screenshot to match the discussion we had in the UI call

benoitf (approved):
just wondering if the creation of the manager object should not be in the beforeEach method as it's a lot of copy paste everywhere and each we add a new parameter we need to change all tests methods

axel7083:
> just wondering if the creation of the manager object should not be in the beforeEach method as it's a lot of copy paste everywhere and each we add a new parameter we need to change all tests methods

Yes, I rewrote those pretty old tests.

PR #1433, opened 2024-07-24
chore(models): remove the hw property

axel7083:
### What does this PR do?

The `hw` property was useless and should not be on the model interface as it does not depend on the model. Removing it, it could be re-introduced on the InferenceServer, but currently it is missleading with the GPU acceleration.

### Screenshot / video of UI

**before**

image

**after**

image

### How to test this PR?


axel7083:
See https://github.com/containers/podman-desktop-extension-ai-lab/issues/1434 for follow-up

jeffmaury (changes requested):
Putting on hold as it was part of the initial mockup and needs discussion

benoitf (approved):

jeffmaury (approved):
LGTM

PR #1448, opened 2024-07-29
hotfix: remove MODEL prefix

axel7083:
### What does this PR do?

The image we are using do not have the `MODEL_` prefix we place on every environment.

See https://github.com/containers/ai-lab-recipes/blob/f9ed8bbcbd518562518fae63fbd1234464f554f4/model_servers/llamacpp_python/src/run.sh#L18

Therefore we have are not able to provide the right `chatFormat`.

### DIFF
--- a/packages/backend/src/utils/modelsUtils.ts
+++ b/packages/backend/src/utils/modelsUtils.ts
@@ -78,7 +78,7 @@ export function getModelPropertiesForEnvironment(modelInfo: ModelInfo): string[]
     envs.push(
       ...Object.entries(modelInfo.properties).map(([key, value]) => {
         const formattedKey = key.replace(/[A-Z]/g, m => `_${m}`).toUpperCase();
-        return `MODEL_${formattedKey}=${value}`;
+        return `${formattedKey}=${value}`;
       }),
     );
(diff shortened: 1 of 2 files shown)


jeffmaury (commented):
That was part of the initial spec so I don't understand why we rollback when some external pieces do not follow

axel7083:
> That was part of the initial spec so I don't understand why we rollback when some external pieces do not follow

They never supported it on https://github.com/containers/ai-lab-recipes, but we switched to use their image because they provide the CUDA, vulkan, romc images.

So in the current state we have a regression as we cannot set the CHAT_FORMAT, because they do not use the prefix

benoitf:
> They never supported it on https://github.com/containers/ai-lab-recipes

it was supported until this PR https://github.com/containers/ai-lab-recipes/pull/531

axel7083:
> > They never supported it on https://github.com/containers/ai-lab-recipes
> 
> it was supported until this PR [containers/ai-lab-recipes#531](https://github.com/containers/ai-lab-recipes/pull/531)

Oh, I hadn't see that one passing through...  that's annoying.

PR #1475, opened 2024-08-05
fix: try to convert to current format catalog having no version

### LINKED ISSUE #1437: Improve the migration of the user catalog

When a user catalog exists and is not versioned, we ignore it and display a notification. This can be improved by performing a conversion.

Conversion:

- add `version` 1.0
- assume `backend` is `llama-cpp` (for models and recipes)
- for recipies copy `models` field as `recommended`

_No response_

### Additional context

_No response_

lstocchi:
### What does this PR do?

This PR enhances the sanitization method to try converting automatically a catalog not having any version to the CURRENT format. If succeed, the user will not see any notification about invalid format.

### Screenshot / video of UI

N/A

it resolves #1437 

### How to test this PR?

1. have a user catalog without any version, you should still see its model/recipes


axel7083 (approved):
LGTM, it works great ! Thanks for taking the time 

My only question: should we save the catalog after migrating it ? Currently it load it, but does not overwrite the file, we have to import/delete something manually to have it updated.

packages/backend/src/utils/catalogUtils.ts
```diff
@@ -26,7 +26,19 @@ export enum CatalogFormat {
...
 }
 
-export function sanitize(raw: object & { version: string }): ApplicationCatalog {
+export function sanitize(rawObject: object): ApplicationCatalog {
+  // if there is no version in the user catalog, we try to adapt it automatically to the CURRENT format
+  let raw: object & { version: string };
+  if (!('version' in rawObject)) {
+    raw = {
+      ...rawObject,
+      version: CatalogFormat.CURRENT,
```
deboer-tim, line 35: Setting any unknown file version to current here (before even trying to adapt) threw me off at first. IMHO it should be `adaptToCurrent(raw:object): object & { version: string }` which adds the version as the last step before returning. (and maybe throws an error if the file is missing some key element and can't be converted)

packages/backend/src/utils/catalogUtils.ts
```diff
@@ -26,7 +26,19 @@ export enum CatalogFormat {
...
+    };
+    raw = adaptToCurrent(raw);
+  } else {
+    raw = rawObject as object & { version: string };
+  }
+
   // ensure version is valid
   if (raw.version !== CatalogFormat.CURRENT) throw new Error('the catalog is using an invalid version');
```
deboer-tim, line 43: If v2 was current and I have a file with `version: 1.0` or 'version: 3.0' it would be rejected, but if I remove the version we would try to convert it to v2 and proceed. As a user, why would I ever want the version in my file if it is less likely to work? :) It seems like we should attempt to convert older versions, be more picky about what we convert, and/or use semantic versioning (e.g. maybe v1 would be converted but v3 would fail).

lstocchi: I see what you're saying but i guess it is too early to anticipate some developments imo. The issue was created bc before there was no version and now we are switching to v1 and we wanted to find a way to avoid showing the notification "your catalog is not compatible". 

> As a user, why would I ever want the version in my file if it is less likely to work?

Well, the version is the value that allows us to know which is the schema we have to use to parse the file. If i have to parse a catalog with unknown version i'm going to look for a `models` field, but in v1 it is called `recommended`. So if you have a catalog v1 and you remove the version field to have it it magically converted to v3 it would failbc we're not able to detect some values.
We could handle it differently of course, but it is too early to think about these things.

ATM we just have 2 cases: unknown (which is the older version) and v1. When we will have more versions we could start thinking which conversion we would want to support and how.

Maybe v3 will have a completely new schema, with new mandatory values, and it would be impossible to convert from an older version to it🤷‍♂️

deboer-tim: I guess my point is also:
- v2, v3 -> 'invalid version'
- random file, v3 with version removed, etc -> 'ok, let's convert!'

Maybe just check for one known element in the file before converting? Makes it a little more explicit that 'we only convert unknown files that look like <v1' and if v2, v3 are added later it is an explicit step to decide which version(s) to convert from vs default to convert anything.

lstocchi: The difference between unknown (older) and v1 (new) is that now there is the backend field and that the models has been replaced by recommended. So you would prefer, instead of checking for the unknown version, to have a check like `if this catalog, no matter its version (unknown, v2, v2.5, ...), has the models field and missed the backend field -> try to convert`?

deboer-tim: No, I meant add a simple check if models exists - i.e. this is most likely the <v1 format and safe to convert. It just makes it explicit that we're only trying to convert this one known format.

lstocchi: Updated, hope i did it right 😅

benoitf (approved):

lstocchi:
> > My only question: should we save the catalog after migrating it ? Currently it load it, but does not overwrite the file, we have to import/delete something manually to have it updated.
> 
> This is something i wasn't sure. My thought was that the file should be overwritten only when the user updates it. So there is a clear action by them to update it. But as a said i'm not sure what Jeff wanted. I'll keep this open until he's back

After talking with Jeff, the catalog should be overwritten after sanitized. I updated the PR.

jeffmaury (approved):
LGTM

PR #1503, opened 2024-08-07
use inference server when model service image is equivalent to inference server image

lstocchi:
### What does this PR do?

This PR is just a POC about using an inference server instead of creating the model service container inside the pod with the recipe app container.

The problem that we may face is that there is no link between the recipe and the inference server in the UI so the user does not know that by stopping the inference server the recipe will not work.

It also misses a new task for the creation/reuse of an inference server
The image id is hardcoded 
The image in the chatbot recipe in ai-lab-recipes is wrong

### DIFF
--- a/packages/backend/src/managers/inference/inferenceManager.ts
+++ b/packages/backend/src/managers/inference/inferenceManager.ts
@@ -109,8 +109,8 @@ export class InferenceManager extends Publisher<InferenceServer[]> implements Di
     // check if model backend is supported
     const backend: InferenceType = getInferenceType([model]);
     const providers: InferenceProvider[] = this.inferenceProviderRegistry
-        .getByType(backend)
-        .filter(provider => provider.enabled());
+      .getByType(backend)
+      .filter(provider => provider.enabled());
     if (providers.length === 0) {
       throw new Error('no enabled provider could be found.');
     }
--- a/packages/backend/src/managers/inference/inferenceManager.ts
+++ b/packages/backend/src/managers/inference/inferenceManager.ts
@@ -105,7 +105,7 @@ export class InferenceManager extends Publisher<InferenceServer[]> implements Di
    * return the first inference server which is using the specific model
    * it throws if the model backend is not currently supported
    */
-  public getServerByModel(model: ModelInfo): InferenceServer | undefined {
+  public findServerByModel(model: ModelInfo): InferenceServer | undefined {
     // check if model backend is supported
     const backend: InferenceType = getInferenceType([model]);
     const providers: InferenceProvider[] = this.inferenceProviderRegistry
--- a/packages/backend/src/managers/inference/inferenceManager.ts
+++ b/packages/backend/src/managers/inference/inferenceManager.ts
@@ -102,6 +102,22 @@ export class InferenceManager extends Publisher<InferenceServer[]> implements Di
   }
 
   /**
+   * return the first inference server which is using the specific model
+   * it throws if the model backend is not currently supported
+   */
+  public getServerByModel(model: ModelInfo): InferenceServer | undefined {
+    // check if model backend is supported
+    const backend: InferenceType = getInferenceType([model]);
+    const providers: InferenceProvider[] = this.inferenceProviderRegistry
+        .getByType(backend)
+        .filter(provider => provider.enabled());
+    if (providers.length === 0) {
+      throw new Error('no enabled provider could be found.');
+    }
+    return this.getServers().find(s => s.models.some(m => m.id === model.id));
+  }
+
+  /**
    * Creating an inference server can be heavy task (pulling image, uploading model to WSL etc.)
    * The frontend cannot wait endlessly, therefore we provide a method returning a tracking identifier
    * that can be used to fetch the tasks
--- a/packages/backend/src/managers/application/applicationManager.spec.ts
+++ b/packages/backend/src/managers/application/applicationManager.spec.ts
@@ -126,7 +126,7 @@ beforeEach(() => {
   vi.resetAllMocks();
 
   vi.mocked(webviewMock.postMessage).mockResolvedValue(true);
-  vi.mocked(recipeManager.buildRecipe).mockResolvedValue([recipeImageInfoMock]);
+  vi.mocked(recipeManager.buildRecipe).mockResolvedValue({ images: [recipeImageInfoMock] });
   vi.mocked(podManager.createPod).mockResolvedValue({ engineId: 'test-engine-id', Id: 'test-pod-id' });
   vi.mocked(podManager.getPod).mockResolvedValue({ engineId: 'test-engine-id', Id: 'test-pod-id' } as PodInfo);
   vi.mocked(podManager.getPodsWithLabels).mockResolvedValue([]);
@@ -312,7 +312,7 @@ describe('pullApplication', () => {
       'model-id': remoteModelMock.id,
     });
     // build the recipe
-    expect(recipeManager.buildRecipe).toHaveBeenCalledWith(connectionMock, recipeMock, {
+    expect(recipeManager.buildRecipe).toHaveBeenCalledWith(connectionMock, recipeMock, remoteModelMock, {
       'test-label': 'test-value',
       'recipe-id': recipeMock.id,
       'model-id': remoteModelMock.id,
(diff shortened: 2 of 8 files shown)


jeffmaury:
Tried but it seems it is still building the image and I cannot see the inference server being started. Seems the lookup is based on the image name but I would prefere to use the backend property as it can be used as an information to tell Podman AI Lab to look for an inference server so that recipes can be specified with:

- only backend => always look for inference server
- only images and build context => always used build image
- both => first try inference server then build image

lstocchi:
> Tried but it seems it is still building the image and I cannot see the inference server being started. 

Weird. Which recipe did you use? 

> Seems the lookup is based on the image name but I would prefere to use the backend property as it can be used as an information to tell Podman AI Lab to look for an inference server so that recipes can be specified with:
> 
> * only backend => always look for inference server
> * only images and build context => always used build image
> * both => first try inference server then build image

Sounds good, gonna update it

lstocchi:
@jeffmaury updated to use the backend as you said. Give it a look, thanks!!

jeffmaury:
> @jeffmaury updated to use the backend as you said. Give it a look, thanks!!

LGTM

packages/backend/src/assets/inference-images.json
```diff
@@ -5,6 +5,6 @@
   "llamacpp": {
     "default": "ghcr.io/containers/llamacpp_python@sha256:b2402f52b757fb2a1cc2b3075f9fb1cb4082a81fbae37631499895619728fdd3",
     "cuda": "ghcr.io/containers/llamacpp_python_cuda@sha256:eda2f13b0fa5b14e0b05414156f48fbf1f4a6a50517169bec25b83dc66f18f20",
-    "vulkan": "quay.io/ai-lab/llamacpp-python-vulkan@sha256:7e7d15086ec33d0547125a0afb479d91246a67b6d0681ae17260ea7665c333a0"
+    "vulkan": "quay.io/ai-lab/llamacpp-python-vulkan@sha256:e061dc2ed25f0f04bb502f08fee2902174f1f05b8da035964195968b9d6790e3"
```
axel7083, line 8: Need rebase ?

axel7083:
So today we have a few classes to create a recipe `ApplicationManager`, `RecipeManager`, `BuilderManager`, `PodManager`. Here is a diagram showing how thoses actors interact between them.

The idea by splitting with those small actors was to have some clear separation between the building part, and the live resources.

The RecipeManager (and BuilderManager) should only be responsible of static resources, `ai-lab.yaml` files, repository and container images. Where the ApplicationManager would have the role to make those resources live using the `PodManager` and `containerEngine`.  

Here is a diagram of how it is today (a bit simplified)

```mermaid
sequenceDiagram
    actor User
    participant ApplicationManager
    participant ContainerEngine
    participant PodManager
    autonumber
    User ->> ApplicationManager: Start Recipe
    rect rgb(0, 213, 255)
        Note over ApplicationManager,RecipeManager: Building part
        ApplicationManager ->> RecipeManager: Clone Recipe
        ApplicationManager ->> RecipeManager: Build Recipe
        RecipeManager ->> ApplicationManager: return RecipeImage[]
    end
    rect rgb(200, 150, 255)
        Note over ApplicationManager,PodManager: Instantiating part
        ApplicationManager ->> PodManager: Create pod
        loop Each RecipeImage
            ApplicationManager->>ContainerEngine: Create container
        end
    end
```

From my understanding of the code and the discussion above, we would like to replace the container(s) with a backend property by an InferenceServer.

IMO the `RecipeManager` should not have access to the lives resources, it should not have the InferenceServer in its constructor: When building the recipe, instead of returning `RecipeImage[]` it should probably return an interface like:

```ts
interface RecipeBuildResult {
  images: RecipeImage[];
  inferences: InferenceServerConfig[];
}
```

The idea here, would have the RecipeManager giving all the configuration required to build the recipe from scratch. Then, the `ApplicationManager` would have the responsibility to either start the inference server or re-use an existing one. 

IMO we should keep as simple as possible the RecipeManager, to delegate the complicated part of creating/dealing the inference server to the application manager. 

What do you think @lstocchi @jeffmaury ?

packages/backend/src/assets/inference-images.json
jeffmaury: Will do

packages/backend/src/managers/inference/inferenceManager.ts
```diff
@@ -101,6 +101,22 @@ export class InferenceManager extends Publisher<InferenceServer[]> implements Di
     return this.#servers.get(containerId);
   }
 
+  /**
+   * return the first inference server which is using the specific model
+   * it throws if the model backend is not currently supported
+   */
+  public getServerByModel(model: ModelInfo): InferenceServer | undefined {
```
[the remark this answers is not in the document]
jeffmaury: Will apply

jeffmaury:
> So today we have a few classes to create a recipe `ApplicationManager`, `RecipeManager`, `BuilderManager`, `PodManager`. Here is a diagram showing how thoses actors interact between them.
> 
> The idea by splitting with those small actors was to have some clear separation between the building part, and the live resources.
> 
> The RecipeManager (and BuilderManager) should only be responsible of static resources, `ai-lab.yaml` files, repository and container images. Where the ApplicationManager would have the role to make those resources live using the `PodManager` and `containerEngine`.
> 
> Here is a diagram of how it is today (a bit simplified)
> 
> ```mermaid
> sequenceDiagram
>     actor User
>     participant ApplicationManager
>     participant ContainerEngine
>     participant PodManager
>     autonumber
>     User ->> ApplicationManager: Start Recipe
>     rect rgb(0, 213, 255)
>         Note over ApplicationManager,RecipeManager: Building part
>         ApplicationManager ->> RecipeManager: Clone Recipe
>         ApplicationManager ->> RecipeManager: Build Recipe
>         RecipeManager ->> ApplicationManager: return RecipeImage[]
>     end
>     rect rgb(200, 150, 255)
>         Note over ApplicationManager,PodManager: Instantiating part
>         ApplicationManager ->> PodManager: Create pod
>         loop Each RecipeImage
>             ApplicationManager->>ContainerEngine: Create container
>         end
>     end
> ```
> 
> From my understanding of the code and the discussion above, we would like to replace the container(s) with a backend property by an InferenceServer.
> 
> IMO the `RecipeManager` should not have access to the lives resources, it should not have the InferenceServer in its constructor: When building the recipe, instead of returning `RecipeImage[]` it should probably return an interface like:
> 
> ```ts
> interface RecipeBuildResult {
>   images: RecipeImage[];
>   inferences: Partial<InferenceServerConfig>[];
> }
> ```
> 
> Note
> 
> Depending on the configuration, the RecipeManager would skip the containers with a backend and return an InferenceServerConfig instead of a RecipeImage
> 
> The idea here, would have the RecipeManager giving all the configuration required to build the recipe from scratch. Then, the `ApplicationManager` would have the responsibility to either start the inference server or re-use an existing one.
> 
> IMO we should keep as simple as possible the RecipeManager, to delegate the complicated part of creating/dealing the inference server to the application manager.
> 
> What do you think @lstocchi @jeffmaury ?
> ## Example
> 
> Given the following `ai-lab.yaml`
> 
> ```yaml
> version: v1.0
> application:
>   type: language
>   name: ChatBot_Streamlit
>   description: Chat with a model service in a web frontend.
>   containers:
>     - name: llamacpp-server
>       contextdir: ../../../model_servers/llamacpp_python
>       containerfile: ./base/Containerfile
>       model-service: true
>       backend:
>         - llama
>       arch:
>         - arm64
>         - amd64
>       ports:
>         - 8001
>       image: quay.io/ai-lab/llamacpp_python:latest
>     - name: streamlit-chat-app
>       contextdir: app
>       containerfile: Containerfile
>       arch:
>         - arm64
>         - amd64
>       ports:
>         - 8501
>       image: quay.io/ai-lab/chatbot:latest
> ```
> 
> The `RecipeManager.buildRecipe` would return something like
> 
> ```ts
> {
>   'images': [{
>     appName: 'streamlit-chat-app',
>     modelService: false,
>     ports: [8501],
>   }],
>   'inferences': [{
>     port: 8001,
>     inferenceProvider: 'llama-cpp',
>   }],
> }
> ```

Such refactoring should be part of a follow up PR as we need to put this into motion ASAP for the release.
The only issue I have for the redesign is that to decide which images to build you need to get access to the live data

packages/backend/src/assets/inference-images.json
axel7083, line 8: Still need to rebase ?

jeffmaury: Done

axel7083 (approved):
Testing the recipes

## `audio_to_text`

image

image

Working great

## `Code Generation`

image

image

Working great too

## `Chat bot` and `Summarizer` at the same time

image

They properly share the same inference server :heavy_check_mark: 

Approving with https://github.com/containers/podman-desktop-extension-ai-lab/issues/1542 as follow up discussion

Issue #1501, opened 2024-08-07
Do not store AI model in the test artifacts when run on GHA CI


odockal:
### Is your enhancement related to a problem? Please describe

Test artifacts now could have > 4GB. 

### Describe the solution you'd like

Do not save the model in the test artifacts.

PR #1505, opened 2024-08-07
chore(test): exclude models storage from CI artifacts

ScrewTSW:
### What does this PR do?
* excludes models storage from CI artifacts

### Screenshot / video of UI
N/A

#1501 

### How to test this PR?
N/A


axel7083:
Not related to this PR: would it possible to cache the models ? so we do not need to re-download them everytime the e2e run ?

axel7083 (approved):

Issue #1520, opened 2024-08-09
`PodmanConnection` should extend `Publisher` to expose the ContainerProviderConnection available


axel7083:
### Is your enhancement related to a problem? Please describe

In https://github.com/containers/podman-desktop-extension-ai-lab/issues/1462 we are showing the need to have a way to let the user choose which machine/connection to use when starting an inference server or a recipe.

### Describe the solution you'd like

The current `PodmanConnection` class handle the connection logic, but currently those are kept in the backend, it should extend the `Publisher` class and expose the available providers.

PR #1521, opened 2024-08-09
feat(PodmanConnection): makes class extend Publisher

axel7083:
### What does this PR do?

Makes the `PodmanConnection` extend the Publisher class to exposes the ContainerProviderConnections.

### Notable change

- Refactoring the `PodmanConnection` to keep track of the Connections available
- Adding onDidUnregisterContainerConnection listener 
- Adding onDidUpdateProvider listener
- Replacing manual type listener by `Event` and `EventEmitter`

### Screenshot / video of UI

https://github.com/user-attachments/assets/c151c8c7-2ac5-4281-9a4b-cbc988088f31

We are still not getting the events when the user is stopping the machine from the terminal.

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1520

### How to test this PR?

- [x] unit tests has been added (a lot)

### DIFF
--- a/packages/backend/src/managers/podmanConnection.ts
+++ b/packages/backend/src/managers/podmanConnection.ts
@@ -16,15 +16,15 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { process, EventEmitter, provider } from '@podman-desktop/api';
 import type {
+  ContainerProviderConnection,
   Disposable,
   Event,
   RegisterContainerConnectionEvent,
   UpdateContainerConnectionEvent,
   Webview,
-  ContainerProviderConnection,
 } from '@podman-desktop/api';
+import { EventEmitter, process, provider } from '@podman-desktop/api';
 import type { MachineJSON } from '../utils/podman';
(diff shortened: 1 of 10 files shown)


packages/backend/src/managers/podmanConnection.ts
```diff
@@ -16,94 +16,163 @@
...
+        return VMType.QEMU;
+      case VMType.LIBKRUN:
+        return VMType.LIBKRUN;
+      // windows
+      case VMType.HYPERV:
+        return VMType.HYPERV;
+      case VMType.WSL:
+        return VMType.WSL;
+      default:
+        return VMType.UNKNOWN;
```
benoitf, line 168: here it looks we're doing a lot of stuff

maybe simpler stuff like that will take into account new properties added to the VMType enum later
```ts
  const type = VMType[vmtype.toUpperCase()];
  if (type === undefined) {
    return VMType. UNKNOWN;
  }
return type
```

axel7083: I will give it a try thanks for the shortening

Issue #1462, opened 2024-07-31
Cannot start inference server if there are more than 1 Podman machine


jeffmaury:
### Bug description

The start is rejected:

![image](https://github.com/user-attachments/assets/56780c87-4904-4fcb-ab55-9b28f749fe90)

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. Have 2 podman machines (only one needs to be started)
2. Start an inference server

axel7083:
We need to provide the frontend the list of providers, and add a select input listing them if there are more than one.

axel7083:
List of requirements:

- https://github.com/containers/podman-desktop-extension-ai-lab/issues/1518

slemeur:
Can we handle the situations where we know there will be "only" one podman machine running and managing it on behalf of the useR?

axel7083:
> Can we handle the situations where we know there will be "only" one podman machine running and managing it on behalf of the useR?

With https://github.com/containers/podman-desktop-extension-ai-lab/pull/1537 we do not display the select input if we only have one machine, we simply use it. 

The input will only be visible if we have multiple

PR #1535, opened 2024-08-13
feature: WSL upload support podman connections

axel7083:
### What does this PR do?

When creating an Inference Server or an Application, on windows we upload the model to the WSL Podman Machine, however we uses the first available connection, which is problematic if multiple connections are available and we want to deploy on a specific machine.

### Notable change

- The `WSLUpload` worker takes a `ContainerProviderConnection` as argument now instead of using the first available machine.

### Screenshot / video of UI

image

When deleting a model, it will try to delete it on every podman-machine (may take a few seconds)

https://github.com/user-attachments/assets/8f9a91e3-fbff-4037-a5ec-c739cdd15680

### What issues does this PR fix or reference?

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1534

### How to test this PR?

- [x] Unit tests has been provided

We need to ensure it works fine on Windows (I can't I don't have a windows machine atm)

 ### Regression testing

- Create two podman machine WSL (`Machine 1` and `Machine 2`)
- Start `Machine 1`
- Start an inference server and start a recipe
- assert the model is uploaded on the running podman machine 
- Stop `Machine 1` and Start `Machine 2` (using podman-desktop UI not terminal)
- Start an inference server and start a recipe
- Start `Machine 1`
- assert all recipes / inferences are visible 
- Delete all recipes and inferences
- Delete the models
- assert all machine have the models delete 

### Advanced testing

- Start `Machine 1` and `Machine 2`
- Open AI Lab dev tools

```ts
// Get all connections
const connections = await studioClient.getContainerProviderConnection();
// Get a llama-cpp model
const model = (await studioClient.getModelsInfo()).find(model => model.file && model.backend === 'llama-cpp')
// start an inference server on Machine 1
studioClient.requestCreateInferenceServer({
    port: 8888,
    modelsInfo: [model],
    connection: connections[0]
});
// Start an inference server on Machine 2
studioClient.requestCreateInferenceServer({
    port: 9999,
    modelsInfo: [model],
    connection: connections[1]
});
```

PR #1537, opened 2024-08-13
feat: adding container provider connection select

axel7083:
### What does this PR do?

Allowing to select the connection the user want to use (if multiple exists) to deploy the Inference Server or Application.

Design is inspired from the `Pull Image` page of Podman Desktop

image

### Screenshot / video of UI

| Single connection | Multiple **started** connections |
| --- | ---- |
| image | image |
| image | image |

**Demo**

https://github.com/user-attachments/assets/916d0533-7f33-4270-97f6-e6461c742006

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1462

> [!CAUTION]
> Requires https://github.com/containers/podman-desktop-extension-ai-lab/pull/1535 for windows

### How to test this PR?

**How to test**

- Start several podman machines
- Go to the  `Create Service` page
- assert all **started** machines are visible
- Select the target machine and create 
- assert `pdoman --connection=<machine-name> container ls` is listed

### DIFF
--- a/packages/frontend/src/lib/select/ContainerProviderConnectionSelect.spec.ts
+++ b/packages/frontend/src/lib/select/ContainerProviderConnectionSelect.spec.ts
@@ -1,0 +1,75 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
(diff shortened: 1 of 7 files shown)


slemeur:
I think, Model and Port should be first as they are the primary things user will need to interact with.

axel7083:
> I don't understand this this could be hidden to the user

This is not hidden in the Podman Pulling page, why should it be hidden ? If user configure a podman remote with more capabilities, it should be able to decide where the resources should be deployed ?

If we only have one, like most cases we do not show this option, it only appear when multiple connections are available.

jeffmaury:
> > I don't understand this this could be hidden to the user
> 
> This is not hidden in the Podman Pulling page, why should it be hidden ? If the user configure a podman remote with more capabilities, it should be able to decide where the resources (recipe / inference server) should be deployed ?
> 
> If we only have one, like most cases we do not show this option, it only appear when multiple connections are available.

Remote would be a use case but I think we should focus on AI app dev features rather that low level technical infra. For example I believe selecting the proper connection based on footprint and load is adding more value

axel7083:
Rebasing after https://github.com/containers/podman-desktop-extension-ai-lab/pull/1535 has been merged

jeffmaury (approved):
LGTM. Just a side note the VMType is displayed as it is where Podman Desktop uses a mapping function (which by the way changes the VMType). We probably need to be aligned

PR #1544, opened 2024-08-14
feat: start REST server

### LINKED ISSUE #1295: Provide infrastructure for starting a REST server
Setttings should have an entry for the port number (default 10434 ?). If port is busy when starting the extension, allocate a random port and update the settings (so that external CLIs have a way to get the port number)

jeffmaury:
### What does this PR do?

1. Starts an API server on port 10434 (or a random one if this port is in use).

Serves two endpoints:

- `GET /spec`: returns the spec in OpenAPI YAML format 
- `GET /api/version`: returns the version of AI Lab

### Validation

Any endpoints (below /api for now - related to discussions points 1 & 2) are validated:
- `curl http://localhost:10434/api/not_found` should return an error
- `curl http://localhost:10434/api/version?bad_param` should return an error  
- modify the response in `getVersion` to add a new property to the returned object, or to rename the `version` property, `curl http://localhost:10434/api/version` should then display logs (at error level) that the response is incorrect 

### To discuss

- Do we want to serve the spec at `/spec` or at `/api/spec`?
- Do we want this endpoint to be documented in the OpenAPI Spec ?

Fixes #1295

### How to test this PR?

- install the extension
  - with image `quay.io/jeffmaury/ai-lab:rest-server`
  - or, by building the image:
    ```
    $ yarn
    $ yarn build
    $ podman build . -t your_image
    $ podman push your_image
    ```
- start Podman Desktop with the AI Lab extension
- check in the status bar that a message appears indicating that AI Lab is listening on port
- click the status
- copy the URL from the message box
- access the different endpoints from this URL


packages/backend/package.json
```diff
@@ -60,22 +60,26 @@
...
+    "@huggingface/gguf": "^0.1.7",
     "@podman-desktop/api": "1.12.0",
+    "@types/express": "^4.17.21",
     "@types/js-yaml": "^4.0.9",
     "@types/mustache": "^4.2.5",
     "@types/node": "^20",
     "@types/postman-collection": "^3.5.10",
-    "vitest": "^2.0.5"
+    "@types/supertest": "^6.0.2",
+    "isomorphic-git": "^1.27.1",
```
feloy, line 74: You may want to move these dependencies to the `dependencies` list

packages/backend/vite.config.js
```diff
@@ -35,6 +35,7 @@ const config = {
       '/@gen/': join(PACKAGE_ROOT, 'src-generated') + '/',
       '@shared/': join(PACKAGE_ROOT, '../shared') + '/',
     },
+    mainFields: ['module', 'jsnext:main', 'jsnext'], //https://github.com/vitejs/vite/issues/16444
```
feloy, line 38: 🚀

feloy:
Tested in dev mode (`yarn watch --extension-folder ...`) and in prod mode (by installing the extension with the image), works in both modes. 

@jeffmaury, you have removed the OpenApi validator (which still causes problem). Do we want to completely ignore the errors, or implement the validation in some other way?

jeffmaury:
> Tested in dev mode (`yarn watch --extension-folder ...`) and in prod mode (by installing the extension with the image), works in both modes.
> 
> @jeffmaury, you have removed the OpenApi validator (which still causes problem). Do we want to completely ignore the errors, or implement the validation in some other way?

Yes I think for the time being we can do manual validation ie reject if required parameter is missing

feloy (approved):
LGTM (at least the configuration changes to bundle express correctly)

packages/backend/src/managers/apiServer.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  getPackageFile(): string {
+    // when plugin is installed, the file is placed in the plugin directory (~/.local/share/containers/podman-desktop/plugins/<pluginname>/)
+    const prodFile = path.join(__dirname, '../package.json');
+    if (existsSync(prodFile)) {
+      return prodFile;
+    }
+    // return dev file
+    return path.join(__dirname, '../../../package.json');
+  }
```
benoitf, line 111: looks like it should be externalized as it's a copy/paste

jeffmaury: Should be good now

benoitf (approved):

PR #1549, opened 2024-08-15
fix: wrong VMtype returned by Podman Desktop causes inference servers…

jeffmaury:
… start failure

### What does this PR do?

Fix error when starting inference server on MacOS

### Screenshot / video of UI

N/A

### What issues does this PR fix or reference?

#1548

### How to test this PR?

Start an inference server or recipe on MacOS

PR #1550, opened 2024-08-15
fix: wrong VMtype returned by Podman Desktop causes inference servers…

jeffmaury:
…… (#1549)

* fix: wrong VMtype returned by Podman Desktop causes inference servers start failure

* fix: fix failing unit tests

---------


benoitf:
How will it work on previous podman desktop versions ?

PR #1570, opened 2024-08-20
feat: make api port configurable

### LINKED ISSUE #1426: Make REST API port configurable

#1391 makes the REST API listen on harcoded port 10434, or a random port if this one is not available.

The port should be configurable by the user. Optionally, when the configured port is in use, the choice may be proposed to the user to change the port with a random free one (through an error msg / dialog in the status bar).

_No response_

### Additional context

_No response_

feloy:
### What does this PR do?

Make tha API port configurable 
 
Fixes #1426 

### How to test this PR?

- Go to Preferences, change the API port, restart the extension and check the API is listening on this port
- Set a port in use, and check that an error is displayed in the status bar

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -24,6 +24,13 @@
           "type": "boolean",
           "default": false,
           "description": "Experimental GPU support for inference servers"
+        },
+        "ai-lab.apiPort": {
+          "type": "number",
+          "default": 10434,
+          "minimum": 1024,
+          "maximum": 65535,
+          "description": "Port on which the API is listening (requires restart of extension)"
         }
       }
     },
--- a/packages/backend/src/managers/apiServer.spec.ts
+++ b/packages/backend/src/managers/apiServer.spec.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 
 import { afterEach, beforeEach, expect, test, vi } from 'vitest';
-import { ApiServer } from './apiServer';
+import { ApiServer, PREFERENCE_RANDOM_PORT } from './apiServer';
 import request from 'supertest';
 import type * as podmanDesktopApi from '@podman-desktop/api';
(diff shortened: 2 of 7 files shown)


packages/backend/package.json
```diff
@@ -24,6 +24,13 @@
           "type": "boolean",
           "default": false,
           "description": "Experimental GPU support for inference servers"
+        },
+        "ai-lab.apiPort": {
+          "type": "number",
+          "default": 10434,
+          "minimum": 1,
```
benoitf, line 31: I think the minimum is > 1024 (root/admin needed ?)

benoitf (approved):

jeffmaury (commented):
How can we have 0 (random) if the minimum is 1024 ?

feloy:
> How can we have 0 (random) if the minimum is 1024 ?

The minimum value is only enforced in the Configuration partmodule, not in the ApiServer class, and so the zero value can be set by the tests when the Configuration module is mocked.

I would be happy to have the zero value settable by the user (so the user can choose to have the API listen on a random port) but there are 2 restrictions:
- there is no way to define ranges ([0] and [1024..65535] in this case) for numeric values
- the zero value is not saved correctly when set, it is overwritten by the default value (https://github.com/containers/podman-desktop/issues/8523)

Another possibility would be to have 2 preferences, but I'm not sure it is worth adding another preference:
- one boolean for setting the port as random
- another one, visible only if not random, to define the port

jeffmaury (commented):
Shouldn't we restart the REST server if the port changes ?

feloy:
> Shouldn't we restart the REST server if the port changes ?

As an event will be sent for every click on -/+ (with a 500ms throttle), this would restart the server frequently in intermediary ports before to start it in the correct port

PR #1579, opened 2024-08-21
feat: validate API requests and responses according to the OpenApi spec

### LINKED ISSUE #1578: Validate API requests and responses

feloy:
### What does this PR do?

Validate API requests and responses according to the OpenApi spec

Using the `express-openapi-validator` needs to patch the `ono` dependency with  https://github.com/JS-DevTools/ono/pull/20

We use the `@rollup/replace` plugin for this (another solution would be to use `yarn patch` to patch the `node_modules` directly, but the project is currently using yarn v1.22, which does not include this command, available only starting from v3)

Fixes #1578

### DIFF
--- a/packages/backend/vite.config.js
+++ b/packages/backend/vite.config.js
@@ -18,6 +18,7 @@
 
 import {join} from 'path';
 import {builtinModules} from 'module';
+import replace from '@rollup/plugin-replace';
 
 const PACKAGE_ROOT = __dirname;
 
@@ -59,6 +60,18 @@ const config = {
     emptyOutDir: true,
     reportCompressedSize: false,
   },
+  plugins: [
+    // This is to apply the patch https://github.com/JS-DevTools/ono/pull/20
+    // can be removed when the patch is merged
+    replace({
+      delimiters: ['', ''],
+      preventAssignment: true,
+      values: {
+        'if (typeof module === "object" && typeof module.exports === "object") {':
+          'if (typeof module === "object" && typeof module.exports === "object" && typeof module.exports.default === "object") {',
+      },
+    }),
+  ],
 };
 
 export default config;
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -71,6 +71,7 @@
   "dependencies": {
     "@huggingface/gguf": "^0.1.9",
     "express": "^4.19.2",
+    "express-openapi-validator": "^5.3.1",
     "isomorphic-git": "^1.27.1",
     "mustache": "^4.2.0",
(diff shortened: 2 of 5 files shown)


packages/backend/vite.config.js
```diff
@@ -34,6 +34,8 @@ const config = {
       '/@/': join(PACKAGE_ROOT, 'src') + '/',
       '/@gen/': join(PACKAGE_ROOT, 'src-generated') + '/',
       '@shared/': join(PACKAGE_ROOT, '../shared') + '/',
+      '@jsdevtools/ono': '@jsdevtools/ono/cjs/index.js',
+      'ono': '@jsdevtools/ono/cjs/index.js',
```
benoitf, line 38: curious why we need that ? these one are usually just for the local source code new links

feloy: Ono provides both versions, cjs and esm, and vite includes the esm one by default.

But the esm version includes this code, which fails because `module.exports.default` is not defined:
```
// CommonJS default export hack
if (typeof module === "object" && typeof module.exports === "object") {
    module.exports = Object.assign(module.exports.default, module.exports);
}
```

benoitf: shouldn't the extension be set as a module ?

type: "module" ?

I did that for the podman extension
https://github.com/containers/podman-desktop/blob/main/extensions/podman/package.json#L7

and we say we generate a cjs file
https://github.com/containers/podman-desktop/blob/main/extensions/podman/package.json#L13

benoitf: the rationale is that we have more and more libraries that are only esm
so instead of trying to grab the commonjs files, we should switch to be esm and allow to use esm easily as well

some libraries are now esm only
https://github.com/containers/podman-desktop/commit/8e6debfa873a9a55441a3bfb65e87a3652b26ce2

feloy: @benoitf I just tested with `type: "module"` (and removing the aliases), but the error is the same. The hack in the esm version of ono seems not compatible with `vite`

benoitf: I think that we can't use cjs as for long term it'll be a burden

we may patch the module using https://github.com/JS-DevTools/ono/pull/20 or getting rid of the lines ?

or switch to another library

feloy: OK, I'll make changes to patch ono's esm version, so we can use it. Thanks @benoitf

benoitf (approved):

gastoner (approved):

Issue #1611, opened 2024-08-28
Container Connection banner should depend on the selected container engine


axel7083:
### Is your enhancement related to a problem? Please describe

Currently the warning banner is using the first available, which could be improved to depend on the selected one

![image](https://github.com/user-attachments/assets/a91498fb-c4da-4a85-b0cf-242a676d0335)

### Describe the solution you'd like

Following https://github.com/containers/podman-desktop-extension-ai-lab/pull/1537 we should use the selected value

PR #1612, opened 2024-08-28
feat: container resource banner depends on selected connection

axel7083:
### What does this PR do?

Makes the resource warning depends on the connection selected. 

### Screenshot / video of UI

https://github.com/user-attachments/assets/b048fd19-015b-4ece-8423-d0977f433989

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1611

### How to test this PR?

**Manually**

- Create a podman machine with little memory (not possible on windows) and one with a lot of memory. 
> `podman machine init --memory=1024 machine-low-memory --now` 
- assert both machines are running
- Go to AI Lab 
- Navigate to `Creating Model service`
- assert both machine are visible
- Select the low memory machine
- assert warning banner is visible
- Select the high memory machine
- assert warning banner is **not** visible


packages/shared/src/models/IContainerConnectionInfo.ts
```diff
@@ -27,11 +27,12 @@ export interface ContainerProviderConnectionInfo {
...
 
 export interface CheckContainerConnectionResourcesOptions {
-  modelInfo: ModelCheckerInfo;
+  model: ModelInfo & { memory: number };
+  context: ModelCheckerContext;
   connection?: ContainerProviderConnectionInfo;
 }
 
-export type ContainerConnectionInfo =
+export type ContainerConnectionResourceInfo =
```
lstocchi, line 35: i don't get why it was renamed as only one of the subtypes directly refers to resources.  So in most cases there are no info about resources

lstocchi (approved):
LGTM, only doubts about the renaming.

jeffmaury (approved):
LGTM

But the included refactoring (renaming) make it harder to review and should be separate

packages/shared/src/models/IContainerConnectionInfo.ts
axel7083: I revert the rename

Issue #1607, opened 2024-08-28
switch to eslint v9.9.1


jeffmaury:
### Is your enhancement related to a problem? Please describe

Updates are starting to fail

### Describe the solution you'd like

Update configuration to eslint v9

PR #1619, opened 2024-08-28
chore: parse svelte files when doing the linting

benoitf:
### What does this PR do?
While wanting to update eslint dependency, I noticed that svelte files are not analyzed

so before doing the library update, include svelte files during the linting process

files are not compliant so fix the automatic fixes and exclude files from some rules to be able to continue

https://github.com/containers/podman-desktop-extension-ai-lab/issues/1607

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ __mocks__
 coverage
 packages/backend/media/**
 src-generated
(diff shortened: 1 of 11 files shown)


jeffmaury (approved):
LGTM

PR #1623, opened 2024-08-28
chore: upgrade to eslint v9

benoitf:
### What does this PR do?
update to the new flat config format of eslint

need to disable 2 rules from sonarjs lint plug-in as it's not yet fully compliant

some code is not 100% compliant with the rules so I've commented the rules, but then after it should be fixed rule by rule

search for `// disabled as code in this project is not yet compliant:` in the source code

Also the cache is enabled for `lint:check` / `lint:fix` rules

### How to test this PR?

PR check should be ✅


jeffmaury (approved):
LGTM

axel7083 (approved):

PR #1626, opened 2024-08-29
feat: implement pull model for API

### LINKED ISSUE #1583: Provide REST endpoint for downloading/pulling model

feloy:
### What does this PR do?

Adds a `/api/pull` endpoint to the API to download models

### Screenshot / video of UI

https://github.com/user-attachments/assets/bcc8fb65-cec6-4c79-b85d-6b55875d9f83

Fixes #1583 

### How to test this PR?

Start AI Lab extension

```
$ OLLAMA_HOST=localhost:10434 ollama pull facebook/detr-resnet-101
// should download model
$ OLLAMA_HOST=localhost:10434 ollama pull facebook/detr-resnet-101
// should indicate success and not reload model
$ OLLAMA_HOST=localhost:10434 ollama pull unknown-model
// should fail with error message
```

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -51,6 +51,30 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ListResponse'
+  /api/pull:
+    post:
+      operationId: pullModel
+      tags:
+        - models
+      description: | 
+        Download a model from the Podman AI Lab catalog.
+      summary: |
+        Download a model from the Podman AI Lab Catalog.
+      requestBody:
+        required: true
+        description: Request to pull a model
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PullRequest'
+      responses:
+        '200':
+          description: Model was successfully pulled
+          content:
+            application/x-ndjson:
+              schema:
+                $ref: '#/components/schemas/ProgressResponse'
+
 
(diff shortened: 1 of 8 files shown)


jeffmaury (commented):
Seems to me that we are always working in streaming mode ?

feloy:
> Seems to me that we are always working in streaming mode ?

You mean that all responses of the `/api/pull` are JSON streams in this implementation?

It also seems to be the case for ollama implementation, for example:

```
$ curl -X POST localhost:10343/api/pull -d '{"model": "unknown"}'
{"status":"pulling manifest"}
{"error":"pull model manifest: file does not exist"}
```

jeffmaury (approved):
LGTM

Only remark is that I found it strange that 200 is returned in case of error in non streaming mode

feloy:
> LGTM
> 
> Only remark is that I found it strange that 200 is returned in case of error in non streaming mode

Where? I can only see 500 codes when returning errors, in non-streaming mode

jeffmaury:
> > LGTM
> > Only remark is that I found it strange that 200 is returned in case of error in non streaming mode
> 
> Where? I can only see 500 codes when returning errors, in non-streaming mode

Sorry that was in streaming mode

gastoner (approved):
Works well

PR #1649, opened 2024-09-03
feat: display instructlab sessions, with mock instructlab manager

### LINKED ISSUE #1568: Implement InstructLab sessions screen

If there are session, display then as a table.

See the **Model Details Knowledges & Skills (0)** screen on https://design.penpot.app/#/workspace/971a4169-b56c-80d0-8002-4ff987ca8805/6e12068a-09a0-801a-8004-24b355541870?page-id=45387374-aad3-80a3-8004-c8e259891437&layout=layers

_No response_

### Additional context

_No response_

### LINKED ISSUE #1616: Provide a mock InstructLab manager and connect it to the InstructLab session screen

feloy:
### What does this PR do?

display instructlab sessions, with mock instructlab manager

### Screenshot / video of UI

https://github.com/user-attachments/assets/4e95c681-a8f8-4735-9473-d58d23848979

Fixes #1568
Fixes #1616

### How to test this PR?

You need to set an hidden setting, in `~/.local/share/containers/podman-desktop/configuration/settings.json`

(you must completely stop Podman Desktop - not only close the window -, edit the config file, then restart Podman Desktop for the change to be taken into consideration) 

```
  "ai-lab.experimentalTuning": true,
```


packages/backend/src/studio.ts
```diff
@@ -49,6 +49,7 @@ import { RecipeManager } from './managers/recipes/RecipeManager';
 import { GPUManager } from './managers/GPUManager';
```
axel7083, line 49: I think we should create a dedicated instrulab-api-impl.ts object, and also registering it in the same way we are doing for studio-api-impl. 

As we do not know yet the implication and complexity of instrutlab implementation, it is probably better to keep it appart.

feloy: You mean to also split the `SutdioAPI`, by having a separate `InstructlabAPI`?

axel7083: Yeah !

feloy: ok, I'll make the change

packages/backend/src/instructlab-api-impl.ts
```diff
@@ -0,0 +1,31 @@
+/**********************************************************************
```
axel7083, line 1: Thanks ! Great job

axel7083 (approved):
LGTM ! :rocket: nice job

jeffmaury (commented):
LGTM. 

Could we get navigation and tooltip from the model column to model details ?

feloy:
> LGTM.
> 
> Could we get navigation and tooltip from the model column to model details ?

That should be ok now

feloy:
> Looks better to me. I expected also the same navigation on target model but could be a follow up PR

As the target model will exist only when the instructlab session will be finished, I prefered not to add it for now.

Issue #1651, opened 2024-09-03
Download of ibm granite-8b-code-instruct-GGUF model fails


alxlenc:
### Bug description

The download  of  `ibm-granite/granite-8b-code-instruct-GGUF` fails.

![image](https://github.com/user-attachments/assets/f1e62414-ccbf-4bbc-bd62-2e13cf152693)

This might be caused by the extension not handling redirects.

```bash 
❯ curl https://huggingface.co/ibm-granite/granite-8b-code-instruct-GGUF/resolve/main/granite-8b-code-instruct.Q4_K_M.gguf
Temporary Redirect. Redirecting to /ibm-granite/granite-8b-code-instruct-4k-GGUF/resolve/main/granite-8b-code-instruct.Q4_K_M.gguf%
```

### Operating system

Fedora 40

### Installation Method

from Podman-Desktop extension page

### Version

1.0.0

### Steps to reproduce

Reopen Podman Desktop and click the download button for the `ibm-granite/granite-8b-code-instruct-GGUF` model.

### Additional context

Podman AI Lab extesion v1.2.3

axel7083:
When calling the url `https://huggingface.co/ibm-granite/granite-8b-code-instruct-GGUF/resolve/main/granite-8b-code-instruct.Q4_K_M.gguf` we are getting the following headers

```json
{
  'content-type': 'text/plain; charset=utf-8',
  'content-length': '130',
  connection: 'keep-alive',
  date: 'Tue, 03 Sep 2024 13:41:28 GMT',
  'x-powered-by': 'huggingface-moon',
  'cross-origin-opener-policy': 'same-origin',
  'referrer-policy': 'strict-origin-when-cross-origin',
  'x-request-id': 'Root=1-66d71208-22931bc4582cf0b917fac1b5',
  'access-control-allow-origin': 'https://huggingface.co',
  vary: 'Origin, Accept',
  'access-control-expose-headers': 'X-Repo-Commit,X-Request-Id,X-Error-Code,X-Error-Message,X-Total-Count,ETag,Link,Accept-Ranges,Content-Range',
  location: '/ibm-granite/granite-8b-code-instruct-4k-GGUF/resolve/main/granite-8b-code-instruct.Q4_K_M.gguf',
  'x-cache': 'Miss from cloudfront',
  via: '1.1 6bbb2da0f4f203dadcd8f0ae1073d674.cloudfront.net (CloudFront)',
[...]

PR #1655, opened 2024-09-03
fix: redirect url on model download

axel7083:
### What does this PR do?

When redirecting through the location header, website may provide only the path (E.g. `/hello`) when we expect a full url. 

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1651

### How to test this PR?

### DIFF
--- a/packages/backend/src/utils/downloader.ts
+++ b/packages/backend/src/utils/downloader.ts
@@ -94,11 +94,10 @@ export class Downloader {
    * @protected
    */
   protected getRedirect(location: string): string {
-    if(URL.canParse(location))
-      return location;
+    if (URL.canParse(location)) return location;
 
     const origin = new URL(this.url).origin;
-    if(URL.canParse(location, origin)) return new URL(location, origin).href;
+    if (URL.canParse(location, origin)) return new URL(location, origin).href;
 
     console.warn(`malformed location: cannot parse ${location}`);
     return location;
--- a/packages/backend/src/utils/downloader.ts
+++ b/packages/backend/src/utils/downloader.ts
@@ -87,6 +87,23 @@ export class Downloader {
     });
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


packages/backend/src/utils/downloader.ts
```diff
@@ -101,7 +117,9 @@ export class Downloader {
     https.get(url, { signal: this.abortSignal }, resp => {
       // Determine the total size
       if (resp.headers.location) {
-        this.followRedirects(resp.headers.location, callback);
+        const redirect = this.getRedirect(resp.headers.location);
+        console.warn(`follow redirect to ${redirect}`);
```
benoitf, line 121:
```suggestion
```

packages/backend/src/utils/downloader.ts
```diff
@@ -87,6 +87,22 @@ export class Downloader {
...
+   * @param location
+   * @protected
+   */
+  protected getRedirect(location: string): string {
+    if (URL.canParse(location)) return location;
+
+    const origin = new URL(this.url).origin;
+    if (URL.canParse(location, origin)) return new URL(location, origin).href;
+
+    console.warn(`malformed location: cannot parse ${location}`);
```
benoitf, line 102: I think that if the log trace needs to stay, it needs more context else a user won't understand if you see this trace being printed

jeffmaury (commented):
When the new URL is computed, it is computed from the URL that initiated the workflow but it should be computed from the URL that caused the redirection in case there are several redirections in the flow

axel7083:
> When the new URL is computed, it is computed from the URL that initiated the workflow but it should be computed from the URL that caused the redirection in case there are several redirections in the flow

Oh nice catch, thanks for noticing that :+1:

jeffmaury (approved):
LGTM

Issue #1741, opened 2024-09-18
Prototype integration of openweb-ui


jeffmaury:
### Is your enhancement related to a problem? Please describe

Seems openweb-ui can be integrated through a container, would be good to prototype

### Describe the solution you'd like

Replace current playground UI will openweb-ui

axel7083:
Shouldn't we have something like an abstract layer, and be able to support multiple playground ui ? Just like the inference servers

slemeur:
This is a prototyping issue - let's not over-engineer this.
Today, our playground UI is not abstracted to support multiple playground ui. We may need it in the future or may never need it. Right now, that's not "the problem" this issue should be focusing on.

PR #1753, opened 2024-09-20
feat: open-webui playground prototype

jeffmaury:
### What does this PR do?

POC for integration of open-webui as playground.

Known issues:
- model not selected by default
- locale using system need to find a way to enforce en
- theme
- better container management: we should monitor that port is available as it take some time to start

#1741


Firewall:
Looks awesome! 

A couple of questions: 
1. When you have the window with the UI inside of the Podman Desktop ui. Maybe I am missing it but would it be useful to have a button to "open in your browser" maybe next to the 🗑️ icon? 
2. Could we add a shortcut to allow users to go from the `Playground` page to the `Containers` page filtered down to the containers that are running the playground?

benoitf (commented):
is it plan to have for example an experimental flag to try/enable it ?

because here it looks it is a drop-in replacement

PR #1810, opened 2024-09-30
feat: improve podman cli execution

axel7083:
### What does this PR do?

Following https://github.com/containers/podman-desktop/pull/8891 ai-lab should uses podman's extension capabilities for execution, and stop using `getPodmanCli` utility method.

- Moved execution logic to `CodmanConnection` class with a `execute` and `executeSSH` method
- Create `protected executeLegacy` for podman-desktop < 1.13.

### What issues does this PR fix or reference?

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1809

### How to test this PR?

- [x] unit tests has been provided

PR #1813, opened 2024-09-30
feat: adding execute method to PodmanConnection class

axel7083:
### What does this PR do?

This PR adds `execute` and `executeSSH` method to the `PodmanConnection` class in AI Lab. The purpose is to centralize the code using the podman cli (E.g. model management, WSL Upload)

The new added code takes advantages of https://github.com/containers/podman-desktop/pull/8891, but still support through `Legacy` function previous version of podman-desktop (<1.13).

I split the feature in two
- Implementation (this PR)
- Usage (https://github.com/containers/podman-desktop-extension-ai-lab/pull/1810)

### Screenshot / video of UI

N/A

Required for https://github.com/containers/podman-desktop-extension-ai-lab/pull/1810

### How to test this PR?

### DIFF
--- a/packages/backend/src/managers/podmanConnection.ts
+++ b/packages/backend/src/managers/podmanConnection.ts
@@ -23,10 +23,12 @@ import type {
   RegisterContainerConnectionEvent,
   UpdateContainerConnectionEvent,
   Webview,
+  RunResult,
+  RunOptions,
+  ProviderContainerConnection,
 } from '@podman-desktop/api';
-import { containerEngine, env, navigation, EventEmitter, process, provider } from '@podman-desktop/api';
-import type { MachineJSON } from '../utils/podman';
-import { MIN_CPUS_VALUE, getPodmanCli } from '../utils/podman';
+import { containerEngine, env, navigation, EventEmitter, process, provider, extensions } from '@podman-desktop/api';
+import { getPodmanMachineName, type MachineJSON, MIN_CPUS_VALUE, getPodmanCli } from '../utils/podman';
 import { VMType } from '@shared/src/models/IPodman';
 import { Publisher } from '../utils/Publisher';
 import type {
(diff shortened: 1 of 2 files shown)


packages/backend/src/managers/podmanConnection.ts
```diff
@@ -54,6 +60,71 @@ export class PodmanConnection extends Publisher<ContainerProviderConnectionInfo[
...
+   * ```
+   * @param connection
+   * @param args
+   * @param options
+   */
+  executeSSH(connection: ContainerProviderConnection, args: string[], options?: RunOptions): Promise<RunResult> {
+    return this.execute(connection, ['machine', 'ssh', this.getNameLegacyCompatibility(connection), ...args], options);
+  }
+
+  /**
```
benoitf, line 106: I'm not sure we need the code for < 1.13 as the extension is already requiring API > 1.13.0

axel7083: We are not ?

https://github.com/containers/podman-desktop-extension-ai-lab/blob/35367499f94a8726bb36dbc3a36045cb4677defa/packages/backend/package.json#L11

We uses the latest api, but we don't have hard requirements on latest version, so we are _in theory_ compatible with older version up to 1.8.

benoitf: in the catalog it says minimum 1.12 for 1.2 of ai lab

https://github.com/containers/podman-desktop-catalog/blob/gh-pages/api/extensions.json#L689

and in package.json, we consume 1.13.0 snapshot API

I'm assuming some API changes (new additions) are also missing since 1.8

so consuming 1.13 API should result in 1.13 minimal requirement

benoitf (approved):
approving as the need for the version compatibility is a separate issue

gastoner (approved):

Issue #1837, opened 2024-10-03
Multiple object registered through the MessageProxy system can lead to conflict.


axel7083:
### Bug description

Podman Destkop offer a basic channel system to the extension to let them communicate between the backend and the frontend.

We build on top of it a message proxy system, offering us a simple way, and avoid having to define channel ids manually (we were using the method name.)

Example;

StudioAPI has a `ping` method

https://github.com/containers/podman-desktop-extension-ai-lab/blob/4d420cb5e32c7abe41827e86c5424ffa5044e8e5/packages/shared/src/StudioAPI.ts#L42

We use the method name as a channel, with some proxy handling, we forward properly.

This works nicely. Very nice and low maintenance, however there is a problem: conflicts.

If two instance are registered and share the same method name, there is conflict.

### Operating system

windows 11

### Installation Method

Other

### Version

next (development version)

PR #1838, opened 2024-10-03
fix(MessageProxy): channel conflict

axel7083:
### What does this PR do?

Every class that is registered to the MessageProxy system must have a static `CHANNEL` string property, which will be used to prefix every channel.

StudioAPI declared a function `ping` before the name of the method was used, which could lead to conflict with other registered method (InstructLabAPI), therefore we uses the CHANNEL as prefix, the channel become `StudioAPI-ping`.

Allowing scaling, and will avoid problem while working on instructlab api.

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1837

### How to test this PR?


jeffmaury (commented):
Can't we use the class name as channel ?

axel7083:
> Can't we use the class name as channel ?

Sadly no, the problem is the following

The frontend do not have access to the `StudioApiImpl` class (file on backend), it only has access to the `StudioAPI` abstract class (file on shared). Therefore on the frontend, we don't know the string `StudioApiImpl`.

benoitf:
IMHO I think there should be some time where brainstorming should occur between between when issues are opened and PR being opened

>> Can't we use the class name as channel ?
> @jeffmaury sadly no, the problem is the following
can we use the interface name instead ?

usually how it works on proxy is that you ask stuff for a given interface

client =
myProxy = Proxy(Interface)

server =
registerProxy(Interface, implementation)

so registering also two times the same interface name would conflict

I think the channel is an implementation details and should not be exposed

axel7083:
> IMHO I think there should be some time where brainstorming

Movind to draft

axel7083:
> I think the channel is an implementation details and should not be exposed

Fixed

benoitf (commented):
I still don't get why we're introducing a channel name


abstract classes have a name


so getProxy(StudioAPI) would work as is, no need to redeclare a constant having the same name than the abstract class

and signature would be something like `getProxy(proxyClass: {name: string})`

axel7083:
> and signature would be something like getProxy(proxyClass: {name: string})

Fixed

axel7083:
Moving to draft: abstract class do not seems to expose their abstract method through the `.prototype` property

axel7083:
After reading a bit about abstract class in Typescript, I found `There is no code generated for an abstract method`[^1]

Therefore, in the `MessageProxy` we are unabled at runtime to list the methods and create the corresponding channels.

https://github.com/containers/podman-desktop-extension-ai-lab/blob/5d6a3428ed534da6ddc7d2949dc921240984df19/packages/shared/src/messages/MessageProxy.ts#L108-L111

We are getting an empty array when providing `StudioAPI`, **only a non-abstract** class would allow us to get a non-empty array. This is why we need `StudioApiImpl`

We cannot transform the `StudioAPI` to an interface for technical reasons, interfaces do not exists at runtime in JS

image

This is the reasons why abstract class was used when implementing the channels system.

## Hardcoded `CHANNEL` property

I still believe my solution with the const CHANNEL is the right one. 

[^1]: https://stackoverflow.com/a/49487628

benoitf:
CHANNEL is in the abstract class, not per method and the `.name` property is on the abstract class as well

so I don't see the difference there as you need to provide the Abstract class both from client and server side

axel7083:
Sorry if I was not clear in the previous message.

As I show above

> image

We can clearly see at runtime, you cannot access to the list of methods from an abstract class compiled to javascript.

Now let's dig into how frontend and backends are communicating.


## Backend

The backends, register a class object, and using reflection, read the prototype of the object provided to list **all the methods available**.

Today (before this PR) we iterate over all methods and the backend register channels. Example `ping` channel is created from the ping method declared in `StudioApiImpl`.

https://github.com/containers/podman-desktop-extension-ai-lab/blob/ea02986c994f9eb60e587bcf7a35a309cde6c315/packages/shared/src/messages/MessageProxy.ts#L113-L116

Then when we received an event from the webview, we can directly call the right method.

https://github.com/containers/podman-desktop-extension-ai-lab/blob/ea02986c994f9eb60e587bcf7a35a309cde6c315/packages/shared/src/messages/MessageProxy.ts#L79-L85

## Frontend

The frontend, **do not create channels**, it create a proxy object, and when you call **anything** on the proxy, it will transform the method you are calling into string, and be used as channel.

If I do `studioClient.ping()` the proxy handler will received a property call, named `ping`. With this information the proxy handler will call the channel `ping` that the backend will receive and makes it corresponds.

https://github.com/containers/podman-desktop-extension-ai-lab/blob/ea02986c994f9eb60e587bcf7a35a309cde6c315/packages/shared/src/messages/MessageProxy.ts#L178-L181

## The `name` problem

The issue with the `name` property is that `StudioApiImpl.name === 'StudioApiImpl'` and `StudioApi.name === 'StudioApi'`.

Yes they both have a name property, but not the same.

Let's look at our options

```
registerInstance<T extends Record<keyof T, UnaryRPC>>(classType: new (...args: any[]) => T, instance: T): void
````

The signature of registerInstance (backend) needs the following arguments

- `classType` is used to list the methods (need to be a real class not abstract, so we cannot use StudioApi).
- `instance`: kept to call when needed 

We could simplified the first argument by not using it to generate the channel, and use the instance. By doing some hacky stuff we can get the prototype (methods name) from the instance `Object.getOwnPropertyNames(Object.getPrototypeOf(instance))`, and then we would provide the interface at first argument.

**Comment by axel7083** [comment]
New problem on the frontend side, we are minifiying the code, which result in class names not being consistent

image

The only possibility I see is:  Disabling the minify.

**Comment by benoitf** [comment]
let's defer the discussion about how proxy should like (no abstract class but interfaces, etc) out of this PR
and merge the current solution that fixes an issue

File: packages/shared/src/messages/MessageProxy.ts [hunk 1]
```diff
@@ -169,15 +169,16 @@ export class RpcBrowser {
     });
   }
 
-  getProxy<T>(): T {
+  getProxy<T>(classType: { name: string; prototype: T }): T {
+    console.log('get proxy', classType);
```
benoitf, line 173: to be removed I guess

packages/frontend/vite.config.js
```diff
@@ -59,6 +59,7 @@ export default defineConfig({
     },
   },
   build: {
+    minify: true,
```
benoitf, line 62: it would need to be introduced outside of the scope of this PR

axel7083: I am against removing the minify

packages/shared/src/messages/MessageProxy.ts
axel7083: Done

axel7083:
> merge the current solution that fixes an issue (with constant)

I revert the code to use the `CHANNEL` property

benoitf (approved):

Issue #1592, opened 2024-08-25
Adjust Inference Server details page UX


slemeur:
### Is your enhancement related to a problem? Please describe

While the inference server page is listing the information, those are not easy to decipher. And we would like to introduce more sections later on.

<img width="1379" alt="Screenshot 2024-08-25 at 15 23 59" src="https://github.com/user-attachments/assets/417d78d0-b3e4-4f92-9bbf-5c301482114f">

### Describe the solution you'd like

We would like the screen to be better guiding the developer onto the different component and information displayed.

- "Server" should be rename "Inference Server"
- The most important information is the "Inference Endpoint" but it is buried into the screen, hard to find the first time you use the tool. It should be first. We would like to put more emphasis on this one. Also, we would like to have a label "Inference Endpoint:" and it should be obvious that the user can easily copy it (maybe displaying the copy icon). Wondering, if we should not use a different color for that one.
- Similarily, the documentation for the swagger UI is hard to spot the first time and people may even not notice it. We should have a label "Inference API documentations"
- Finally, along with https://github.com/containers/podman-desktop-extension-ai-lab/issues/495, we could put the "GPU Inference" onto a dedicated section.

axel7083:
cc @ekidneyrh @MariaLeonova

MariaLeonova:
@axel7083 how much UX support is needed? Shall I create a penpot mockup for this screen enhancements?

axel7083:
> @axel7083 how much UX support is needed? Shall I create a penpot mockup for this screen enhancements?

I think yes that would help, as Stevan is talking about including some metrics in the page which would be a lot of changes https://github.com/containers/podman-desktop-extension-ai-lab/issues/495 and would probably need some discussion as this is currently something we do in podman-destkop (we uses the donuts for resources)

Issue #1691, opened 2024-09-11
Redesigned inference server details mockup


jeffmaury:
### Is your enhancement related to a problem? Please describe

See 

### Describe the solution you'd like

A mockup for the redesigned UI

MariaLeonova:
@jeffmaury @slemeur @axel7083 I have a couple questions to help me understand the issue better: 

1. is there any documentation for this UI and working with model services in AI Lab?
2.  how are these two different? Which is the Endpoint URL to be copied? ![Screenshot 2024-09-12 at 10 23 13](https://github.com/user-attachments/assets/c786f8a4-837b-43e6-be16-3bbb7ffc4b87)
3. what is the user's main goal on this page and what have they come to do? If it's the URL they want, what would they do with it next?
4. where is the Inference API documentation now?
  
I have highlighted the issues at hand on this screenshot, so I will address them when redesigning the UI:

![Screenshot 2024-09-12 at 10 26 19](https://github.com/user-attachments/assets/e104d7c3-ff0c-4c4f-9023-beec2323cc99)

jeffmaury:
> @jeffmaury @slemeur @axel7083 I have a couple questions to help me understand the issue better:
> 
>     1. is there any documentation for this UI and working with model services in AI Lab?

No only the README from the repo: https://github.com/containers/podman-desktop-extension-ai-lab?tab=readme-ov-file#usage
> 
[...]

PR #1840, opened 2024-10-04
feat: refresh Inference Server Details page

axel7083:
### What does this PR do?

Implement the @MariaLeonova mock-up for the Inference Server Details page from https://github.com/containers/podman-desktop-extension-ai-lab/issues/1691#issuecomment-2387868931.

### Diff with mock-ups

- Adding `arrow-up-right-from-square`[^1] to the Model as the link is clickable and it was not clear
- Removing the `name` property bellow the model, as it would be repetitive with the model name above
- Removing the `quantisation` as cannot get this information easily.

[^1]: https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid

### Screenshot / video of UI

**Server Running**

image

**Server Stopped**

image

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1592

### How to test this PR?

- [x] Existing unit tests ensure no regression

### DIFF
--- a/packages/frontend/src/pages/InferenceServerDetails.svelte
+++ b/packages/frontend/src/pages/InferenceServerDetails.svelte
@@ -231,6 +231,7 @@ function handleOnChange(): void {
                         <!-- API URL -->
                         {#if 'api' in service.labels}
                           <CopyButton
+                            top
                             class="bg-[var(--pd-label-bg)] text-[var(--pd-label-text)] rounded-md p-2 flex flex-row w-min h-min text-xs text-nowrap items-center"
                             content={service.labels['api']}>
                             {service.labels['api']}
--- a/packages/frontend/src/pages/InferenceServerDetails.svelte
+++ b/packages/frontend/src/pages/InferenceServerDetails.svelte
@@ -240,10 +240,10 @@ function handleOnChange(): void {
                         {/if}
 
                         <!-- Documentation URL -->
-                        <div class="grow">
+                        <div class="grow text-[var(--pd-label-text)]">
                           {#if 'docs' in service.labels}
+                            Access
                             <Tooltip tip="Open swagger documentation">
-                              Access
                               <Link
                                 aria-label="swagger documentation"
                                 on:click={openLink.bind(undefined, service.labels['docs'])}>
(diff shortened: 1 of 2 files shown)


MariaLeonova (commented):
Looks good to me!

benoitf (approved):

gastoner (commented):
When opening details of this model (one on the beginning)

image

I would also add at least some details about the model or don't offer the option to open the model if there are no details

I'm not sure of the tooltip should show under the element (I would imagine that the tooltip would show on right - over the `Access` which should reflect light/dark mode)



https://github.com/user-attachments/assets/faf066f8-5a93-455b-bce7-ff1bda9aa9aa


Otherwise LGTM

axel7083:
> I would also add at least some details about the model or don't offer the option to open the model if there are no details

Thanks for reporting that, however this is models specific, for each model we have a corresponding description (markdown), and it is displayed in here.

We had some project at some point to add `metadata` page to the models, but this has not been going further (https://github.com/containers/podman-desktop-extension-ai-lab/pull/1345).

I still think we should keep our system reactive, and allow redirection between the resources, even tho they might be a bit empty for some of them.

> I'm not sure of the tooltip should show under the element

Yeah I also agree, thanks for the feedback I will update!

MariaLeonova:
@axel7083 do you know why this particular model doesn't have any description? It seems like a big and important one - can we add something?

axel7083:
> @axel7083 do you know why this particular model doesn't have any description? It seems like a big and important one - can we add something?

Not related to this PR; you can open an issue

MariaLeonova:
> > @axel7083 do you know why this particular model doesn't have any description? It seems like a big and important one - can we add something?
> 
> Not related to this PR; you can open an issue

https://github.com/containers/podman-desktop-extension-ai-lab/issues/1881

packages/frontend/src/pages/InferenceServerDetails.svelte
```diff
@@ -218,84 +214,124 @@ function handleOnChange(): void {
...
-                        {service.labels['api']}
-                        <Fa class="ml-2" icon={faPlug} />
-                      </CopyButton>
-                    {/if}
-                  {/if}
+                        <!-- Documentation URL -->
+                        <div class="grow">
+                          {#if 'docs' in service.labels}
+                            <Tooltip tip="Open swagger documentation">
+                              Access
```
gastoner, line 246: Access should reflect light/dark theme 


https://github.com/user-attachments/assets/3496a4f6-20ba-46df-bf7d-c937a508e9cb

Also I would say that the tooltip should be visible only when you hover the link `swagger documentation` and not the text `Access`

axel7083: image

image

Nice catch! Fixed :heavy_check_mark:

jeffmaury (approved):
LGTM

gastoner (approved):

Issue #1617, opened 2024-08-28
Create a form to start a new InstructLab session


jeffmaury:
Only the UI workflow should be implemented as we are investigating the images

axel7083:
Here is my first draft implentation for the `StartInstructLabSession` page.

![image](https://github.com/user-attachments/assets/40c310bf-f1ce-4f92-ba26-55d7802a00fb)

![image](https://github.com/user-attachments/assets/51201dc6-3e33-4e3e-b342-16a666e52f97)

## Mock-ups vs Implementation

![image](https://github.com/user-attachments/assets/641055c6-06fc-41a7-9601-1736f290177a)

Above is the mock-ups available [on penpot](https://design.penpot.app/#/workspace/97b0253e-42f9-817e-8004-e3d617836067/6e12068a-09a0-801a-8004-24b355541870?page-id=f03273e9-8179-80f3-8004-e7f92eac3e77).

1. Why are we using dropdowns for sections ? This is not something we are doing anywhere in podman-desktop. 
2. Why the session name is the last field ? This feel a bit weird to give a name at the end
3. Model selection do not match design of `StartRecipe` and `CreateService`.

The inspiration for `(+) Adding knowledge` and `(+) Adding skills` are from the `Container Import` page

![image](https://github.com/user-attachments/assets/0d3995d1-1cbb-4948-b8f2-0f3dab8cc0eb)

## Going deeper

I was curious to see how we could give more information about each section, and redirect user to the Instructlab repository to learn more on what is a skill and what is a knowledge, but not sure.

![image](https://github.com/user-attachments/assets/0f534ac0-23e6-4dcc-a467-e4e9bff2b8da)

I tried adding `(?)` next to the form labels, but I feel like the link color get redundant and weird on the eyes.

## Form validation

To make the `submit` button valid, I used the following 

https://github.com/containers/podman-desktop-extension-ai-lab/blob/2a1f82e5dd93c5f08894e1650cea7ae4f97dba42/packages/frontend/src/pages/StartInstructLabSession.svelte#L22

This mean either a skills or a knowledge is necessary (a combination is possible but at least one of them is required)

## Current branch

I am working on https://github.com/axel7083/ai-studio/tree/feat/start-instructlab-session if people want to experiment with the form.

cc @slemeur @jeffmaury

slemeur:
cc @MariaLeonova

MariaLeonova:
[...]

PR #1849, opened 2024-10-07
feat: start instructlab session

axel7083:
### What does this PR do?

After discussion happening in https://github.com/containers/podman-desktop-extension-ai-lab/issues/1617, here is the frontend implementation for the `New Instruct Lab session` page
 
### Screenshot / video of UI

https://github.com/user-attachments/assets/66421239-8a15-426f-8f04-d1dbdb491e3e

ℹ️ I took inspiration from the `Play Kubernetes YAML` page for the radio selection. It give better consistency has it was the only place with similar design

image

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/1617

### How to test this PR?


packages/frontend/src/pages/NewInstructLabSession.svelte
```diff
@@ -0,0 +1,260 @@
...
+              <div class="flex flex-col">
+                {#each skills_files as file (file)}
+                  <div
+                    class="bg-[var(--pd-label-bg)] text-[var(--pd-label-text)] max-w-full rounded-md px-2 py-1 mb-2 flex flex-row w-min h-min text-sm text-nowrap items-center">
+                    <Fa class="mr-2" icon={faFile} />
+                    <span class="overflow-x-hidden text-ellipsis max-w-full">
+                      {file}
+                    </span>
+                    <Button
+                      on:click={function (): void {
```
benoitf, line 223: wondering here why we need the function () and not give the removeSkills directly

packages/frontend/src/pages/NewInstructLabSession.svelte
```diff
@@ -0,0 +1,260 @@
...
+                <div
+                  class="pl-2"
+                  class:text-[var(--pd-content-card-text)]={trainingType === 'skills'}
+                  class:text-[var(--pd-input-field-disabled-text)]={trainingType !== 'skills'}>
+                  Add skills
+                </div>
+              </button>
+              <!-- files list -->
+              <div class="flex flex-col">
+                {#each skills_files as file (file)}
```
benoitf, line 215: camelCase for skilsFiles

packages/frontend/src/pages/NewInstructLabSession.svelte
```diff
@@ -0,0 +1,260 @@
...
+  studioClient
+    .openURL(
+      'https://github.com/instructlab/instructlab?tab=readme-ov-file#-creating-new-knowledge-or-skills-and-training-the-model',
+    )
+    .catch((err: unknown) => {
+      console.error(err);
+    });
+}
+
+function submit(): void {}
```
benoitf, line 88: What's the goal of this function ?

axel7083: We don't have a backend, so pointless.

packages/frontend/src/pages/NewInstructLabSession.svelte
axel7083: done, used binding

jeffmaury:
I'm questioning the Skills or Knowledges option, I don't see where in InstructLab this is a requirement.

MariaLeonova:
@axel7083 looks good - I'm glad to see this coming together! Couple minor things I'd like to request: 
- Can the `+add knowledge/skill to use` button be left-aligned as the rest of the UI elements, since there's no input field next to it?
- The suggestion was for section to be called "Fine-tune with InstructLab" - I see the UI is structured a bit differently from [the mock](https://private-user-images.githubusercontent.com/14108590/373563894-91dd53f4-423a-41f7-be61-0d7fa16e6a86.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjgzMzI5ODMsIm5iZiI6MTcyODMzMjY4MywicGF0aCI6Ii8xNDEwODU5MC8zNzM1NjM4OTQtOTFkZDUzZjQtNDIzYS00MWY3LWJlNjEtMGQ3ZmExNmU2YTg2LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMDclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDA3VDIwMjQ0M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTJkNDViMDZmNzBlODI0YjEzY2I2YWFhOTgyZDJjZTU1ZmU0YzhjNDQ3OGY3NmZlMjQzYjQzZDlmZjgzNDE3NTcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.dI_1_cGZqQVZ66pbnOcMQDHTeRSJcedxHvQY155_GCg), but I think it will benefit from this title.
- On that note, let's agree on the consistent spelling: my suggestion is Fine-tune and fine-tuning, as on [Hugging Face](https://huggingface.co/docs/transformers/en/training#fine-tune-a-pretrained-model)

@jeffmaury let's check with the InstructLab team - do you know who would be the best person to contact there for more info?

gastoner:
I would change the "not selected" knowledge/skill `faCircleCheck` to some blank one or just some circle, because since we have two `faCircleCheck`, it seems to me like you have both enabled and you are just adding skills/knowledge to both of them in a same time

MariaLeonova:
> I would change the "not selected" knowledge/skill `faCircleCheck` to some blank one or just some circle, because since we have two `faCircleCheck`, it seems to me like you have both enabled and you are just adding skills/knowledge to both of them in a same time. Or use just radio buttons image

@axel7083 I second that - the pattern I added in the mock already exists in Podman Desktop UI, soo I think that element can be reused. 
Screenshot 2024-10-08 at 8 35 20

jeffmaury:
> @axel7083 looks good - I'm glad to see this coming together! Couple minor things I'd like to request:
> 
>     * Can the `+add knowledge/skill to use` button be left-aligned as the rest of the UI elements, since there's no input field next to it?
> 
>     * The suggestion was for section to be called "Fine-tune with InstructLab" - I see the UI is structured a bit differently from [the mock](https://private-user-images.githubusercontent.com/14108590/373563894-91dd53f4-423a-41f7-be61-0d7fa16e6a86.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjgzMzI5ODMsIm5iZiI6MTcyODMzMjY4MywicGF0aCI6Ii8xNDEwODU5MC8zNzM1NjM4OTQtOTFkZDUzZjQtNDIzYS00MWY3LWJlNjEtMGQ3ZmExNmU2YTg2LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMDclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDA3VDIwMjQ0M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTJkNDViMDZmNzBlODI0YjEzY2I2YWFhOTgyZDJjZTU1ZmU0YzhjNDQ3OGY3NmZlMjQzYjQzZDlmZjgzNDE3NTcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.dI_1_cGZqQVZ66pbnOcMQDHTeRSJcedxHvQY155_GCg), but I think it will benefit from this title.
> 
>     * On that note, let's agree on the consistent spelling: my suggestion is Fine-tune and fine-tuning, as on [Hugging Face](https://huggingface.co/docs/transformers/en/training#fine-tune-a-pretrained-model)
> 
> 
> @jeffmaury let's check with the InstructLab team - do you know who would be the best person to contact there for more info?

Best is to ask on one of their Slack channels

MariaLeonova:
Folks, the InstructLab beta UI is out, and they have nice short descriptions for **knowledge and skills** contributions, shall we consider adding those to fields helper text? 

🌟
`Knowledge in InstructLab is represented by question and answer pairs that involve facts, data, or references.`
🌟
`Skills are performative. When you create a skill for the model, you are teaching it how to do something: "write me a song," "rearrange words in a sentence" or "summarize an email."`
🌟

packages/frontend/src/pages/NewInstructLabSession.svelte
```diff
@@ -0,0 +1,250 @@
...
+
+let valid: boolean = $state(false);
+
+let sessionName: string = $state('');
+let model: ModelInfo | undefined = $state(undefined);
+
+let trainingType: 'knowledge' | 'skills' = $state('knowledge');
+
+$effect(() => {
+  console.log(model);
```
benoitf, line 23:
```suggestion
```

packages/frontend/src/pages/NewInstructLabSession.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    // models store
+    modelsInfoSubscribeMock: vi.fn<() => ModelInfo[]>(),
+    modelsInfoQueriesMock: {
+      subscribe: (f: (msg: unknown) => void) => {
+        f(mocks.modelsInfoSubscribeMock());
+        return (): void => {};
+      },
+    },
+  };
+});
```
benoitf, line 36: why do we need the hoisted pattern there ?

packages/frontend/src/pages/NewInstructLabSession.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+vi.mock('../stores/modelsInfo', async () => {
+  return {
+    modelsInfo: mocks.modelsInfoQueriesMock,
+  };
+});
+
+vi.mock('../utils/client', async () => ({
+  studioClient: {},
+}));
```
benoitf, line 46: do we need to mock the module or just spy the method

packages/frontend/src/pages/NewInstructLabSession.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+    },
+  };
+});
+
+vi.mock('../stores/modelsInfo', async () => {
+  return {
+    modelsInfo: mocks.modelsInfoQueriesMock,
+  };
+});
```
benoitf, line 42: do we need to mock the module or just spy the method

(because  using mock with relative path is not working well on refactoring)

packages/frontend/src/pages/NewInstructLabSession.spec.ts
axel7083: We do not use the svelte writters directly, we use a custom mechanism to be able to sync properties from the backend in the stores

https://github.com/containers/podman-desktop-extension-ai-lab/blob/8a1b04135c2629243dc2a7e1ae0936be2eaa6a3a/packages/frontend/src/stores/modelsInfo.ts#L24

We need the mocks pattern because we cannot spied the subscribe function. This is the same we are doing in all other forms mocking this store.

packages/frontend/src/pages/NewInstructLabSession.spec.ts
axel7083: The studioClient need to be mocked because we are declaring a lot of stuff globally, and registering proxy etc.

packages/frontend/src/pages/NewInstructLabSession.spec.ts
axel7083: Yes, as mention above, this module import the studioClient, which declare tons of stuff globally, and need to be mocked completely, like we are doing in every tests in ai lab.

MariaLeonova:
Adding a comment from Máirín Duffy on the InstructLab side: 

> for laptop training, full pipeline, right now - you can train knowledge or skills. not both in a training run.

packages/frontend/src/pages/NewInstructLabSession.spec.ts
benoitf: AFAIK It's possible to mock, so I don't see why you're saying we can't

no need to use any hoisted

```ts
vi.mock('../stores/modelsInfo', async () => ({
  modelsInfo: {
    subscribe: vi.fn(),
    unsubscribe: vi.fn(),
  },
  }));

beforeEach(() => {
  const infos: Writable<ModelInfo[]> = writable([]);
  vi.mocked(modelsInfo).subscribe.mockImplementation((run) => (infos.subscribe(run)));
});
```

benoitf (changes requested):
please avoid to use hoisted, as we can do it with less code

also I've noticed that we're almost not testing anything in this file

most of functions called are never checked, I think only the training type callback is being checked

but for example openInstructLabDocumentation is not, addKnowledge is not, removeSkills is not, removeKnowledge is not or requestExplorerModal is not, etc

packages/frontend/src/pages/NewInstructLabSession.spec.ts
axel7083: Oh yeah that's better, thanks for noticing, I get lost with the many different manner we using vitest

packages/frontend/src/pages/TuneSessions.svelte
```diff
@@ -13,7 +13,9 @@ import InstructlabColumnStatus from '../lib/table/instructlab/InstructlabColumnS
 import { router } from 'tinro';
 import Route from '../Route.svelte';
 
-function start(): void {}
+function start(): void {
+  router.goto('/tune/start');
+}
```
benoitf, line 18: it's probably untested

benoitf (commented):
reviewing that PR I noticed that we have a store to get the modelInfos and a method on the studioClient `studioClient.getModelsInfo` and the store delegates to the method, so I'm finding this confusing

Like here, in the component NewInstructlabSession, we import both studioClient and the store but there are overlap between the two

axel7083:
> reviewing that PR I noticed that we have a store to get the modelInfos and a method on the studioClient `studioClient.getModelsInfo` and the store delegates to the method, so I'm finding this confusing

We cannot subscribe to the `studioClient.getModelsInfo`. We could argue that we could get those value in the onMount, and there is little chance that the models will change on this page, but this is not the choice made in other forms, for reactivity purpose.

> Like here, in the component NewInstructlabSession, we import both studioClient and the store but there are overlap between the two

The store is automatically updated every time the backend notify the webview through the `Messages.MSG_NEW_MODELS_STATE`  channel 

https://github.com/containers/podman-desktop-extension-ai-lab/blob/c165dbbb689773b2c264dbbb311a9820357d7da2/packages/frontend/src/stores/modelsInfo.ts#L24

This is why we have two components here, they don't overlap has the store has an auto update logic based on event, that we do not have with a simple `studioClient.getModelsInfo`.

We import the studioClient to access to two method `studioClient.openDialog` (open the explorer to select a folder) and `studioClient.openUrl` to open in the browser the documentation url.

In AI Lab we do not declare global method like in Podman Desktop, where all the communication between the renderer and the preload are made through the `window.*` object.

benoitf:
> This is why we have two components here, they don't overlap has the store has an auto update logic based on event, that we do not have with a simple studioClient.getModelsInfo.

From consumer side, you've two different equivalent as it's from the same source. Exposure is changing.
Someone onboarding into the project might think it's more convenient to import only the client and then call client.getModelsInfo() But then, why I should have access to something not being reactive ? Also you can use get(modelsInfo) to get the value for one time. (no subscribe/unsubscribe)

I'm also offtrack of this PR, I don't require any change on this PR about that, but probably need to have a follow-up issue

my point is more : it looks that there is something more 'internal' to grab data from the backend side but client should never consume this endpoint directly. And then there are methods exposed to the client that he can use.
While here, it's a mix of two. Because if you need to ask something like 'publicly'

And stores being deprecated over $state() it could be revisited at that point.

axel7083:
> my point is more : it looks that there is something more 'internal' to grab data from the backend side but client should never consume this endpoint directly. And then there are methods exposed to the client that he can use.
While here, it's a mix of two. Because if you need to ask something like 'publicly'

I get your point, and totally understand it, there is clearly room for improvement, especially with the svelte 5 new requirements.

> I'm also offtrack of this PR, I don't require any change on this PR about that, but probably need to have a follow-up issue

I agree, I opened https://github.com/containers/podman-desktop-extension-ai-lab/issues/1883 to continue this conversation, as I think this is a valid point, and I would welcome change to improve this system.

I want to precise however that this system has been design and put together the first days of the hackathons in January, and has not been touch much since.

benoitf (approved):
thanks for the changes, it looks like the TuneSessions.svlete change has still no unit test

packages/frontend/src/pages/TuneSessions.svelte
axel7083: Now it is

packages/frontend/src/pages/NewInstructLabSession.svelte
```diff
@@ -0,0 +1,255 @@
...
+          <div class="flex flex-col gap-y-1">
+            <!-- knowledge -->
+            <div
+              class:border-2={true}
+              class="rounded-md p-5 cursor-pointer bg-[var(--pd-content-card-inset-bg)] flex flex-col gap-y-2"
+              aria-label="Knowledge"
+              aria-pressed={trainingType === 'knowledge' ? 'true' : 'false'}
+              class:border-[var(--pd-content-card-border-selected)]={trainingType === 'knowledge'}
+              class:border-[var(--pd-content-card-border)]={trainingType !== 'knowledge'}>
+              <button
```
jeffmaury, line 131: Why Button is not used ?

jeffmaury (commented):
Can we make the session name optional ?

packages/frontend/src/pages/NewInstructLabSession.svelte
axel7083: We don't have backend so no submit see https://github.com/containers/podman-desktop-extension-ai-lab/pull/1849#discussion_r1795073662

axel7083:
> Can we make the session name optional ?

Feel reasonable to require a name cc @MariaLeonova

MariaLeonova:
> > Can we make the session name optional ?
> 
> Feel reasonable to require a name cc @MariaLeonova

@jeffmaury why do you suggest to make it optional? 

The idea there was for people to be able to easily find their session in a list later. That's something that's done in the IBM Cloud flow, as well. I suppose it can be pre-filled with an automatically generated name, if we don't want users to have to type anything in?

axel7083:
> That's something that's done in the IBM Cloud flow, as well.

Personal preference to have it mandatory

jeffmaury:
We don't require it for playgrounds and services

jeffmaury:
> @jeffmaury ultimately it's up to you, as I don't have enough background info on user behaviour in playgrounds and services. Just as long as users are allowed to change the name there to something that makes sense to them - I think in this case it's important to give them control over their session list. Which seems to be the case for playgrounds, at least. Screenshot 2024-10-14 at 9 14 46

Yes I agree session name should stay

MariaLeonova (commented):
The +add button shouldn't be centered, if we are going with this pattern.

axel7083:
> The +add button shouldn't be centered, if we are going with this pattern.

image

Updated :rocket:

slemeur:
@MariaLeonova: I have a couple of UX feedback.

**When the box is not selected / no files added**
It really feels that we are duplicating the labels. There is "Add Knowledge" title and below "Add Knowledge to use", that feels redundant to me. 
I'd also callout that the "Add Knowledge to use" does not look like a button and that does not feel very intuitive.

**When a file has been added**
The description of what the section is about feels not at the right location. That should probably be before the action.

MariaLeonova:
> @MariaLeonova: I have a couple of UX feedback.
> 
> **When the box is not selected / no files added** It really feels that we are duplicating the labels. There is "Add Knowledge" title and below "Add Knowledge to use", that feels redundant to me. I'd also callout that the "Add Knowledge to use" does not look like a button and that does not feel very intuitive.
>

I agree! It used to say 'add another knowledge file' in the mock.
New IL Session Form- 4 _ 10 (1)

> **When a file has been added** The description of what the section is about feels not at the right location. That should probably be before the action.
 
@axel7083 can you please add a screenshot of what it looks like when there are 2 files added already? We might want to shuffle things a bit - the addition of `+add more knowledge/skills` button has rearranged elements in the form.

MariaLeonova:
@axel7083 I am not opposed to opening a follow-up issue for design improvement, so we don't have to go back and forth and can change all at once, if you'd rather merge this now.

Issue #1661, opened 2024-09-04
feat: allow to disable wsl upload through hidden configuration


axel7083:
### Is your enhancement related to a problem? Please describe

In the e2e tests we often have a problem where uploading the model to the WSL leads to freeze. The source of the problem is unknown, and is blocking for QE.

The main lead are related to the WSL Upload code, which is using the podman cli (E.g. podman.exe) which freeze for unknown reasons.

### Describe the solution you'd like

As discussed with @jeffmaury and @ScrewTSW, we could add an hidden option in the configuration to disable the WSL Upload.

axel7083:
@ScrewTSW do we want to close this one ? or are we still interested to have this option ?

ScrewTSW:
> @ScrewTSW do we want to close this one ? or are we still interested to have this option ?

I'd say we are still interested in having the option.

PR #1857, opened 2024-10-08
Parse system environment if property is not set or does not exist

ScrewTSW:
### What does this PR do?
* Allows for setting model upload hidden property using an environment variable

### Screenshot / video of UI
N/A

#1661 

### How to test this PR?
N/A

### Extra context
This change was proposed, because at the time of fresh installation, pre-setting the property in settings.json would cause Podman Desktop to instantly crash at the moment the configuration storage tries to initialize. This environment variable allows us to circumvent that problem, especially on CI use

### DIFF
--- a/packages/backend/src/registries/ConfigurationRegistry.ts
+++ b/packages/backend/src/registries/ConfigurationRegistry.ts
@@ -50,7 +50,8 @@ export class ConfigurationRegistry extends Publisher<ExtensionConfiguration> imp
       experimentalGPU: this.#configuration.get<boolean>('experimentalGPU') ?? false,
       apiPort: this.#configuration.get<number>('apiPort') ?? API_PORT_DEFAULT,
       experimentalTuning: this.#configuration.get<boolean>('experimentalTuning') ?? false,
-      modelUploadDisabled: this.#configuration.get<boolean>('modelUploadDisabled') ?? false,
+      modelUploadDisabled:
+        this.#configuration.get<boolean>('modelUploadDisabled') ?? process.env.AI_LAB_MODEL_UPLOAD_DISABLED === 'true',
     };
   }
 


axel7083:
I thought the podman desktop e2e testing was suppose to be able to set configuration value ?

axel7083:
> Podman Desktop to instantly crash at the moment the configuration storage tries to initialize

Maybe an issue should be open in podman desktop? This is not something that is expected probably.

jeffmaury:
> @axel7083 I think that @ScrewTSW hsa opened the issue already on podman-desktop side.
> 
> The problem is that at the time of e2e starting podman-desktop and having a chance to touch settings.json is way before we have the configuration option available from ai-lab extension, which is added once the extension is installed.
> 
> The question is if the bug/issue is resolved and podman desktop can spin up with settings.json with property not known at the time, it will be considered after ai-lab installation?
> 
> So the one of two options we have is to be able to define the value of the property from outside the running app -> by using env. var. Or we would need to develop our own extension that would adjust settings the way we need. Or make make property visible (unlikely, right?)

I think this is a dirty patch, why can't you generate settings.json before you launch Podman Desktop or if there is any equivalent between settings and env variables it should be in core Podman Desktop (like what Quarkus does for properties)

odockal:
@axel7083 @ScrewTSW I just tried again to run the e2e tests with settings adding the `"ai-lab.modelUploadDisabled": true,` at the pd start up without ai-lab installed and it did not crashed this time. I not sure what has fixed the problem. I cannot even find the original issue we were facing. Once we will have confirmation we can close the PR and ... nevermind I guess...

odockal:
> I think this is a dirty patch, why can't you generate settings.json before you launch Podman Desktop or if there is any equivalent between settings and env variables it should be in core Podman Desktop (like what Quarkus does for properties)

We do, but there was a failure causing podman desktop to crash. This was the state like two weeks ago, as I said above, I tried again, and the bug is not there anymore. At least I tried on linux right now and it works as expected.

ScrewTSW:
original bug not reproducible on windows CI either, closing

Issue #1871, opened 2024-10-09
Implement E2E test case covering Code Generation project deployment from Recipes Catalog


odockal:
### Is your enhancement related to a problem? Please describe

Cover the use case with e2e tests that installs, starts up and deploys Code Generation from Recipes Catalog.

### Describe the solution you'd like

Use playwright tests to cover the scenario.

PR #1891, opened 2024-10-11
chore(test): add code gen test

cbr7:
### What does this PR do?
Adds ai lab code generation e2e smoke test

https://github.com/containers/podman-desktop-extension-ai-lab/issues/1871


axel7083:
50 commits ? missing some rebase ?

cbr7:
The commits will get squashed upon merge.

odockal:
> 50 commits ? missing some rebase ?

@cbr7 I think that @axel7083 meant that this branch is not synced with main. So rebase would be nice.

cbr7:
It was only slightly behind main, I was keeping it updated, anyway it's synced now.

odockal (changes requested):
@cbr7 please take a look at pr check error: `Error while cleaning up service models: TypeError: Cannot read properties of undefined (reading 'navigationBar')`

Otherwise I think we are good with PR. We need to make sure we can run it on CI.
Also, we need to make sure to keep actual workflows running without problems, like pr check.

cbr7:
@odockal that error is written from try-catching an error, it's fine for it to be there, it happens when the framework tries to execute the afterAll hook after skipping all the tests since this is on Linux. There is no issue with it.

odockal:
@cbr7 I have rerun the job and now it is ok. Strange! Anyway, these tests are not run on PR check and Nightly on Linux, so they do not block anything. And we need to figure out to make them run on windows.. or Mac?

odockal:
@cbr7 I would be for merging and creating follow up issue to make them stable on both Windows and Mac. wdyt @ScrewTSW @axel7083  ?

cbr7:
@odockal about windows I don't know what more I can do, I've managed to isolate where the freeze is located, but at this point investigating further is beyond my capabilities, as for mac, once it's merged I can look into it as part of the other task that stabilizes e2e tests on macOS that I have.

odockal:
@cbr7 Np. So what will happen when we encounter the problem on the first app to test and it takes full timeout of the test? Am I correct that since it is serial it wont try to run the rest of recipes? 

It would be nice that it wont spend resources if we know it wont be successful.

odockal (approved):

cbr7:
@benoitf @jeffmaury can you please review if you have some free time

benoitf (approved):

axel7083 (approved):

PR #1979, opened 2024-10-23
chore: enable pre-commit

maryamtahhan:
### What does this PR do?

- Enables [pre-commit](https://pre-commit.com/) hook for the repo.
- Runs the pre-commit plugins on the files in the repo and applied the relevant changes.

### How to test this PR?

```bash
pip install pre-commit
pre-commit install
pre-commit run --all-files
```


axel7083:
Thanks @maryamtahhan, I do not think podman-desktop are familair with pre-commit, on podman-desktop repository we seems to be using husky, we also have an issue related to using it on ai-lab https://github.com/containers/podman-desktop-extension-ai-lab/issues/1633

jeffmaury:
Thanks for your contribution. However, it requires Python which is too much. Podman Desktop uses Husky for such, would you mind going that direction ?

benoitf (commented):
thanks for the contribution,

as the PR is doing lot of different things, I would encourage you to open small chunks before providing a full solution.

For example, file a PR fixing all the codespell issue

then, file an issue prompting if it's possible to enable some code spelling issues on the repository. Probably here I would recommend that it's ensured on the PR check / GitHub action because pre-commit hooks are great but people can bypass them

then, open a dedicated PR just removing trailing whitespaces or EOF, etc.

maryamtahhan:
No problem - I can break down the issues the pre-commit hooks picked up into separate PRs (without the actual pre-commit hook itself)...

I will have a look at the husky route.

benoitf:
also for the trailing spaces, EOF the repository is using prettier but maybe some files are ignored/skipped explaining why few files have problems

Issue #989, opened 2024-04-24
Navbar text and icons not aligned


cdrage:
![Screenshot 2024-04-24 at 7 48 19 PM](https://github.com/containers/podman-desktop-extension-ai-lab/assets/6422176/a25f5e08-9f39-4c75-9003-4d47f39efac3)

Looks like the side navbar isn't aligned?

slemeur:
has been fixed

cdrage:
This has not been fixed and I still see this on the main branch, opening up a PR

PR #2011, opened 2024-10-29
chore: fix unaligned navigation bar

cdrage:
chore: fix unaligned navigation bar

### What does this PR do?

The icons are not aligned because the "faBookOpen" icon is actually
576x512 and not the usual 512x512, causing the icons to be misaligned.

Inspection of faBookOpen:

```html
<svg aria-hidden="true" role="img" xmlns="http://www.w3.org/2000/svg" class="svelte-fa svelte-fa-base mr-4 svelte-l34xo3" viewBox="0 0 576 512">
```

This PR modifies the icon size to actually be 512px width to avoid
misalignment

### Screenshot / video of UI

**Look at the "open book" icons and text vs the ones below**

Before:
Screenshot 2024-10-29 at 9 03 25 AM

After:
Screenshot 2024-10-29 at 9 01 32 AM

Closes https://github.com/containers/podman-desktop-extension-ai-lab/issues/989

### How to test this PR?

See that it has been aligned

Signed-off-by: Charlie Drage <charlie@charliedrage.com>

### DIFF
--- a/packages/frontend/src/lib/Navigation.svelte
+++ b/packages/frontend/src/lib/Navigation.svelte
@@ -7,12 +7,20 @@ import { onDestroy, onMount } from 'svelte';
 import { configuration } from '../stores/extensionConfiguration';
 import type { ExtensionConfiguration } from '@shared/src/models/IExtensionConfiguration';
 import type { Unsubscriber } from 'svelte/store';
+import type { IconDefinition } from '@fortawesome/free-regular-svg-icons';
 
 export let meta: TinroRouteMeta;
 let experimentalTuning: boolean = false;
 let cfgUnsubscribe: Unsubscriber;
 
+// By default, faBookOpen is 576x512, but we want it to be 512x512
+// we cannot modify the width and height in SettingsNavItem so just modify the icon instead
+let copyFaBookOpenIcon: IconDefinition | undefined = undefined;
+
 onMount(() => {
+  copyFaBookOpenIcon = structuredClone(faBookOpen);
+  copyFaBookOpenIcon.icon[0] = 512;
+
   cfgUnsubscribe = configuration.subscribe((val: ExtensionConfiguration | undefined) => {
     experimentalTuning = val?.experimentalTuning ?? false;
   });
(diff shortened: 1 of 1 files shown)


gastoner (approved):
Nice fix!

packages/frontend/src/lib/Navigation.svelte
```diff
@@ -12,6 +12,11 @@ export let meta: TinroRouteMeta;
 let experimentalTuning: boolean = false;
 let cfgUnsubscribe: Unsubscriber;
 
+// By default, faBookOpen is 576x512, but we want it to be 512x512
+// we cannot modify the width and height in SettingsNavItem so just modify the icon instead
+let faBookOpenIcon = faBookOpen;
```
benoitf, line 17: issue: I think you should clone the values of the original object using spread operator

cdrage: Unsure what you mean? Is using 'let' part of cloning it?

benoitf: something like const myIcon = {...originalIcon};

benoitf: but in your case as you've array inside property it won't do a deep clone

So maybe using https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone#description method

benoitf: or the hack  JSON.parse(JSON.stringify(object))

cdrage: Hmmm unfortunatley using both:

```
let faBookOpenIcon = JSON.parse(JSON.stringify(faBookOpen))
faBookOpenIcon.icon[0] = 512;
```

or:

```
let faBookOpenIcon = structuredClone(faBookOpen);
```

Actually causes issue. Okay if we just keep the original solution?

We do not modify any of the "original object values" since all are default anyways, we just force 512 width.

benoitf: > We do not modify any of the "original object values" since all are default anyways, we just force 512 width.

You modify the original definition of the icon. The problem is that if I import the same icon in another place, you're changing the value everywhere, not only in Navigation.svelte

I tried
```
let copyFaBookOpenIcon: IconDefinition | undefined = undefined;

onMount(() => {
// By default, faBookOpen is 576x512, but we want it to be 512x512
// we cannot modify the width and height in SettingsNavItem so just modify the icon instead
 copyFaBookOpenIcon = structuredClone(faBookOpen);
 copyFaBookOpenIcon.icon[0] = 512;

```

and then replace occurence of `faBookOpen` by `copyFaBookOpenIcon` and I didn't see issue

what are the issues you're seing ?

cdrage: I was getting errors when it was copying over the JSON. I think I found the issue, I wasn't doing it within onMount! Thanks for the explanation. Never realized it was going to change the value everywhere within svelte.

cdrage:
@benoitf updated!

benoitf (approved):
thx

cdrage:
fixed formatting (pnpm format:fix), setting as auto-merge now!

PR #2052, opened 2024-11-05
chore: changed destination of ai lab port settings

### LINKED ISSUE #1978: Move the AI Lab listening port in its own configuration page

Image

Today, when you are enabling AI Lab, it starts an Ollama compatible endpoint, so a user can programmatically interact with Podman AI Lab. 

The problem is this is not very discoverable and the configuration of the endpoint port feels disconnected as well.

**1 - Remove AI Lab port from the status bar**

**2 - Create a new page accessible from Podman AI Lab sidebar**
Name of the page: "AI Lab Service" 

In this page, we will provide the following explanation:
Integrate Podman AI Lab directly into your development workflows by using its REST API endpoints.  Compatible with Ollama's endpoints, you can seamlessly access and utilize the capabilities of Podman AI Lab without relying on its graphical interface

We will add a link to the APIs we are supporting.

Then, we should have a section with "Local Server" and a dedicated page "Server Settings".
The first item, will be the Port to listen on.

We should be keeping the possibility to change the Port in the settings of Podman AI Lab extension as well

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Changes destination of ai lab port settings

### Screenshot / video of UI
image

Closes #1978 

### How to test this PR?
Go to AI Lab, there should be a new navigation item with the settings


gastoner:
TODO: add unit test

benoitf:
question: it's not clear to me in the issue if it means that the configuration should be removed from the classic configuration page of any extension.
IMHO it feels an anti-pattern if it's the case

benoitf:
side comment: also it's possible to provide a markdown description in the properties. So it looks like we're trying to reinvent the wheel of how each extension will be configured

gastoner:
cc @slemeur

packages/backend/package.json
```diff
@@ -44,13 +44,6 @@
           "default": false,
           "description": "Experimental GPU support for inference servers"
         },
-        "ai-lab.apiPort": {
```
jeffmaury, line 47: This should not be removed

jeffmaury (changes requested):

slemeur:
@benoitf: The intent was to expose the REST Server capabilities from Podman AI Lab, by providing its own page. Right now, it would only explain what it does and provides the ability to configure the server port. 

In the future the user could in the future:
- start/stop the local server
- access to end points API documentation
- access the logs

Does that make sense to you?

Considering your feedback, I agree that the "Configurable Options" is going to be confusing. 

Maybe, we should call the section: "Local Server" and the subpage "Server Information", wdyt?

benoitf:
@slemeur my point was that any configuration should stay at a minimum in the "generic configuration" settings (what is being exposed by an extension)

then the extension can still decide to expose it through a different UI, but being "on top of the configuration settings"

so, more keep it at both place (if I go to all settings and filter for AI Lab extension, I should be able to find something there)

slemeur:
Ok that makes sense indeed.

slemeur:
@gastoner: I've updated the initial issue.

Here are the changes for your PR:
- Section title "Local Server" 
- Dedicated page "Server Information"
- We will be keeping the possibility to change the Server Port in the settings of Podman AI Lab extension as well

packages/frontend/src/App.svelte
```diff
@@ -110,6 +111,11 @@ onDestroy(() => {
           <Preferences />
         </Route>
 
+        <!-- AI Lab Service -->
+        <Route path="/ai_lab_service">
+          <AiLabService />
+        </Route>
+
```
benoitf, line 118: question: why name it AI Lab service (the AI lab prefix ?), as it's already part of AI lab, needs to only be /service 

The same for AiLabService (need to remove the prefix)

packages/frontend/src/lib/Navigation.svelte
```diff
@@ -69,6 +70,16 @@ onDestroy(() => {
     <SettingsNavItem icon={faRocket} title="Services" selected={meta.url === '/services'} href="/services" />
     <SettingsNavItem icon={faMessage} title="Playgrounds" selected={meta.url === '/playgrounds'} href="/playgrounds" />
 
+    <!-- Settings -->
+    <div class="pl-3 mt-2 ml-[4px]">
+      <span class="text-[color:var(--pd-secondary-nav-header-text)]">Server Information</span>
```
benoitf, line 75: here it's confusing, in the image I see another text: Configurable Options

packages/frontend/src/App.svelte
benoitf: anyway, as it's within a section named ServerInformation , then it should be `ServerInformation` and route path `/server-information`

packages/frontend/src/lib/Navigation.svelte
```diff
@@ -69,6 +70,16 @@ onDestroy(() => {
...
+    </div>
+    <SettingsNavItem
+      icon={faGear}
+      title="Local Server"
+      selected={meta.url === '/ai_lab_service'}
+      href="/ai_lab_service" />
```
benoitf, line 81: name of the url/title should match the parent

like : `/server-information/service`

benoitf (changes requested):
need to change the naming


App should do the main routing for /server-information

ServerInformation should do the routing for this section (/server-information/service)

AiLabService should be in a folder `server-information` with filename Service.svelte

gastoner:
> need to change the naming
> 
> App should do the main routing for /server-information
> 
> ServerInformation should do the routing for this section (/server-information/service)
> 
> AiLabService should be in a folder `server-information` with filename Service.svelte

@benoitf 
It looks like the other https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/frontend/src/lib/Navigation.svelte#L70 navigation items (models, services, playgrounds,...) do not use the additional folder

benoitf (commented):
@gastoner you should update the image of the PR as I think it's not corresponding to the PR

packages/frontend/src/pages/server-information/LocalServer.svelte
```diff
@@ -0,0 +1,89 @@
...
+import { configuration } from '/@/stores/extensionConfiguration';
+import { onMount, onDestroy } from 'svelte';
+
+// The AI Lab Port is the bind value to form input
+let aiLabPort: number | undefined = $state(undefined);
+let invalidPort: boolean = $state(false);
+let errorMsg: string | undefined = $state(undefined);
+
+onMount(() => {
+  const port = $configuration?.apiPort;
```
benoitf, line 13: if there are some external changes to the port, it won't be reflected in the UI as it's only reading apiPort value when displaying the first time the value

so let say, another process update the apiPort configuration value, this page will still display the same value than when the page was loaded

benoitf (approved):
I see an issue about the reactivity of the configuration

the display of the value can be inconsistent with the last update of the value by the configuration system.

packages/frontend/src/pages/server-information/LocalServer.svelte
```diff
@@ -0,0 +1,89 @@
...
+      <div class="px-5 space-y-5">
+        <div class="bg-[var(--pd-content-card-bg)] m-5 space-y-6 px-8 sm:pb-6 xl:pb-8 rounded-lg h-fit">
+          <div class="w-full">
+            <!-- aiLabPort input -->
+            <label for="aiLabPort" class="pt-4 block mb-2 font-bold text-[var(--pd-content-card-header-text)]"
+              >Port on which the API is listening (requires restart of extension):</label>
+            <Input
+              type="number"
+              value={String(aiLabPort ?? 0)}
+              on:input={async (e): Promise<void> => await onAiLabPortInput(e)}
```
benoitf, line 76: nitpick: could we just say `on:input={onAiLabPortInput}`

packages/frontend/src/pages/server-information/LocalServer.svelte
```diff
@@ -0,0 +1,89 @@
...
+              class="w-full ml-2"
+              placeholder="10434"
+              name="aiLabPort"
+              aria-label="Port input" />
+          </div>
+          {#if errorMsg !== undefined}
```
benoitf, line 82:
```suggestion
          {#if errorMsg}
```

packages/frontend/src/pages/server-information/LocalServer.svelte
gastoner: Yes, I think that this is a reason why you need to restart the extension too, to reflect the changes

packages/frontend/src/lib/Navigation.svelte
jeffmaury: See https://github.com/containers/podman-desktop-extension-ai-lab/pull/2052#issuecomment-2459973798

jeffmaury (approved):
LGTM

packages/frontend/src/lib/Navigation.svelte
benoitf: my point was that the image attached to this PR should be updated to reflect the code changes

cc @gastoner

benoitf:
@gastoner please use an accurate screenshot for the PR description
https://github.com/containers/podman-desktop-extension-ai-lab/pull/2052#pullrequestreview-2446094245

I still se AI Lab Service and Configurable options https://github.com/containers/podman-desktop-extension-ai-lab/pull/2052#issue-2635115455 there

PR #2074, opened 2024-11-11
feat: Add the chatbot-java recipe

jmartisk:
### What does this PR do?

Add the `chatbot-java` recipe (created added via https://github.com/containers/ai-lab-recipes/pull/809) to the default recipe catalog

### DIFF
--- a/packages/backend/src/assets/ai.json
+++ b/packages/backend/src/assets/ai.json
@@ -70,6 +70,23 @@
       "backend": "llama-cpp"
     },
     {
+      "id": "chatbot-java-quarkus",
+      "description": "This is a Quarkus-based chat demo application.",
+      "name": "Java-based ChatBot (Quarkus)",
+      "repository": "https://github.com/containers/ai-lab-recipes",
+      "ref": "a7b03d7bd025e009a03147391a00cb1ee472ad3f",
+      "icon": "natural-language-processing",
+      "categories": ["natural-language-processing"],
+      "basedir": "recipes/natural_language_processing/chatbot-java",
+      "readme": "# Java-based chatbot application\n\nThis application implements a simple chatbot backed by Quarkus and its\nLangChain4j extension. The UI communicates with the backend application via\nweb sockets and the backend uses the OpenAI API to talk to the model served\nby Podman AI Lab.\n\nDocumentation for Quarkus+LangChain4j can be found at\nhttps://docs.quarkiverse.io/quarkus-langchain4j/dev/.",
+      "recommended": [
+        "hf.instructlab.granite-7b-lab-GGUF",
+        "hf.instructlab.merlinite-7b-lab-GGUF",
+        "hf.lmstudio-community.granite-3.0-8b-instruct-GGUF"
+      ],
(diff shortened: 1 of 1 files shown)


packages/backend/src/assets/ai.json
```diff
@@ -69,6 +69,23 @@
       ],
       "backend": "llama-cpp"
     },
+    {
+      "id": "chatbot-java",
+      "description": "This is a Java-based chat demo application.",
+      "name": "Java-based ChatBot",
+      "repository": "https://github.com/containers/ai-lab-recipes",
+      "ref": "main",
```
jmartisk, line 77: Not sure what to put here until it's part of a release, so I used `main` for now.

feloy: We would prefer to have a fixed ref here. You can use the full commit ID of the current HEAD for this

jmartisk: Ok @feloy, done

packages/backend/src/assets/ai.json
```diff
@@ -69,6 +69,23 @@
       ],
       "backend": "llama-cpp"
     },
+    {
+      "id": "chatbot-java",
```
feloy, line 73: perhaps do we want to specify it is quarkus-based, as we will probably add other java frameworks

feloy (commented):
Tested, works as expected.
Thanks for this contribution.

packages/backend/src/assets/ai.json
maxandersen: I personally would prefer simple names to not crowd the UI in podman desktop catalog view
yeah I was wondering same thing; but as the list in here just says "chatbot" without mentioning python nor its streamlit based .... so the biggest diff which in this case is "java" seemed good balance.

maybe `chatbot-java-quarkus`, `chatbot-python-streamlit`, `chatbot-python-flask` etc. is a good pattern to apply?

benoitf, line 73:
```suggestion
      "id": "chatbot-java-quarkus",
```

so it'll match suggested `exampleName-language-framework/platform`

benoitf (approved):

feloy (approved):
LGTM

packages/backend/src/assets/ai.json
jmartisk: @rhatdan / @jeffmaury agree with renaming this recipe (I'd first change it in https://github.com/containers/ai-lab-recipes for consistency)?

benoitf: yes please change first into ai-lab-recipes for consistency

jmartisk:
I've sent https://github.com/containers/ai-lab-recipes/pull/813 for the rename, so merging this one should wait for that one (I'll also have to update the commit hash after https://github.com/containers/ai-lab-recipes/pull/813 is merged)

jmartisk:
Ok, renamed it and updated the git ref, should be all good now

benoitf:
I launched the CI workflows

benoitf:
thanks @jmartisk

PR #2146, opened 2024-11-21
chore: updates id to correct form on linux

### LINKED ISSUE #1942: Details page of the imported model is empty
### Bug description

I have imported granite-8b-code-instruct.Q4_K_M (downloaded from demo, copied elsewhere, delete from models in podman desktop and reimported). Everything works fine, except the model details page is empty page. Other models offers some details page.

I did not have a readme by the model, but at least an info with the model name might be shown. 

Image

### Operating system

Fedora 40

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Updates model ID when importing on Linux to be in the correct form for Routes

Closes #1942 

### How to test this PR?
Steps in the issue (any locally imported model) 
Broken is only Linux, but just in case would be great to try out W and Mac

### DIFF
--- a/packages/backend/src/managers/catalogManager.ts
+++ b/packages/backend/src/managers/catalogManager.ts
@@ -23,7 +23,7 @@ import defaultCatalog from '../assets/ai.json';
 import type { Recipe } from '@shared/src/models/IRecipe';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import { Messages } from '@shared/Messages';
-import { type Disposable, type Event, EventEmitter, type Webview, window } from '@podman-desktop/api';
+import { type Disposable, type Event, EventEmitter, type Webview, window, env } from '@podman-desktop/api';
 import { JsonWatcher } from '../utils/JsonWatcher';
 import { Publisher } from '../utils/Publisher';
 import type { LocalModelImportInfo } from '@shared/src/models/ILocalModelInfo';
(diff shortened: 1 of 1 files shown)


gastoner:
I'll investigate further and try to find a place where would be better place to update the id - to fix the current models

benoitf (commented):
should we fix the id or the routes ?

like do we need the id be based from a path or it's just o have a key, we could use any identifier

if we're using the id to check against some path it could be an issue
but for that we have the file attribute


if we fix the route, we could compute a hash from the id (we do that in Podman Desktop) so the id stay the same as of today

gastoner:
The hash would be great. 
On Windows it works without problem because the path does not contain "/"
But I would imagine same problem on Mac with the Routes

gastoner:
`ModelInfo` have sha256 optional attribute - so during import I would compute sha and store it to the model and use it instead of path inside the `id`

benoitf:
PTAL @jeffmaury

packages/backend/src/managers/catalogManager.ts
```diff
@@ -223,6 +225,7 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
           },
           memory: statFile.size,
           backend: local.backend ?? InferenceType.NONE,
+          sha256: sha256,
```
benoitf, line 228: question: are we not expecting that sha256 is the sha of the model I suppose, not the sha of the path ?

jeffmaury (commented):
Does not work for me, still have an empty screen

import

gastoner:
@jeffmaury you have to import the model again

benoitf (approved):
just wondering if it impacts existing models (id mismatch)

benoitf (commented):
also maybe a unit test ?

axel7083 (approved):

Issue #2214, opened 2024-12-05
Select first compatible GPU when multiple are detected


axel7083:
### Is your enhancement related to a problem? Please describe

On certain machine we may have access to multiple GPU (certain GPU are integrated to the CPU, such as AMD Radeon(TM) Graphics.) but you can have an additional GPU (NVIDIA), since only the first one is detected, we may end up in a situation where we ignore the _real_ one.

### Describe the solution you'd like

In the best world forever we should let the user decide: this is probably too much work, so we can do a workaround and search for the first compatible one when multiple are available

PR #2238, opened 2024-12-09
feat: improve GPU support when multiple GPUs

mhdawson:
### What does this PR do?

If there are multiple GPUs use the first one that is of known type instead of first GPU. If no GPUs of are known type fall back to the first GPU as before.

Add another string for vendor that is accepted as an NVIDIA GPU when doing GPU detection based on what was seen on linux with an NVIDIA 4070 Ti Super

The VM in which I did testing for potential GPU support on linux (https://github.com/containers/podman-desktop-extension-ai-lab/pull/2180) had 2 CPUs, one of which is NVIDIA but it was not gpu[0]. This should have no effect in people only have 1 gpu. With this change I could use the second GPU.

### Screenshot / video of UI

N/A

https://github.com/containers/podman-desktop-extension-ai-lab/issues/2214

### How to test this PR?

Run unit tests as updated

### DIFF
--- a/packages/backend/src/workers/provider/LlamaCppPython.ts
+++ b/packages/backend/src/workers/provider/LlamaCppPython.ts
@@ -197,9 +197,10 @@ export class LlamaCppPython extends InferenceProvider {
     if (this.configurationRegistry.getExtensionConfiguration().experimentalGPU) {
       const gpus: IGPUInfo[] = await this.gpuManager.collectGPUs();
       if (gpus.length === 0) throw new Error('no gpu was found.');
-      if (gpus.length > 1)
-        console.warn(`found ${gpus.length} gpus: using multiple GPUs is not supported. Using ${gpus[0].model}.`);
-      gpu = gpus[0];
+
+      // Look for a GPU that is of a known type, use the first one found.
+      // Fall back to the first one if no GPUs are of known type.
+      gpu = gpus.find(({ vendor }) => vendor !== GPUVendor.UNKNOWN) ?? gpus[0];
     }
 
     let connection: ContainerProviderConnection | undefined = undefined;
--- a/packages/backend/src/managers/GPUManager.spec.ts
+++ b/packages/backend/src/managers/GPUManager.spec.ts
@@ -105,3 +105,25 @@ test('NVIDIA controller should return intel vendor', async () => {
     },
   ]);
 });
+
+test('NVIDIA controller can have vendor "NVIDIA Corporation"', async () => {
+  vi.mocked(graphics).mockResolvedValue({
+    controllers: [
+      {
+        vendor: 'NVIDIA Corporation',
+        model: 'NVIDIA GeForce GTX 1060 6GB',
+        vram: 6144,
+      } as unknown as Systeminformation.GraphicsControllerData,
+    ],
+    displays: [],
+  });
+
+  const manager = new GPUManager(webviewMock);
+  expect(await manager.collectGPUs()).toStrictEqual([
+    {
+      vendor: GPUVendor.NVIDIA,
+      model: 'NVIDIA GeForce GTX 1060 6GB',
+      vram: 6144,
+    },
+  ]);
+});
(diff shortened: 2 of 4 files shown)


packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -197,9 +197,21 @@ export class LlamaCppPython extends InferenceProvider {
...
+      let selectedGPU = 0;
+      if (gpus.length > 1) {
+        // Look for a GPU that is of a known type, use the first one found.
+        // Fall back to the first one if no GPUs are of known type.
+        for (let i = 0; i < gpus.length; i++) {
+          if (gpus[i].vendor !== GPUVendor.UNKNOWN) {
+            selectedGPU = i;
+            break;
+          }
+        }
```
benoitf, line 209: nit: use the Array.find method ?

benoitf (approved):
cannot test locally as I only have one GPU on my computer

packages/backend/src/workers/provider/LlamaCppPython.ts
mhdawson: I think we would need to use findIndex versus find. With that we'd have to check for -1 as that is what's returned if there is no match. Not sure it would be much less code or clearer.

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -197,9 +197,21 @@ export class LlamaCppPython extends InferenceProvider {
...
+        console.warn(
+          `found ${gpus.length} gpus: using multiple GPUs is not supported. Using ${gpus[selectedGPU].model}.`,
+        );
+      }
+      gpu = gpus[selectedGPU];
```
axel7083, line 214:
```suggestion
      if (gpus.length > 1) {
        // Look for a GPU that is of a known type, use the first one found.
        // Fall back to the first one if no GPUs are of known type.
        gpu = gpus.find(({vendor}) => vendor !== GPUVendor.UNKNOWN) ?? gpus[0];
      }
```

packages/backend/src/workers/provider/LlamaCppPython.ts
jeffmaury: If there is a single GPU we will take it but won't check if the vendor is known. Should we check it as well ?

mhdawson: @jeffmaury I left it not checking since that was the way it was before and I was not sure if that would break anything. I can add the check if that is the right thing to do.

packages/backend/src/workers/provider/LlamaCppPython.ts
mhdawson: The change missed the case with 0, cpus. pulled it in minus the check for number of cpus and it seems to do the right thing. Also rebaseed.

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -197,9 +197,10 @@ export class LlamaCppPython extends InferenceProvider {
     if (this.configurationRegistry.getExtensionConfiguration().experimentalGPU) {
       const gpus: IGPUInfo[] = await this.gpuManager.collectGPUs();
       if (gpus.length === 0) throw new Error('no gpu was found.');
-      if (gpus.length > 1)
-        console.warn(`found ${gpus.length} gpus: using multiple GPUs is not supported. Using ${gpus[0].model}.`);
-      gpu = gpus[0];
+
+      // Look for a GPU that is of a known type, use the first one found.
+      // Fall back to the first one if no GPUs are of known type.
+      gpu = gpus.find(({ vendor }) => vendor !== GPUVendor.UNKNOWN) ?? gpus[0];
```
mhdawson, line 203: If as @benoitf suggested we should only use recognized GPUs it should be as simple as removing the ?? gpus[0]

packages/backend/src/workers/provider/LlamaCppPython.ts
mhdawson: It is a lot more compact which is great.

axel7083 (approved):
LGTM

packages/backend/src/workers/provider/LlamaCppPython.ts
jeffmaury: Would be good because I understood your configuration have 2 GPUs returned one which is not to be selected so I wonder we could have a configuration where we could have this single GPU returned

axel7083:
Merging as got two approval and tests are passing

axel7083:
Thanks @mhdawson this is great !

PR #2240, opened 2024-12-09
feat: initial support for GPU on linux

mhdawson:
### What does this PR do?

Adds initial support for GPU acceleration for linux. It requires a version of Podman which
includes this PR - https://github.com/podman-desktop/podman-desktop/pull/10251

### Screenshot / video of UI

N/A

Refs: https://github.com/containers/podman-desktop-extension-ai-lab/issues/2162

### How to test this PR?

1. Unit tests are added.

2. With a version of podman which has https://github.com/podman-desktop/podman-desktop/pull/10251 landed, build a version of podman-desktop-extension-ai-lab with this PR and run podman desktop with the updated extension on Fedora with an NVIDIA GPU and NVIDIA CDI configured. Create a service and see that is is GPU accelerated and after the first chat request response are fast.

### DIFF
--- a/packages/backend/src/workers/provider/LlamaCppPython.ts
+++ b/packages/backend/src/workers/provider/LlamaCppPython.ts
@@ -35,6 +35,7 @@ import { VMType } from '@shared/src/models/IPodman';
 import type { PodmanConnection } from '../../managers/podmanConnection';
 import type { ConfigurationRegistry } from '../../registries/ConfigurationRegistry';
 import { llamacpp } from '../../assets/inference-images.json';
+import * as fs from 'node:fs';
 
 export const SECOND: number = 1_000_000_000;
 
@@ -135,6 +136,26 @@ export class LlamaCppPython extends InferenceProvider {
             CgroupPermissions: '',
           });
           break;
+        case VMType.UNKNOWN:
+          // This is linux with podman locally installed
+
+          // Linux GPU support currently requires NVIDIA GPU with CDI configured
+          if (!this.isNvidiaCDIConfigured(gpu)) break;
+
+          supported = true;
+          devices.push({
+            PathOnHost: 'nvidia.com/gpu=all',
+            PathInContainer: '',
+            CgroupPermissions: '',
+          });
+
+          user = '0';
+
+          entrypoint = '/usr/bin/sh';
+
+          cmd = ['-c', 'chmod 755 ./run.sh && ./run.sh'];
+
+          break;
       }
 
       // adding gpu capabilities in supported architectures
@@ -224,7 +245,7 @@ export class LlamaCppPython extends InferenceProvider {
     const containerCreateOptions: ContainerCreateOptions = await this.getContainerCreateOptions(
       config,
       imageInfo,
-      connection.vmType as VMType,
+      vmType,
       gpu,
     );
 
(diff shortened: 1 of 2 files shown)


mhdawson:
I'm off on PTO until Dec 23 so if I don't respond to comments/questions on the PR that's why. I check in on the 23rd when I return.

benoitf (approved):
codewise it's 👍 but I have not been able to test it (no GPU on Linux)

mhdawson:
I'm back. @axel7083 anything else I should do to help move this forward?

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -254,8 +275,29 @@ export class LlamaCppPython extends InferenceProvider {
...
+    // Check the known locations for the configuration file
+    const knownLocations = [
+      '/etc/cdi/nvidia.yaml', // Fedora
+    ];
+
+    if (gpu?.vendor !== GPUVendor.NVIDIA) return false;
+
+    let cdiSetup = false;
+    for (const location of knownLocations) {
+      if (fs.existsSync(location)) {
```
axel7083, line 296: This would only work on Linux, which may not be directly an issue, but I would like to have a safeguard to this function.
 
```
If(!env.isLinux) throw new Error('cannot check nvidia CDI on non-Linux platform')
```

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -254,8 +275,29 @@ export class LlamaCppPython extends InferenceProvider {
...
+        cdiSetup = true;
```
axel7083, line 297: Return true

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -254,8 +275,29 @@ export class LlamaCppPython extends InferenceProvider {
...
+        break;
+      }
+    }
+    return cdiSetup;
```
axel7083, line 301: Return false

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -254,8 +275,29 @@ export class LlamaCppPython extends InferenceProvider {
...
         return gpu ? llamacpp.vulkan : llamacpp.default;
       // no GPU support
+      case VMType.UNKNOWN:
+        return this.isNvidiaCDIConfigured(gpu) ? llamacpp.cuda : llamacpp.default;
       default:
         return llamacpp.default;
     }
   }
+
+  protected isNvidiaCDIConfigured(gpu?: IGPUInfo): boolean {
```
axel7083, line 285: You don't need to take the gpu as argument, this function name is not GPU specific, it is to check nvidia CDI, so when called, juste check for known location

mhdawson: @axel7083 it does this check

if (gpu?.vendor !== GPUVendor.NVIDIA) return false;

which also guards againts there being no gpu or gpu accelerating not being enabled.  If we don't do the check in the function we'd have to do it in addition to calling isNvidiaCDIConfigured. I think we could either:

1) change the name of the function
2) do the additional check before calling the function.

I think its less code if we do 1). @axel7083 which would you prefer ?

mhdawson: In terms of a new function name maybe something like `useGpuOnLinux`.  The function is called in two places so I think doing the check inside the shared function does make sense.

axel7083: Let's merge this for now, I re-read after your comment, and this is an acceptable solution

axel7083 (approved):
LGTM, thanks for the feature :rocket:

mhdawson:
@axel7083 based on your last comment my understanding is I don't need to do any updates, let me know if that is not the case. Many thanks for reviewing and helping to get it landed.

slemeur:
Very nice!! Thanks for the contribution @mhdawson

Issue #1693, opened 2024-09-11
Richer reciper catalog UI - Filters


jeffmaury:
### Is your enhancement related to a problem? Please describe

Implement from #1299

### Describe the solution you'd like

Implement from #1299

PR #2297, opened 2024-12-24
feat: add optional language and frameworks fields to recipes

feloy:

### What does this PR do?

Add optional language and frameworks fields to recipes. 

I don't know if we want to increase the catalog version number when adding optional fields. @axel7083 @jeffmaury WDYT?

Related to #1693

### DIFF
--- a/packages/shared/src/models/IRecipe.ts
+++ b/packages/shared/src/models/IRecipe.ts
@@ -57,6 +57,6 @@ export interface Recipe {
    * server the recipe requires
    */
   backend?: string;
-  language?: string;
+  languages?: string[];
   frameworks?: string[];
 }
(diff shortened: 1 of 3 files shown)


jeffmaury (commented):
As those fields are optional, I think we should keep the existing recipe catalog version to favor backward compatibility

Wondering if language should not be multi valued @slemeur WDYT ?

slemeur:
Yeah, I agree that the language could have multiple values. It's highly possible that a recipe might bridge different tools and possibly multiple microservices written with different languages.

packages/shared/src/models/IRecipe.ts
```diff
@@ -57,4 +57,6 @@ export interface Recipe {
    * server the recipe requires
    */
   backend?: string;
+  language?: string;
```
axel7083, line 60: A recipe might be backend / frontend so rust in backend and js in frontend, might be a good idea to have it as an array

gastoner (approved):
Codewise LGTM

axel7083 (approved):
LGTM

Issue #1299, opened 2024-06-26
UX: Richer Recipe Catalog


slemeur:
### Epic domain

This issue is dedicated for the creations of the mockups for building a richer recipe catalog.

## Expand the technologies

Each recipes are focused on a particular use case, example "creating a chatbot" or "building a text summarizer". 
Those recipes could, be build with different tools, frameworks and languages. 

- As a user, I should be able to define the technologies (tools, frameworks and languages) I want to filter the recipes, so I can see only those who are interesting to me.
- As a user, I should be able to see the technologies used by the recipes while I'm browsing the catalog of recipes
- As an end-user, I want to be able to search for recipes that use a specific combination of programming languages and technologies so that I can find complex workflows or integrations.

## Hardware specific

A recipe, might be optimized for a particular hardware architecture. We should be surfacing this to the end-user.
Example: a sample is optimized to be running on Intel infrastructure and as such, use specific Intel components (openvino). 

- As an end-user, I want to know which specific hardware components a recipe uses so that I can ensure compatibility with my existing infrastructure.
- As an end-user, I want to be able to run a recipe optimized for my specific hardware architecture without having to worry about compatibility issues so that I can focus on using the recipe effectively. 
- As an end-user, I want to be able to filter recipes in the catalog by hardware architecture so that I can find recipes optimized for my specific setup.

## Application Patterns

There are more and more application patterns that are being worked on by the community. Those could be related to pre-processing, post-processing, rag, guardrails. 
We should also provide categories of recipes that are covering different application patterns.

## Contributions

We would like to include a way in the recipe catalog to indicate to the end-users and the community, where they can find the catalog, how they can expand it with their own catalog and eventually contribute to the upstream catalog.

ekidneyrh:
Love the detail that is in this ticket! I have a few comments: 

Just wanted to add what the Recipe Catalog looks like at the moment for context.

![Screenshot from 2024-07-24 12-23-40](https://github.com/user-attachments/assets/7daa6105-6916-4607-8436-6b8c6d5227c9)

> As a user, I should be able to define the technologies (tools, frameworks and languages) I want to filter the recipes, so I can see only those who are interesting to me.

How would we filter them? Include a filter option. Preset filters or type search? Maybe both?

> As a user, I should be able to see the technologies used by the recipes while I'm browsing the catalog of recipes

[...]

Issue #1628, opened 2024-08-29
UX: Add a tagging system to recipes cards and metadata


MariaLeonova:
### Is your enhancement related to a problem? Please describe

In order to support filtering and searching for Recipes in the Catalog, they need to have certain metadata items attached to them. Some of these tags will be shown in the UI and others might be hidden. 

### Describe the solution you'd like

Tags added to the UI and metadata. 

### Additional context

This issue is to support the Epic: https://github.com/containers/podman-desktop-extension-ai-lab/issues/1299 
[Doc to collaborate on tags list ](https://docs.google.com/document/d/1l4uyXl0e6CaUu99Bk6i7QGqc8V6KbkimqkzYCEQ26E0/edit?usp=sharing)

slemeur:
I like the proposed inspitation. Another one which can be nice to look at, is how github is handling labels on issues.
The one example from huggingface is making a lot of sense for me, as it allows to have meta-categories of tags and sub-categories. 

I think it would make sense to display the tags by default in the layout of the catalog of recipes - with the option to hide them, if they take too much spaces.

slemeur:
You can check the metadata we have on each of the recipes at the moment in this file: https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/backend/src/assets/ai.json 

The minimum will be to have: 
- Language: what coding langage is used (Java, Python, NodeJS...) 
- Tools: Is it using vllm, llamacpp
- Frameworks: some of our recipes are using llamaindex or solution from this type, we would call this out too.

I think we would show all the tags on each of the recipes. Of course, when there are more tags than what can be displayed, we will need a mecanism to "show more" or expanding the card to view all the tags (can also be a popover).

slemeur:
@MariaLeonova: 

I think it's a good idea to display the labels on the details of the recipe, but that is something we can do in a second state.
In this ticket, we want to have a mockup that shows the recipe's metadata on the list of recipe, similarly to what you suggested in RHDH.

slemeur:
Should be adding tags on the card recipes from this mockup: https://github.com/containers/podman-desktop-extension-ai-lab/issues/1627#issuecomment-2321429148

slemeur:
@MariaLeonova: Thank you for the update. I personally prefer the option 1 as it allows to keep the action button in a consistent place and users are usually expecting the buttons to be in the corners.

- How does it work when there are more than 2 tags to be displayed? 

[...]

PR #2326, opened 2025-01-06
feat: filter recipes

feloy:

### What does this PR do?

Backend provides a method to filter recipes

### Screenshot / video of UI

Backend only

Part of #1299

### How to test this PR?

Unit tests included

### DIFF
--- a/packages/backend/src/managers/catalogManager.spec.ts
+++ b/packages/backend/src/managers/catalogManager.spec.ts
@@ -324,3 +324,145 @@ describe('spy on catalogUtils.sanitize', () => {
     expect(promises.writeFile).toHaveBeenCalled();
   });
 });
+
+test('filter recipes by language', async () => {
+  vi.mocked(existsSync).mockReturnValue(true);
+  vi.mocked(promises.readFile).mockResolvedValue(JSON.stringify(userContent));
+
+  catalogManager.init();
+  await vi.waitUntil(() => catalogManager.getModels().some(model => model.id === 'model1'));
+  const result1 = catalogManager.filterRecipes({
+    languages: ['lang1'],
+  });
+  expect(result1.result.map(r => r.id)).toEqual(['recipe1']);
+  expect(result1.choices).toEqual({
+    languages: ['lang1', 'lang10', 'lang11', 'lang2', 'lang3'],
+    frameworks: ['fw1', 'fw10'],
+    tools: ['tool1'],
+  });
+
+  const result2 = catalogManager.filterRecipes({
+    languages: ['lang2'],
+  });
+  expect(result2.result.map(r => r.id)).toEqual(['recipe2']);
+  expect(result2.choices).toEqual({
+    languages: ['lang1', 'lang10', 'lang11', 'lang2', 'lang3'],
+    frameworks: ['fw10', 'fw2'],
+    tools: ['tool2'],
+  });
+});
+
+test('filter recipes by tool', async () => {
+  vi.mocked(existsSync).mockReturnValue(true);
+  vi.mocked(promises.readFile).mockResolvedValue(JSON.stringify(userContent));
+
+  catalogManager.init();
+  await vi.waitUntil(() => catalogManager.getModels().some(model => model.id === 'model1'));
+
+  const result1 = catalogManager.filterRecipes({
+    tools: ['tool1'],
+  });
+  expect(result1.result.map(r => r.id)).toEqual(['recipe1']);
+  expect(result1.choices).toEqual({
+    frameworks: ['fw1', 'fw10'],
+    languages: ['lang1', 'lang10'],
+    tools: ['tool1', 'tool2', 'tool3'],
+  });
+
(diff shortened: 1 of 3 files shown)


packages/backend/src/managers/catalogManager.spec.ts
```diff
@@ -80,7 +80,7 @@ vi.mock('@podman-desktop/api', async () => {
 let catalogManager: CatalogManager;
 
 beforeEach(async () => {
-  vi.resetAllMocks();
+  vi.restoreAllMocks();
```
benoitf, line 83: question: curious why we need to replace the default reset pattern ?

feloy: With `resetAllMocks`, some `vi.spyOn(...).mockImplementation(...)` in tests were not restored to their initial implementation.

packages/backend/src/managers/catalogManager.ts
```diff
@@ -274,4 +285,68 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
...
+          for (const value of values) {
+            res = [...res, ...result.filter(r => r.languages?.includes(value))];
+          }
+          result = res;
+          break;
+        }
+        case 'tools':
+          result = result.filter(r => values.includes(r.backend ?? ''));
+          break;
+        case 'frameworks': {
```
gastoner, line 304: Cant we merge languages and frameworks (If I see correctly it is the same)?
case 'languages':
case 'frameworks': {
.... break
} case 'tools'...

packages/backend/src/managers/catalogManager.spec.ts
gastoner: I think I've faced this behavior too

packages/backend/src/managers/catalogManager.ts
feloy: They are slightly different, we are filtering on `r.languages` vs `r.frameworks`. I'm not sure I want to generalize before to have more fields to filter on, to be sure to use the good abstraction

packages/backend/src/managers/catalogManager.spec.ts
feloy: OK, I'm adding a code guideline to be sure we are talking about the same thing ->  https://github.com/podman-desktop/podman-desktop/pull/10541

packages/backend/src/managers/catalogManager.ts
```diff
@@ -274,4 +286,68 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
...
+        case 'languages': {
+          let res: Recipe[] = [];
+          for (const value of values) {
+            res = [...res, ...result.filter(r => r.languages?.includes(value))];
+          }
+          result = res;
+          break;
+        }
+        case 'tools':
+          result = result.filter(r => values.includes(r.backend ?? ''));
```
jeffmaury, line 303: Why is backend used ?

packages/backend/src/managers/catalogManager.ts
```diff
@@ -274,4 +286,68 @@ export class CatalogManager extends Publisher<ApplicationCatalog> implements Dis
...
   }
+
+  public filterRecipes(filters: RecipeFilters): FilterRecipesResult {
+    let result = this.getRecipes();
+    for (const [filter, values] of Object.entries(filters)) {
+      switch (filter) {
+        case 'languages': {
+          let res: Recipe[] = [];
+          for (const value of values) {
+            res = [...res, ...result.filter(r => r.languages?.includes(value))];
```
jeffmaury, line 297: why not result = result.filter ?

packages/backend/src/managers/catalogManager.ts
feloy: the goal would be to include backend as a tool (and for now, only the backend would be part of tools)

packages/backend/src/managers/catalogManager.ts
feloy: To get an OR over all the values of a field: for example, if the user selects Javascript and Go in languages,  this will reutrn all the recipes having the lang Javascript OR the lang Go.

gastoner (approved):
LGTM codewise

packages/backend/src/managers/catalogManager.ts
jeffmaury: So we don't filter against the tools field ?

feloy: In https://github.com/containers/podman-desktop-extension-ai-lab/pull/2297/files, I added only `languages` and `frameworks` fields to the catalog.

My intention was to use the `backend` field for the `tools` search. I suppose we can adjust when the filter fields are more clearly defined. For now, the only reference to tools in the spec is at https://github.com/containers/podman-desktop-extension-ai-lab/issues/1628#issuecomment-2507766815, which seems related to the backend only.

jeffmaury (approved):
LGTM

PR #2345, opened 2025-01-08
refactor: extension specs

feloy:

### What does this PR do?
refactor extension specs to follow guideline https://github.com/podman-desktop/podman-desktop/blob/main/CODE-GUIDELINES.md#use-vimocked-not-a-generic-myfunctionmock

### Screenshot / video of UI

Tests only

### How to test this PR?

Run tests

### DIFF
--- a/packages/backend/src/extension.spec.ts
+++ b/packages/backend/src/extension.spec.ts
@@ -21,12 +21,7 @@ import type { ExtensionContext } from '@podman-desktop/api';
 import { activate, deactivate } from './extension';
 import { Studio } from './studio';
 
-vi.mock('./studio', () => {
-  const Studio = vi.fn();
-  Studio.prototype.activate = vi.fn();
-  Studio.prototype.deactivate = vi.fn();
-  return { Studio };
-});
+vi.mock('./studio');
 
 beforeEach(() => {
   vi.clearAllMocks();
--- a/packages/backend/src/extension.spec.ts
+++ b/packages/backend/src/extension.spec.ts
@@ -19,23 +19,17 @@
 import { beforeEach, expect, test, vi } from 'vitest';
 import type { ExtensionContext } from '@podman-desktop/api';
 import { activate, deactivate } from './extension';
+import { Studio } from './studio';
 
-const mocks = vi.hoisted(() => ({
-  studioActivateMock: vi.fn(),
-  studioDeactivateMock: vi.fn(),
-  studioConstructor: vi.fn(),
-}));
-
-vi.mock('./studio', () => ({
-  Studio: mocks.studioConstructor,
-}));
+vi.mock('./studio', () => {
+  const Studio = vi.fn();
+  Studio.prototype.activate = vi.fn();
+  Studio.prototype.deactivate = vi.fn();
+  return { Studio };
+});
 
 beforeEach(() => {
   vi.clearAllMocks();
-  mocks.studioConstructor.mockReturnValue({
-    activate: mocks.studioActivateMock,
-    deactivate: mocks.studioDeactivateMock,
-  });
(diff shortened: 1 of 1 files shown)


packages/backend/src/extension.spec.ts
```diff
@@ -19,23 +19,17 @@
...
-
-vi.mock('./studio', () => ({
-  Studio: mocks.studioConstructor,
-}));
+vi.mock('./studio', () => {
+  const Studio = vi.fn();
+  Studio.prototype.activate = vi.fn();
+  Studio.prototype.deactivate = vi.fn();
+  return { Studio };
+});
```
axel7083, line 29: I would argue we should avoid using the `Class#prototype` maybe something like the following is an alternative

```ts
import { Studio } from './studio';

const STUDIO_MOCK: Studio = {
  activate: vi.fn(),
  deactivate: vi.fn(),
} as unknown as Studio;

vi.mock('./studio', () => ({
  Studio: vi.fn(),
}));

beforeEach(() => {
  vi.resetAllMock();
  vi.mocked(Studio).mockReturnValue(STUDIO_MOCK);
});

test('ensure activate', () => {
  // ....
  
  expect(Studio).toBeCalledTimes(1);
  expect(STUDIO_MOCK.activate).toHaveBeenCalled();
});
```

axel7083: With their solution, it is annoying to do `expect().haveBeenCalledWith(??)` because you have the specify the `this`. 

We would need to use `expect(Studio.prototype.activate).toHaveBeenCalledWith(expect.anything(), ...arguments)` :thinking: .

Moreover this prevent us from mocking multiple value. I cannot do `vi.mocked(studio).mockReturnOnce()` then another..

I understand that them having this in their documentation is a strong argument, but I am not a big fan of playing with Class#prototype :(

feloy:
setting as draft for now, the time to have more discussion on how to mock classes

benoitf:
I don't really see why not following the recommended solutions over multiple places would not be a good solution ?

Almost all samples are using prototype. It's what is behind a TypeScript/es6 classes.


> With their solution, it is annoying to do expect().haveBeenCalledWith(??) because you have the specify the this.
> We would need to use expect(Studio.prototype.activate).toHaveBeenCalledWith(expect.anything(), ...arguments) 🤔 .

I don't see annoyance there. It's as usual. You just do: `expect(Studio.prototype.activate).toHaveBeenCalledWith(... all arguments used to call the activate method...)` no `this` property at all

> Moreover this prevent us from mocking multiple value. I cannot do vi.mocked(studio).mockReturnOnce() then another..
> I understand that them having this in their documentation is a strong argument, but I am not a big fan of playing with Class#prototype :(

I don't see how it's preventing to do things. It's just empty mocks (everything is `vi.fn`) so you can return all values anywhere in all the mocks

if activate was returning a property you can do so the first call will return retValue and the second retValue2

```
  vi.mocked(Studio.prototype.activate).mockResolvedValueOnce('retValue');
  vi.mocked(Studio.prototype.activate).mockResolvedValueOnce('retValue2');
```

benoitf (approved):

packages/backend/src/extension.spec.ts
```diff
@@ -19,23 +19,17 @@
...
 
 beforeEach(() => {
   vi.clearAllMocks();
```
axel7083, line 32: THe clear is weird? should be resetAllMock?

feloy: as we don't redefine any mock, but only count the usage of them, `resetAllMock` is not strictly necessary

axel7083 (approved):
LGTM

Issue #2366, opened 2025-01-13
Start inference server is broken (Windows/WSL)


jeffmaury:
### Bug description

I can't start an inference server against Podman Windows/WSL.

![Image](https://github.com/user-attachments/assets/da7d4add-1073-4ac2-8f22-8f130c8af985)

Seems to be related to 2587bf56ae69f4ec10b8bb5c486027d482fd22c5 as previous commit is working ok

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

axel7083:
Let's deep dive 

For WSL we upload the model to the WSL

https://github.com/containers/podman-desktop-extension-ai-lab/blob/6a229f45f433039136554eb2f8605b4c9651a43c/packages/backend/src/managers/inference/inferenceManager.ts#L212-L222

This is the reasons why we have the path `\home\user\ai-lab\models\granite-3.0-8b-instruct-Q4_K_M.gguf`. Something is going wrong here, as we are suppose to use posix join, since the path is in the WSL and therefore is a linux based.

![Image](https://github.com/user-attachments/assets/cc4b0cf1-5a07-46f2-992e-98b27e3be635)

The provider `LlamaCppPython` or `WhisperCppPython` should not have to deal with platform specific..

PR #2385, opened 2025-01-15
fix(wsl): model mounting

axel7083:
### What does this PR do?

Replace the `getLocalModelFile` by the new `getMountPath`.

The `getMountPath` ensure that **uploaded** model get the path properly joined using posix and not platform specific join.
The `getLocalModelFile` must be used only to get the local model file.

### Screenshot / video of UI

N/A

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2366

### How to test this PR?

- [x] unit tests are covering the changes

### DIFF
--- a/packages/backend/src/utils/modelsUtils.spec.ts
+++ b/packages/backend/src/utils/modelsUtils.spec.ts
@@ -20,12 +20,14 @@ import { process as apiProcess } from '@podman-desktop/api';
 import {
   deleteRemoteModel,
   getLocalModelFile,
+  getMountPath,
   getRemoteModelFile,
   isModelUploaded,
   MACHINE_BASE_FOLDER,
 } from './modelsUtils';
 import type { ModelInfo } from '@shared/src/models/IModelInfo';
 import { getPodmanCli } from './podman';
+import { join, posix } from 'node:path';
 
 vi.mock('@podman-desktop/api', () => {
   return {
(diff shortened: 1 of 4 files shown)


packages/backend/src/utils/modelsUtils.ts
```diff
@@ -20,7 +20,7 @@ import { join, posix } from 'node:path';
 import { getPodmanCli } from './podman';
 import { process } from '@podman-desktop/api';
 
-export const MACHINE_BASE_FOLDER = '/home/user/ai-lab/models/';
+export const MACHINE_BASE_FOLDER = '/home/user/ai-lab/models';
```
axel7083, line 23: This is necessary to remove the trailing slash, to compare with `ModelInfo#file#path` later without having to use magic trick

packages/backend/src/utils/modelsUtils.spec.ts
```diff
@@ -91,7 +135,7 @@ describe('getRemoteModelFile', () => {
       },
     } as unknown as ModelInfo);
 
-    expect(path).toBe(`${MACHINE_BASE_FOLDER}dummy.guff`);
```
axel7083, line 94: Since the trailing slash has been removed, the test was failing, but it was nonsense to concat manually, let's use posix.join has it is the remoteModelFile

gastoner (approved):

jeffmaury (approved):
LGTM

Issue #2410, opened 2025-01-20
E2E scenario for model service with LLM model that send api request


odockal:
Please implement E2E scenario, which in addition to using LLM model, will test the sending of the api request to the model service. Something like "What is the capital of Spain", expecting to get "Madrid" in answer.

Of course, we can think of checking http code, etc.

PR #2449, opened 2025-01-29
chore(test): add e2e test to check model response

cbr7:
### What does this PR do?
Creates e2e test cases to AI lab suite that query the service model in order to validate functionality

https://github.com/containers/podman-desktop-extension-ai-lab/issues/2410


gastoner (commented):
Cant we use different models to avoid test skipping?

cbr7:
@gastoner unfortunately almost all the models are very big, at least a few GB in size so it wouldn't matter, we're already using the smaller models to test.

gastoner (approved):

PR #2456, opened 2025-01-31
chore(test): extend timeout for download and set test to be skipped on windows

cbr7:
### What does this PR do?
Attempts to fix failing e2e test due to timeout expiring during model download.

### DIFF
--- a/tests/playwright/src/ai-lab-extension.spec.ts
+++ b/tests/playwright/src/ai-lab-extension.spec.ts
@@ -78,10 +78,13 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
       extensionsPage = await navigationBar.openExtensions();
       await playExpect(extensionsPage.header).toBeVisible();
     });
+
     test(`Install AI Lab extension`, async () => {
       test.skip(AI_LAB_EXTENSION_PREINSTALLED, 'AI Lab extension is preinstalled');
+      test.setTimeout(120_000);
       await extensionsPage.installExtensionFromOCIImage(AI_LAB_EXTENSION_OCI_IMAGE);
     });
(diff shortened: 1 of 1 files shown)


tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -209,7 +213,7 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
       });
 
       test(`Make POST request to the model service for ${modelName}`, async ({ request }) => {
-        test.skip(modelName === 'ggerganov/whisper.cpp', `Skipping POST request for ${modelName}`);
+        test.skip(isWindows || modelName === 'ggerganov/whisper.cpp', `Skipping POST request for ${modelName}`);
```
benoitf, line 216: question: it looks like this change is not related to the download timeout ?

cbr7: After changing the download timeout I've found there is a second issue after the model is correctly downloaded which causes the test to not work on Windows. I was investigating  This on Friday, however I was not able to fix it yet so I think in the short term the test should be skipped on windows until it can be fixed.

jeffmaury: > question: it looks like this change is not related to the download timeout ?

+1

jeffmaury (commented):
See comment

odockal:
@cbr7 I think that changing the PR title (and commits) could help us out? @benoitf @jeffmaury

cbr7:
I've renamed the PR to better reflect the changes.

odockal (approved):
LGTM

PR #2477, opened 2025-02-04
feat: alow to show model metrics in playground

gastoner:
### What does this PR do?
Shows metrics of a model in playground 

### Screenshot / video of UI

https://github.com/user-attachments/assets/4780fdcd-1c0a-43b6-bc3a-51e69fd5c7c7

### What issues does this PR fix or reference?
Closes #438

PR #2480, opened 2025-02-05
refactor(svelte5): migrated playground component to svelte5

gastoner:
### What does this PR do?
Migrates playground component to svelte5

Required for #2477

### How to test this PR?
Unit tests

### DIFF
--- a/packages/frontend/src/pages/Playground.svelte
+++ b/packages/frontend/src/pages/Playground.svelte
@@ -10,7 +10,6 @@ import {
   isErrorMessage,
 } from '@shared/src/models/IPlaygroundMessage';
 import { catalog } from '../stores/catalog';
-import { afterUpdate } from 'svelte';
 import ContentDetailsLayout from '../lib/ContentDetailsLayout.svelte';
 import RangeInput from '../lib/RangeInput.svelte';
 import Fa from 'svelte-fa';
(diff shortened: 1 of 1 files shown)


packages/frontend/src/pages/Playground.svelte
```diff
@@ -24,24 +23,30 @@ import { router } from 'tinro';
...
-$: messages =
-  conversation?.messages.filter(message => isChatMessage(message)).filter(message => !isSystemPrompt(message)) ?? [];
-$: model = $catalog.models.find(model => model.id === conversation?.modelId);
-$: {
+let conversation = $derived($conversations.find(conversation => conversation.id === playgroundId));
+let messages = $derived(
+  conversation?.messages.filter(message => isChatMessage(message)).filter(message => !isSystemPrompt(message)) ?? [],
+);
+let model = $derived($catalog.models.find(model => model.id === conversation?.modelId));
+$effect(() => {
```
benoitf, line 49: here it looks like it could be a derived state as well as we update `sendEnabled`

so change `let sendEnabled = $state(false);` to be a `derived.by` ?

gastoner: We are assigning to the `setEnabled` which does not let me use the derived

benoitf: hello, you can't replace 

```
let sendEnabled = $state(false);
$effect(...)
```

by

```
let sendEnabled = $derived.by(() => {
  the same content of the existing $effect but just returning the value instead of assigning sendEnabled
   like if (isSystemPrompt(latest) { return true or false}
})
```

?

gastoner: No, I have 
```
let sendEnabled = $derived.by(() => {
  if (conversation?.messages.length) {
    const latest = conversation.messages[conversation.messages.length - 1];
    if (isSystemPrompt(latest) || (isAssistantChat(latest) && !isPendingChat(latest))) {
      return true;
    }
    if (isErrorMessage(latest)) {
      errorMsg = latest.error;
      return true;
    }
  } else {
    return true;
  }
  return false;
});
```
But in the `askPlayground()` function we have `sendEnabled = false;` which says `Cannot assign to derived state`

gastoner: https://github.com/containers/podman-desktop-extension-ai-lab/pull/2480/files#diff-c124817fe6e9029472447102ee97abea44562ed1210e150a9fae0b0da0ce5ccaR53

benoitf: add a inProgress = true/false

then in askPlayground: inProgress = true

enabled = derived.by.....(. if (inProgress) return false)

or something like that

jeffmaury (approved):
LGTM

benoitf (approved):
LGTM (not tested)

gastoner:
I'll wait for one more review, I would say it is ok, but I've did not worked with playground much

axel7083 (approved):
Tested and LGTM 🚀

PR #2518, opened 2025-02-12
feat(state-management): add error handling and service resolution to …

sparesparrow:
…studioAPI

Signed-off-by: sparesparrow <dallheimal@gmail.com>

## Enhanced State Management & Error Handling

This PR establishes a robust foundation for state management and error handling across the application, aligning with the [state management proposal](docs/proposals/state-management.md).

### Core Features

**🛡️ Centralized Error System**
- Track errors with UUIDs, timestamps, and acknowledgement status
- Expose `getErrors()`/`acknowledgeError()` via StudioAPI
- Real-time sync using `MSG_NEW_ERROR_STATE` RPC channel
- Helper methods for error creation/notification ([studio-api-impl.ts:117-126](packages/backend/src/studio-api-impl.ts))

**🔄 Reactive State Infrastructure**
- Generic `Publisher` class with async notification and error propagation
- Abstract `StateManager` for persistence operations
- Standardized error types (`ApplicationStateError`, `ErrorState`)
- 5 new RPC channels for state synchronization

**Key Architectural Improvements**
- Type-safe API extensions for error handling
- Structured error classification via `ApplicationStateErrorType`
- Service metadata resolution via `resolveServiceUri`
- Podman Desktop API type definitions

### State Management Alignment

1. **Backend Authority** - Single source of truth for error state
2. **Reactive Frontend** - Built-in RPC notifications enable Svelte stores
3. **Persistence Safety** - Unified error handling for state operations ([StateManager.ts:40-53](packages/backend/src/utils/StateManager.ts))
4. **Error Lifecycle** - Full cycle from creation to acknowledgement

### Implementation Highlights

**Error Propagation Flow**
```typescript:packages/backend/src/utils/Publisher.ts
protected async notify(): Promise<void> {
  try {
    const data = this.stateGetter();
    await this.webview.postMessage({
      id: this.channel,
      body: data,
    });
  } catch (error) {
    console.error(`Error publishing to ${this.channel}:`, error);
    throw error;
  }
}
```

**State Persistence Guardrails**
```typescript:packages/backend/src/utils/StateManager.ts
protected async persistState(): Promise<void> {
  try {
    await this.notify();
  } catch (err) {
    throw new ApplicationStateError(
      ApplicationStateErrorType.PERSISTENCE_ERROR,
      'Failed to persist state',
      { originalError: err, details: { operation: 'persist' } }
    );
  }
}
```
### How to test this PR?

   - Trigger error conditions by simulating a failure in a state getter or forcing the webview's `postMessage` method to fail.  
   - Verify in the console that errors are logged with messages like “Error publishing to ...” and that they are properly rethrown.


packages/backend/src/utils/Publisher.ts
```diff
@@ -22,17 +22,19 @@ export class Publisher<T> {
...
-        body: this.getter(),
-      })
-      .catch((err: unknown) => {
-        console.error(`Something went wrong while emitting ${this.channel}: ${String(err)}`);
+        body: data,
       });
+    } catch (error) {
+      console.error(`Error publishing to ${this.channel}:`, error);
+      throw error; // Re-throw to allow error handling by caller
+    }
```
axel7083, line 38: At the time when we created the `Publisher` class, we made it so the notify was not a promise, as the caller often do not want to deal with a promise.

The idea is also to consider that the notify cannot really fails, so there is no need to catch the `error(s)`. The rare cases where it could fails would be:

- notify is called too early in the startup of podman-desktop and the webview is not yet defined
- notify is called during the shutdown of podman-desktop and the webview is being destroyed

IMHO those two edge case are not _worth_ the need effort to add a try catch every-time we want to call the `notify` function.

packages/backend/src/utils/StateManager.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  protected async persistState(): Promise<void> {
+    try {
+      await this.notify();
+    } catch (err: unknown) {
+      const details: ApplicationStateErrorDetails = {
+        operation: 'persist',
+        timestamp: Date.now(),
+      };
+      throw new ApplicationStateError(ApplicationStateErrorType.PERSISTENCE_ERROR, 'Failed to persist state', {
```
axel7083, line 48: We don't have the same definition for `persistence` here, we usually use the _persistence_ wording for resources saved after reboot.

Here we just send the information to the frontend (using notify), so there is nothing persisted, if the user open and leave the webview, the resources are lost, they are very volatile.

axel7083 (changes requested):
Hey @sparesparrow !

I am not sure to understand the end goal of this PR, I understand you want to add a better state management, but for what goal?

You linked some proposal document in the PR description, however the link seems to be invalid?

I would love to learn more about the internal goal here, and the reasons you choose to implement this. We could open a dedicated issue/discussion to further discuss the matter you are trying to solve/enhance here.

sparesparrow:
> Hey @sparesparrow !
> 
> I am not sure to understand the end goal of this PR, I understand you want to add a better state management, but for what goal?
> 
> You linked some proposal document in the PR description, however the link seems to be invalid?
> 
> I would love to learn more about the internal goal here, and the reasons you choose to implement this. We could open a dedicated issue/discussion to further discuss the matter you are trying to solve/enhance here.

Hi @axel7083 , thanks for your review. The idea behind this PR was simple: show come coding capabilities; I've recently applied for Software Engineer role in team who owns this repository. Unfortunately, I was not selected for an interview. I've even created my own [podman-desktop extension](https://github.com/sparesparrow/podman-desktop-extension-mcp) - primarily to practice typescript and to prepare in general.

I suggest closing this PR, as it brings no value without refactoring significant part of applicationManager.ts (I didn't commit that because it broke too many tests. And I have as well reconsidered the effort.

PR #2566, opened 2025-02-23
refactor: rework RPC to use interfaces and not abstract classes

benoitf:
### What does this PR do?
RPC is using abstract classes but there is no need to use class definition, we can use only interfaces.
Also use a per interface scoping rather than a per method channel.

Replace the channel constant definition of the channel part of the interfaces and delegate to an helper method.
This helper having generics can also ensure we provide the right data at the right time.

note:
- global noTimeoutChannels is removed to make it part of the getProxy method as it's where the client can decide how to interact.
- invoke is no longer public (it's now a protected method) to enforce the usage of getProxy

### Screenshot / video of UI

image

Long ongoing issue where we are using abstract classes rather than interfaces, allowing to cleanup a lot of `abstract` words reducing the definition

### How to test this PR?

unit tests updated and also added more unit tests (this is why while removing code it's still show


jeffmaury (commented):
I don't see the corresponding issue ? Also seems there is still a type issue

axel7083 (commented):
> Long ongoing issue where we are using abstract classes rather than interfaces, allowing to cleanup a lot of abstract words reducing the definition

On this discussion there was some pros and cons. I like the fact we move the `noTimeoutChannels `closer to the client creation and avoid having it global, thus removing it from the `shared` package, as this is a client aspect.

The cons I am founding here is having the channel IDs moved out of the classes and globally defined. You created a generic `RpcChannel<T>` class, but this class has no type protection, you could provide anything you want on the client then call it, and it would fails. 

The idea of having an abstract class was to be able to tied the `CHANNEL` property, and above all: prevent developer to make a mistake, the signature was having a lot of constraint to ensure that.

https://github.com/containers/podman-desktop-extension-ai-lab/blob/4b3d39e3fec564c334b2ce0c7db5094f062bb45c/packages/shared/src/messages/MessageProxy.ts#L177

Here the `CHANNEL` must be tied to the prototype of the object you are trying to get. You cannot get a proxy for something that is not meant to be proxied, this force the developer to be explicit.

Here, we lose that and that's the point that I am not a big fan of...  Developer could provide the wrong channel on the client, use the wrong interface etc.

The current definition is not even enough typed in my opinion, when writing this signature we should have added the following constraint to match the `registerInstance`.

```ts
getProxy<T extends Record<keyof T, UnaryRPC>>(
	classType: { CHANNEL: string; prototype: T }
): T
```

If we really want to go this way (I am not opposed to found solution to improve the code) we could try to stronger the `T` for example at a start, and found a way to help developer avoiding making mistake.

```
createRpcChannel<T extends Record<keyof T, UnaryRPC>>()
```

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -52,9 +51,16 @@ export function isMessageResponse(content: unknown): content is IMessageResponse
   return isMessageRequest(content) && 'status' in content;
 }
 
+// instance has methods that are callable
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type ObjectInstance<T> = {
+  [key: string]: (...args: unknown[]) => unknown;
```
axel7083, line 57:
```suggestion
  [key: string]: (...args: unknown[]) => Promise<unknown>;
```

Can we force Promise return type?

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -174,25 +167,33 @@ export class RpcBrowser {
     });
   }
 
-  getProxy<T>(classType: { CHANNEL: string; prototype: T }): T {
-    // eslint-disable-next-line @typescript-eslint/no-this-alias
-    const thisRef = this;
+  getProxy<T>(channel: RpcChannel<T>, options?: { noTimeoutMethods: Array<keyof T> }): T {
```
axel7083, line 170: if we says T must extend ObjectInstance, does it automatically infers keyof T as string?

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -203,12 +204,13 @@ export class RpcBrowser {
...
-      channel: channel,
-      args: args,
+      channel,
+      method,
+      args,
     } as IMessageRequest);
 
     // Add some timeout
-    if (!noTimeoutChannels.includes(channel)) {
+    if (Array.isArray(noTimeoutMethodsValues) && !noTimeoutMethodsValues.includes(method)) {
```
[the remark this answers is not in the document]
benoitf: it seems some unit tests are still calling invoke method somehow and they don't comply with the signature 🤦

packages/shared/src/messages/MessageProxy.ts
```diff
@@ -235,3 +237,20 @@ export class RpcBrowser {
     return !!content && 'id' in content && 'body' in content && this.subscribers.has(content.id);
   }
 }
+
+// identifier for a given interface
+export class RpcChannel<T> {
+  // variable used to use the marker interface T
+  protected _marker: T | undefined;
+
+  constructor(private channel: string) {}
```
axel7083, line 246:
```suggestion
  constructor(private readonly channel: string) {}
```

packages/shared/src/messages/MessageProxy.ts
benoitf: done

axel7083 (approved):

benoitf:
> I don't see the corresponding issue ? Also seems there is still a type issue

@jeffmaury 
https://github.com/podman-desktop/podman-desktop/issues/11326

And all checks are 💚 now

feloy (approved):

Issue #2695, opened 2025-03-12
move open-ai swaggerUI to AI Lab


benoitf:
### Is your enhancement related to a problem? Please describe

Today, we redirect people to the openapi-ui/swagger ui of llama.cpp python

but for example llama.cpp does not provide the openapi UI

### Describe the solution you'd like

Make AI Lab providing the webapp so it can work with any backend/OCI image even if the image is not providing such UI

### Describe alternatives you've considered

remove the OpenAPI UI

### Additional context

switching from llama cpp python to llama cpp would remove this app

PR #2703, opened 2025-03-13
feat: make AI lab providing swagger UI for the openAI docs

benoitf:
### What does this PR do?
inference OCI images may not have Swagger UI, so start the swagger UI in Podman AI Lab and make calls from there to the inference server

fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2695

### How to test this PR?

start a new playground for example, then go to the services list, click on "swagger documentation"
you'll notice you're using Swagger UI and you can query the inference server URL from there

so swagger UI is served by AI Lab

unit test added

### DIFF
--- a/packages/backend/src/workers/provider/LlamaCppPython.ts
+++ b/packages/backend/src/workers/provider/LlamaCppPython.ts
@@ -179,8 +179,13 @@ export class LlamaCppPython extends InferenceProvider {
       }
     }
 
+    // add the link to our openAPI instance using the instance as the host
+    const aiLabPort = this.configurationRegistry.getExtensionConfiguration().apiPort;
+    // add in the URL the port of the inference server
+    const aiLabDocsLink = `http://localhost:${aiLabPort}/api-docs/${config.port}`;
+    console.log('set the ai lab docs link to', aiLabDocsLink);
     // adding labels to inference server
-    labels['docs'] = `http://localhost:${config.port}/docs`;
+    labels['docs'] = aiLabDocsLink;
     labels['api'] = `http://localhost:${config.port}/v1`;
 
     return {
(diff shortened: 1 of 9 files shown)


packages/backend/vite.config.js
```diff
@@ -16,8 +16,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import {join} from 'path';
-import {builtinModules} from 'module';
+import { join } from 'path';
+import { builtinModules } from 'module';
```
benoitf, line 20:
```suggestion
import {join} from 'path';
import {builtinModules} from 'module';
```

feloy (approved):
works great, can access the doc and send requests

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -179,8 +179,13 @@ export class LlamaCppPython extends InferenceProvider {
       }
     }
 
+    // add the link to our openAPI instance using the instance as the host
+    const aiLabPort = this.configurationRegistry.getExtensionConfiguration().apiPort;
+    // add in the URL the port of the inference server
+    const aiLabDocsLink = `http://localhost:${aiLabPort}/api-docs/${config.port}`;
+    console.log('set the ai lab docs link to', aiLabDocsLink);
```
jeffmaury, line 186: Remove ?

jeffmaury (approved):
Simple remark: if the API server is not started, I think we should not expose a link in the UI at all.

All I would love to see the model name displayed in the Swagger UI rather that the generic "AI Lab Inference Server"

benoitf:
> Simple remark: if the API server is not started, I think we should not expose a link in the UI at all.

in which case the server is not started ?

> All I would love to see the model name displayed in the Swagger UI rather that the generic "AI Lab Inference Server"

it could be another path parameter, but I would call it improvement as it's not the case today

packages/backend/src/workers/provider/LlamaCppPython.ts
benoitf, line 186:
```suggestion
```

PR #2735, opened 2025-03-20
feat: make AI lab providing swagger UI for the openAI docs (#2703)

benoitf:
### What does this PR do?
introduce again the swagger-ui served by ApiServer.
This time, copy the swagger-ui-dist content to the vite dist folder so files will be found (so no need to exclude swagger-ui-express from vite)

fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2695

### How to test this PR?

try dev mode and production mode (create the image, push it to a OCI registry and install extension from there)

go to services/service and click on 'Swagger link' and try it

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,9 @@
   "dependencies": {
     "js-yaml": "^4.1.0"
   },
+  "scarfSettings": {
+    "enabled": false
+  },
   "pnpm": {
     "overrides": {
       "postman-collection>semver": "^7.5.2"
--- a/packages/backend/src/managers/apiServer.spec.ts
+++ b/packages/backend/src/managers/apiServer.spec.ts
@@ -220,6 +220,7 @@ test('/api-docs/9000 returns swagger UI', async () => {
   }
   const response = await request(listener).get('/api-docs/9000/').expect(200);
   expect(response.status).toBe(200);
+
   // Ensure it returns the Swagger UI page
   expect(response.text).toContain('<title>Swagger UI</title>');
 });
--- a/packages/backend/src/managers/apiServer.spec.ts
+++ b/packages/backend/src/managers/apiServer.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2024 Red Hat, Inc.
+ * Copyright (C) 2024-2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,7 @@
 
 /* eslint-disable sonarjs/no-nested-functions */
 
-import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { afterEach, assert, beforeEach, describe, expect, test, vi } from 'vitest';
 import { ApiServer, PREFERENCE_RANDOM_PORT } from './apiServer';
 import request from 'supertest';
 import type * as podmanDesktopApi from '@podman-desktop/api';
@@ -209,6 +209,19 @@ test('/api/tags returns ok', async () => {
     name: 'model-name',
     model: 'model-name',
   });
+});
+
+test('/api-docs/9000 returns swagger UI', async () => {
+  expect(server.getListener()).toBeDefined();
+  vi.mocked(modelsManager.getModelsInfo).mockRejectedValue({});
+  const listener = server.getListener();
+  if (!listener) {
+    assert.fail('listener is not defined');
+  }
+  const response = await request(listener).get('/api-docs/9000/').expect(200);
+  expect(response.status).toBe(200);
+  // Ensure it returns the Swagger UI page
+  expect(response.text).toContain('<title>Swagger UI</title>');
 });
 
 test('verify listening on localhost', async () => {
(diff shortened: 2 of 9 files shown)


benoitf:
will fix the conflict...

benoitf:
rebased

packages/backend/src/managers/apiServer.ts
```diff
@@ -144,6 +151,31 @@ export class ApiServer implements Disposable {
     app.use('/api', router);
     app.use('/spec', this.getSpec.bind(this));
 
+    const options = {};
```
axel7083, line 154: why declaring options here? if we don't do put anything in it, the option is optional 

```ts
serveFiles(swaggerDoc?: JsonObject, opts?: SwaggerUiOptions): RequestHandler[];
```

let's remove it

packages/backend/src/workers/provider/LlamaCppPython.ts
```diff
@@ -179,8 +179,12 @@ export class LlamaCppPython extends InferenceProvider {
       }
     }
 
+    // add the link to our openAPI instance using the instance as the host
+    const aiLabPort = this.configurationRegistry.getExtensionConfiguration().apiPort;
+    // add in the URL the port of the inference server
+    const aiLabDocsLink = `http://localhost:${aiLabPort}/api-docs/${config.port}`;
     // adding labels to inference server
-    labels['docs'] = `http://localhost:${config.port}/docs`;
+    labels['docs'] = aiLabDocsLink;
```
axel7083, line 187: why not inline?

axel7083: maybe this should be in a dedicated function, like `getDocumentionUrl(inferencePort: number)` or something like that

packages/backend/vite.config.js
```diff
@@ -53,6 +55,35 @@ const config = {
...
+            // Copy files
+            const entries = await readdir(source, { withFileTypes: true });
+
+            for (const entry of entries) {
+              const srcFile = join(source, entry.name);
+              const destFile = join(destination, entry.name);
+
+              if (entry.isFile()) {
+                await copyFile(srcFile, destFile);
+              }
```
axel7083, line 81: why coping only the files ? can't we just copy the full folder using [fs.promises.cp](https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options) and use recursive option?

packages/backend/src/managers/apiServer.spec.ts
```diff
@@ -211,6 +211,19 @@ test('/api/tags returns ok', async () => {
   });
 });
 
+test('/api-docs/9000 returns swagger UI', async () => {
+  expect(server.getListener()).toBeDefined();
+  vi.mocked(modelsManager.getModelsInfo).mockRejectedValue({});
+  const listener = server.getListener();
+  if (!listener) {
```
axel7083, line 218: not related to this PR, but always feel weird than a **get** return something potentially undefined

package.json
```diff
@@ -84,6 +84,9 @@
   "dependencies": {
     "js-yaml": "^4.1.0"
   },
+  "scarfSettings": {
+    "enabled": false
+  },
```
axel7083, line 89: what this settings is about?

axel7083 (commented):
Working in dev mode :+1: on linux

jeffmaury (commented):
Not working for me: when I click on the docs link, I have a blank window:

image

I did:
- `pnpm install`
- `pnpm build`
- `podman build . -t quay.io/jeffmaury/ailab`
- `podman push quay.io/jeffmaury/ailab`

jeffmaury:
> Not working for me: when I click on the docs link, I have a blank window:
> 
> image
> 
> I did:
> 
>     * `pnpm install`
> 
>     * `pnpm build`
> 
>     * `podman build . -t quay.io/jeffmaury/ailab`
> 
>     * `podman push quay.io/jeffmaury/ailab`

As my dist folder did not contain the `swagger-ui` folder I changed the `emptyOutDir` to `false` in `packages/backend/vite.config.js` and folder is now there

benoitf:
@jeffmaury so your dist folder is never containing the swagger directory but the extension.cjs ?

packages/backend/vite.config.js
benoitf: different attempt

I switched to copy and added a filter to not copy the .map files

packages/backend/src/managers/apiServer.spec.ts
```diff
@@ -211,6 +211,19 @@ test('/api/tags returns ok', async () => {
   });
 });
 
+test('/api-docs/9000 returns swagger UI', async () => {
```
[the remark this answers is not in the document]
benoitf: the issue is that AFAIK it can't be handle by the unit test. It looks more what the e2e test should do while trying to click on the links.

From the unit test, the only thing I receive is raw HTML and .js files

benoitf:
@jeffmaury I've added emptyOutFolder to false

feloy (approved):
LGTM
Works on mac with `pnpm watch` and with an install from an OCI image

PR #2737, opened 2025-03-20
feat: add HuggingFace model handler

### LINKED ISSUE #2734: Add support for HuggingFace models
### Is your feature request related to a problem? Please describe

Add a way to add HuggingFace models using a specially crafted URL and provide the implementation to manage them

Implement a new ModelHandler for HuggingFace

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #2734

### What does this PR do?

Add a new model handler for downloaded models from HuggingFace

#2734 

### How to test this PR?

Add the following snippet in your user catalog:

```
{
    "id": "openvino-mistral-7b-instruct-v0.2-int8-ov",
    "name": "openvino-mistral-7b-instruct-v0.2-int8-ov",
    "description": "OpenVINO/mistral-7b-instruct-v0.2-int8-ov",
    "url": "huggingface:/OpenVINO/mistral-7b-instruct-v0.2-int8-ov",
    "backend": "openvino"
}
```

- Verify it appears in the model list
- Verify it can be downloaded/deleted
- Verify an inference server cannot be started

### DIFF
--- a/packages/backend/src/models/HuggingFaceModelHandler.spec.ts
+++ b/packages/backend/src/models/HuggingFaceModelHandler.spec.ts
@@ -1,0 +1,156 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import { EventEmitter } from '@podman-desktop/api';
+import type { TelemetryLogger, Webview } from '@podman-desktop/api';
+import { beforeEach, expect, test, vi } from 'vitest';
+import { ModelsManager } from '../managers/modelsManager';
+import type { CatalogManager } from '../managers/catalogManager';
+import type { ModelInfo } from '@shared/src/models/IModelInfo';
+import { TaskRegistry } from '../registries/TaskRegistry';
+import type { CancellationTokenRegistry } from '../registries/CancellationTokenRegistry';
+import type { PodmanConnection } from '../managers/podmanConnection';
+import type { ConfigurationRegistry } from '../registries/ConfigurationRegistry';
+import { ModelHandlerRegistry } from '../registries/ModelHandlerRegistry';
+import { HuggingFaceModelHandler } from './HuggingFaceModelHandler';
+import { snapshotDownload } from '@huggingface/hub';
+
+vi.mock('@podman-desktop/api', () => {
+  return {
+    EventEmitter: vi.fn(),
+  };
+});
+
+vi.mock('@huggingface/hub', () => {
+  return {
+    scanCacheDir: vi.fn(),
+    snapshotDownload: vi.fn(),
+  };
+});
+
+const WEBVIEW_MOCK = {
+  postMessage: vi.fn(),
+} as unknown as Webview;
+
+const catalogManagerMock = {
+  getModels(): ModelInfo[] {
+    return [
+      { id: 'model-id-1', name: 'model-id-1-model' } as ModelInfo,
+      { id: 'model-id-2', name: 'model-id-2-model' } as ModelInfo,
+    ];
+  },
+  onUpdate: vi.fn(),
+} as unknown as CatalogManager;
+
+const telemetryLogger = {
+  logUsage: vi.fn(),
+  logError: vi.fn(),
+} as unknown as TelemetryLogger;
+
+const taskRegistry: TaskRegistry = new TaskRegistry(WEBVIEW_MOCK);
+
+const cancellationTokenRegistryMock = {
+  createCancellationTokenSource: vi.fn(),
+} as unknown as CancellationTokenRegistry;
+
+const podmanConnectionMock = {
+  getContainerProviderConnections: vi.fn(),
+} as unknown as PodmanConnection;
+
+const configurationRegistryMock = {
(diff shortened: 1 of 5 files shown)


packages/backend/src/models/HuggingFaceModelHandler.ts
```diff
@@ -0,0 +1,143 @@
...
+  createDownloader(model: ModelInfo, abortSignal: AbortSignal): Downloader {
+    const result = parseURL(model.url!);
+    if (result) {
+      return new HuggingFaceDownloader(model.url!, result.repo, result.revision, abortSignal);
+    }
+    throw new Error(`Invalid URL: ${model.url} for model ${model.name}`);
+  }
+
+  async deleteModel(model: ModelInfo): Promise<void> {
+    if (model.file) {
```
axel7083, line 105: we should throw an error if we try to delete something that does not exists

packages/backend/src/models/HuggingFaceModelHandler.ts
```diff
@@ -0,0 +1,143 @@
...
+      await fs.rm(model.file?.path, { recursive: true });
+    }
+  }
+
+  dispose(): void {}
```
axel7083, line 110: need to be disposable? if yes, we should implement the Disposable interface for the class

packages/backend/src/models/HuggingFaceModelHandler.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  async getLocalModelsFromDisk(): Promise<void> {
```
axel7083, line 112: getLocalModelsFromDisk return nothing, so probably should be renamed to reflect its purpose (scan, collect etc.) or return the models

packages/backend/src/models/HuggingFaceModelHandler.ts
```diff
@@ -0,0 +1,143 @@
...
+    const hfModels = this.modelsManager
+      .getModelsInfo()
+      .filter(model => model.url && this.accept(model.url))
+      .map(model => {
+        return { model: model, repo: parseURL(model.url!) };
+      })
+      .filter(info => info.repo);
+
+    scanCacheDir()
```
axel7083, line 121: await with try catch instead of using `.then`?

packages/backend/src/models/HuggingFaceModelHandler.ts
```diff
@@ -0,0 +1,143 @@
...
+      .then(hfinfo => {
+        for (const repo of hfinfo.repos) {
+          for (const revision of repo.revisions) {
+            for (const ref of revision.refs) {
+              const model = hfModels.find(m => m.repo?.repo === repo.id.name && m.repo?.revision === ref);
```
axel7083, line 126: I am not sure to understand, do we filter the models based on our catalog? why ignoring user models?

packages/backend/src/models/HuggingFaceModelHandler.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+export class HuggingFaceModelHandler extends ModelHandler {
+  constructor(modelsManager: ModelsManager) {
+    super('huggingface model registry', modelsManager);
+  }
+
+  accept(url: string): boolean {
+    return url.startsWith('huggingface') || url.startsWith('hf');
+  }
```
axel7083, line 94: Can we add a comment telling which can of URL are accepted ?

For example `https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503` would be failing.

We may need to use something like
```suggestion
  accept(url: string): boolean {
    const url = new URL("https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503");
    return url.hostname === "huggingface"
  }
```

jeffmaury: The idea is that ModelHandler are selected based on the URL they accept and I followed the URL provided by ramalama. Supports of https prefix would introduce ModelHandler ordering

packages/backend/src/models/HuggingFaceModelHandler.ts
jeffmaury: Not related to this PR as it's from the ModelHandler interface. Follow up issue

packages/backend/src/models/HuggingFaceModelHandler.ts
jeffmaury: ModelHandler does that

packages/backend/src/models/HuggingFaceModelHandler.ts
jeffmaury: We are not ignoring user models: we are scanning the local HF cache and trying to find a corresponding model in the catalog: if found then we update the metadata (just like in the other model handler)

packages/backend/src/models/HuggingFaceModelHandler.ts
jeffmaury: The reason why I choose this pattern is that it will decrease startup time as those method are called on startup

jeffmaury:
> Thanks for comments, LGTM 🚀

Can you approve ?

packages/backend/src/models/HuggingFaceModelHandler.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  vi.mocked(EventEmitter).mockReturnValue({
+    event: vi.fn().mockImplementation(callback => {
+      listeners.push(callback);
+    }),
+    fire: vi.fn().mockImplementation((content: unknown) => {
+      listeners.forEach(listener => listener(content));
+    }),
+  } as unknown as EventEmitter<unknown>);
+});
```
feloy, line 105: not a blocker,
using vi.fn().mockImplementation loses the typing.

You may do instead smth like:
```
const eventReturned = {
  event: vi.fn(),
  fire: vi.fn(),
};

vi.mocked(EventEmitter).mockReturnValue(eventReturned);

vi.mocked(eventReturned.event).mockImplementation(...);
vi.mocked(eventReturned.fire).mockImplementation(...);
```

feloy (approved):

packages/backend/src/models/HuggingFaceModelHandler.spec.ts
jeffmaury: I think I copied it from elsewhere but will update

axel7083 (approved):
Forgot to approve oupsi

PR #2753, opened 2025-03-21
chore: check that we can't create multiple time a RPCChannel with same name

benoitf:
### What does this PR do?
ensure we can't create twice a RPC Channel with the same name (else we got an error)

follow-up of https://github.com/containers/podman-desktop-extension-ai-lab/pull/2596#discussion_r2007611049


axel7083 (approved):
LGTM, thanks for adding this :rocket:

PR #2763, opened 2025-03-25
fix: use null coalescing operator

jeffmaury:
fix: use null coalescing operator


packages/frontend/src/stores/rpcReadable.ts
```diff
@@ -65,16 +65,14 @@ export function RPCReadable<T>(
...
-          timeoutThrottle = setTimeout(() => {
-            if (timeoutId) {
-              clearTimeout(timeoutId);
-              timeoutId = undefined;
-            }
-            resolve(func());
-          }, 5000);
-        }
+        // eslint-disable-next-line sonarjs/no-nested-functions
+        timeoutThrottle ??= setTimeout(() => {
```
axel7083, line 69: just learned new JS synthax!

Issue #2628, opened 2025-03-04
Provide the ability to simply start a llama-stack from Podman AI Lab


slemeur:
### Is your feature request related to a problem? Please describe

As a very basic and preliminary work, we would provide a new tab "Llama Stack" with a screen enabling to simply start a llama-stack from Podman AI Lab and choosing a specific model

### Describe the solution you'd like

As a very basic and preliminary work, we would provide a new tab "Llama Stack" with a screen enabling to simply start a llama-stack from Podman AI Lab and choosing a specific model

The UI should be similar to what has been done to start the InstructLab container. So if the container is not already started, the steps are:

- start the container for the Podman AI Lab distribution providing the PODMAN_AI_LAB_URL environment variable
- once the container is started, using the llama stack typescript client register the models from the catalog
- switch to container is already started

if the container is already started, the UI should offer a link of the Swagger/OpenAPI file and client so that users can experiment with the llama-stack APIs
Also there should be a text describing the different llama-stack APIs available on this distribution:

- inference
- guard
- safety
- vectors dbs

slemeur:
excellent!

MariaLeonova:
Where would this tab be located in the navigation? Would it be a new section in AI Apps?

PR #2831, opened 2025-04-09
feat: create a Llama stack page

feloy:

### What does this PR do?

Create a Llama stack page where the user can start a llama stack.

For the moment, the user needs to register the model (see demo)

### Screenshot / video of UI

https://github.com/user-attachments/assets/cba1a4d5-b108-4717-a45a-3c796d9c69d9

First step of #2628


packages/backend/src/managers/llama-stack/llamaStackManager.spec.ts
```diff
@@ -0,0 +1,180 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
```
axel7083, line 2:
```suggestion
 * Copyright (C) 2025 Red Hat, Inc.
```

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+    private taskRegistry: TaskRegistry,
+    private podmanConnection: PodmanConnection,
+    private containerRegistry: ContainerRegistry,
+    private configurationRegistry: ConfigurationRegistry,
+    private telemetryLogger: TelemetryLogger,
+  ) {
+    this.#initialized = false;
+    this.podmanConnection.onPodmanConnectionEvent(this.watchMachineEvent.bind(this));
+    this.containerRegistry.onStartContainerEvent(this.onStartContainerEvent.bind(this));
+    this.containerRegistry.onStopContainerEvent(this.onStopContainerEvent.bind(this));
```
axel7083, line 58: feels like this should be in an init, moreover we should probably have LlamaStackManager implementing Disposable, and dispose those events in it.

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (!this.#initialized) {
+      const containers = await containerEngine.listContainers();
+      this.#containerInfo = containers
+        .filter(c => c.State === 'running' && c.Labels && LLAMA_STACK_CONTAINER_LABEL in c.Labels)
+        .map(c => ({
+          containerId: c.Id,
+          port: parseInt(c.Labels[LLAMA_STACK_API_PORT_LABEL]),
+        }))
+        .at(0);
+      this.#initialized = true;
```
axel7083, line 105: feels weird to have a `get*` method setting `initialized` probably need to be renamed to something like `init` or `setup`, or `config` etc.

Moreover, I have some issue understanding the logic, could you add some comments to the function? thanks

Last: `get*` method should not return possibly undefined, feel more natural to use find

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+  private async refreshLlamaStackContainer(id?: string): Promise<void> {
+    const containers = await containerEngine.listContainers();
+    const containerInfo = containers
+      .filter(c => !id || c.Id === id)
+      .filter(c => c.State === 'running' && c.Labels && LLAMA_STACK_CONTAINER_LABEL in c.Labels)
+      .at(0);
+    if ((id && containerInfo) || !id) {
+      this.#containerInfo = containerInfo
+        ? {
+            containerId: containerInfo.Id,
```
axel7083, line 70: shound't we also get the engineId? the containerId alone does not allow us to use `stop`, `start` etc.

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+          containerId: c.Id,
+          port: parseInt(c.Labels[LLAMA_STACK_API_PORT_LABEL]),
+        }))
+        .at(0);
+      this.#initialized = true;
+    }
+    return this.#containerInfo;
+  }
+
+  async requestCreateLlamaStackContainer(config: LlamaStackContainerConfiguration): Promise<void> {
```
axel7083, line 110: Feel weird to have an hardcoded trackingId (`LLAMA_STACK_CONTAINER_TRACKINGID`) when we usually return Promise<string> for `request*` method

https://github.com/containers/podman-desktop-extension-ai-lab/blob/878854ad6ddcc8d8551404b11fd124f595b81e96/packages/backend/src/managers/inference/inferenceManager.ts#L131

what happen if I call it once, delete the container, call it again (without restarting the extension) the tracking Id would be in multiple tasks

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+    });
+
+    let connection: ContainerProviderConnection | undefined;
+    if (config.connection) {
+      connection = this.podmanConnection.getContainerProviderConnection(config.connection);
+    } else {
+      connection = this.podmanConnection.findRunningContainerProviderConnection();
+    }
+
+    if (!connection) throw new Error('cannot find running container provider connection');
```
axel7083, line 129: feels like this should be above the createTask, as this does not make the task fails, and just make the all process running forever

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+        UsernsMode: 'keep-id:uid=0,gid=0',
+        PortBindings: {
+          '8321/tcp': [
+            {
+              HostPort: `${llamaStackApiPort}`,
+            },
+          ],
+        },
+      },
+      Env: [`PODMAN_AI_LAB_URL=http://host.docker.internal:${aiLabApiPort}`],
```
axel7083, line 222: shouldn't we be using `host.containers.internal`?

https://github.com/containers/podman-desktop-extension-ai-lab/blob/878854ad6ddcc8d8551404b11fd124f595b81e96/packages/backend/src/managers/application/applicationManager.spec.ts#L408

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+      throw err;
+    } finally {
+      this.taskRegistry.updateTask(containerTask);
+    }
+  }
+
+  private async getLlamaStackContainerFolder(): Promise<string> {
+    const llamaStackPath = path.join(this.appUserDirectory, 'llama-stack', 'container');
+    await fs.mkdir(llamaStackPath, { recursive: true });
+    await fs.mkdir(path.join(llamaStackPath, '.llama'), { recursive: true });
```
axel7083, line 248:
```suggestion
    await fs.mkdir(path.join(llamaStackPath, '.llama'), { recursive: true });
```

not needed to make the parent first, as we use recursive

packages/backend/src/managers/llama-stack/llamaStackManager.ts
```diff
@@ -0,0 +1,251 @@
...
+        [LLAMA_STACK_CONTAINER_LABEL]: image,
+        [LLAMA_STACK_API_PORT_LABEL]: `${llamaStackApiPort}`,
+      },
+      HostConfig: {
+        AutoRemove: true,
+        SecurityOpt: [DISABLE_SELINUX_LABEL_SECURITY_OPTION],
+        Mounts: [
+          {
+            Target: '/app/.llama',
+            Source: path.join(folder, '.llama'),
```
axel7083, line 209: You seems to only use `getLlamaStackContainerFolder` once (L191) and you join the path to `.llama`

I don't understand why not just returning the `.llama`? why do we need the parent folder?

packages/frontend/src/pages/llama-stack/StartLlamaStackContainer.spec.ts
```diff
@@ -0,0 +1,123 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
+ *
```
feloy, line 3:
```suggestion
 * Copyright (C) 2025 Red Hat, Inc.
 *
```

packages/backend/src/managers/llama-stack/llamaStackManager.ts
feloy: I added some comment. 

I'm trying to be consistent with `instructlabManager`. 

For me, setting initialized if for caching the result, seems legit in a `get` method

packages/backend/src/managers/llama-stack/llamaStackManager.ts
feloy: For the moment, it is not necessary (we just start the container, and access its details page). Let's add it when necessary

packages/backend/src/managers/llama-stack/llamaStackManager.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+  llamaStackManager = new LlamaStackManager(
+    '',
+    taskRegistry,
+    podmanConnection,
+    containerRegistry,
+    configurationRegistry,
+    telemetryMock,
+  );
+  llamaStackManager.init();
+  taskRegistry.deleteByLabels({ trackingId: LLAMA_STACK_CONTAINER_TRACKINGID });
```
axel7083, line 80: I would rather using `taskRegistry.dispose()` as here we could have side effect between tests (made a PR https://github.com/containers/podman-desktop-extension-ai-lab/pull/2838)

I would even say we should probably re-create both the taskRegistry and the llamaStackManager in the beforeEach.

I would not include the init in the beforeEach, as you don't test that the init is doing what we expect (registering events listener)

packages/backend/src/managers/llama-stack/llamaStackManager.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+  });
+  vi.mocked(containerEngine.listImages).mockRejectedValue(new Error());
+  await llamaStackManager.requestCreateLlamaStackContainer({});
+  const tasks = await waitTasks(LLAMA_STACK_CONTAINER_TRACKINGID, 2);
+  expect(tasks.some(task => task.state === 'error')).toBeTruthy();
+});
+
+test('requestCreateLlamaStackContainer returns id and error if listImage returns image', async () => {
+  vi.mocked(podmanConnection.findRunningContainerProviderConnection).mockReturnValue({
+    name: 'Podman Machine',
```
axel7083, line 143: to avoid repetition in all tests, please declare a constant in the top of the file

https://github.com/containers/podman-desktop-extension-ai-lab/blob/e34d59f8e3ab8ed84fb154827dbcf46313d07de5/packages/backend/src/workers/provider/LlamaCppPython.spec.ts#L69

(if you could use only uppercase for it,  that would be even better 🙏 )

packages/backend/src/managers/llama-stack/llamaStackManager.ts
jeffmaury: That is also done in InstructLab I do,'t recall the reason but tasks are cleared once the container is started and that is probably related to the fact that we have a single container to start

jeffmaury (commented):
It is not working for me on WSL because host.containers.internal allows to connect to the WSL machine but not the the host where the REST server is running.

Possible solutions: 
- start a relay on the host
- change the REST server so that is can accept request from WSL (there is a virtual network card that we can use to get the IP address)

feloy:
> It is not working for me on WSL because host.containers.internal allows to connect to the WSL machine but not the the host where the REST server is running.
> 
> Possible solutions:
> 
> * start a relay on the host
> * change the REST server so that is can accept request from WSL (there is a virtual network card that we can use to get the IP address)

Can it be part of a follow-up PR ?

jeffmaury:
> > It is not working for me on WSL because host.containers.internal allows to connect to the WSL machine but not the the host where the REST server is running.
> > Possible solutions:
> > 
> > * start a relay on the host
> > * change the REST server so that is can accept request from WSL (there is a virtual network card that we can use to get the IP address)
> 
> Can it be part of a follow-up PR ?

Yes and let assign it to me

feloy:
@axel7083 @jeffmaury I refactored llamaStackManager to use the TaskRunner, feel free to review again

packages/backend/src/managers/llama-stack/llamaStackManager.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+  vi.spyOn(taskRegistry, 'deleteByLabels');
+  vi.mocked(containerEngine.listContainers).mockResolvedValueOnce([LLAMA_STACK_CONTAINER_RUNNING]);
+  vi.mocked(containerRegistry.onStartContainerEvent).mockImplementation(f => {
+    f({
+      id: 'dummyId',
+    });
+    return NO_OP_DISPOSABLE;
+  });
+  vi.mocked(containerRegistry.onStopContainerEvent).mockImplementation(f => {
+    setTimeout(() => {
```
axel7083, line 301: same remark as in https://github.com/containers/podman-desktop-extension-ai-lab/pull/2831/files#r2041688355

PR #2850, opened 2025-04-11
refactor: add failFastSubtasks to TaskRunner and refactor inferenceManager

feloy:
### What does this PR do?

Add failFastSubtasks to TaskRunner and refactor inferenceManager

### Screenshot / video of UI

N/A

Part of #2628 

### How to test this PR?

Start inference services

### DIFF
--- a/packages/backend/src/managers/TaskRunner.ts
+++ b/packages/backend/src/managers/TaskRunner.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { Task } from '@shared/models/ITask';
 import type { RunAsTaskOptions, TaskRunnerTools } from '../models/TaskRunner';
 import type { TaskRegistry } from '../registries/TaskRegistry';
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/managers/TaskRunner.ts
```diff
@@ -48,10 +49,26 @@ export class TaskRunner {
...
+        this.setErrorForSubtasksOnError(task, labels);
+      }
       throw err;
     } finally {
       task.progress = undefined;
       this.taskRegistry.updateTask(task);
     }
   }
+
+  private setErrorForSubtasksOnError(task: Task, labels: Record<string, string>): void {
```
axel7083, line 62: this is not really a helpful name to understand the purpose

axel7083: 1. the parameters: I can see task, and labels, no errors?
2. `set error for sub-task on error` I understand the purpose of the function by reading the code inside it, not the same

axel7083: Some ideas, let's take inspiration from github, `fail-fast` [^1], nice same, easy.

[^1]: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

axel7083: I don't think you need the first argument, as the task is already in error state, so it will not be in loading state

feloy: thanks, I've changed with `failFastSubtasks`

axel7083 (approved):

Issue #2665, opened 2025-03-06
[Bug] InstructLab 'Start FineTuning' button redirects to empty page


ScrewTSW:
### Bug description

same as title

### Operating system

Windows 10 22H2

### Installation Method

from `ghcr.io/containers/podman-desktop-extension-ai-lab` container image

### Version

next (development version)

### Steps to reproduce

* Install AI Lab 1.5.0
* Click on About InstructLab
* Click on Start FineTuning

### Relevant log output

```shell

```

### Additional context

![Image](https://github.com/user-attachments/assets/9227dc91-9c57-4a24-8e4a-0b89e511ae1e)
![Image](https://github.com/user-attachments/assets/354a6eb0-937c-4770-94b5-cd21a8a22f42)

PR #2853, opened 2025-04-13
fix: reworked fine-tuning button in about section

bmahabirbu:
### What does this PR do?

This PR fixes the issue where clicking on the fine-tune button will lead to a blank page
Now I reworked the button so that it will enable the setting in experimental feature for fine tuning

fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2665

### How to test this PR?

(have both folders in the same directory)
pnpm watch (in podman ai lab)
pnpm watch --extension-folder ../podman-desktop-extension-ai-lab/packages/backend (in podman desktop)

### DIFF
--- a/packages/frontend/src/pages/instructlab/AboutInstructLab.spec.ts
+++ b/packages/frontend/src/pages/instructlab/AboutInstructLab.spec.ts
@@ -1,0 +1,127 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


packages/backend/package.json
```diff
@@ -55,7 +55,7 @@
           "type": "boolean",
           "default": false,
           "description": "Display InstructLab Tuning screens (experimental)",
-          "hidden": true
+          "hidden": false
```
jeffmaury, line 58: It should be kept hidden as it is not fully implemented

jeffmaury (changes requested):

packages/backend/package.json
bmahabirbu: I understand. So in that case can we remove the fine-tuning button on the instructlab about page since the issue above describes it leading to a blank screen since experimentalTuning is hidden

benoitf (commented):
some comments from others already left

bmahabirbu:
Updated!

axel7083 (approved):

jeffmaury (approved):
LGTM. Would be good to have a unit test

PR #2848, opened 2025-04-11
feat: add OpenVINO inference provider

jeffmaury:
Related to #2847

### What does this PR do?

Add a new OpenVINO inference provider

### Screenshot / video of UI

TBA

### What issues does this PR fix or reference?

#2847 

### How to test this PR?

1. Add the following model in your catalog:

```
    {
      "id": "openvino-mistral-7B-instruct-v0.2-int8-ov",
      "name": "OpenVINO/mistral-7B-instruct-v0.2-int8-ov",
      "description": "OpenVINO/mistral-7B-instruct-v0.2-int8-ov",
      "url": "huggingface:/OpenVINO/mistral-7B-instruct-v0.2-int8-ov",
      "backend": "openvino"
    }
```
2. Start an inference server for that model
3. Start a playground

Issue #2613, opened 2025-03-04
Implementation Selecting inferencing runtime when starting a playground


slemeur:
### Is your enhancement related to a problem? Please describe

As a user, I should be able to configure the inferencing runtime when starting a playground service. It should be one more option, using a select box as we do for the model. 

<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/2cb5dd7b-15c2-488d-80c6-da182fbd5a4e" />

Also, the playground environment is using the existing model service, so we'll need to be careful about that. One way to handle it, would be to show the existing model services that are already running. I believe a "Model Service" section would be needed in the form. That section would be displaying the combination between the model and the inferencing runtime - this way, the user could select an already running one, or alternatively create a new one.

**Details:**
- As a user, I should be able to identify the default runtime
- As a user, I should be able to start multiple model services with the same configuration (model + inferencing runtime) using different ports
- As a user, I should be able to know if the inferencing runtime image is already pulled locally
- As a user, I should not be proposed the inferencing runtime if it is not compatible with my environment

### Describe the solution you'd like

Similar pattern than in the UX ticket: https://github.com/containers/podman-desktop-extension-ai-lab/issues/2502

Issue #2612, opened 2025-03-04
Implementation of Configuring Inference Runtime when starting a recipe


slemeur:
### Is your enhancement related to a problem? Please describe

As a user, I should be able to configure the inferencing runtime when starting a recipe. It should be one more option, using a select box as we do for the model.

<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/82711b82-4c51-4797-8ccd-5b15bcdcbc84" />

**Details:**
- As a user, I should be able to know if the inferencing runtime image is already pulled locally
- As a user, I should not be proposed the inferencing runtime if it is not compatible with my environment
- As a user, I should be able to identify the default runtime

### Describe the solution you'd like

Implementation of the UX ticket: https://github.com/containers/podman-desktop-extension-ai-lab/issues/2502

axel7083:
Before implementing this we need more inference providers
- https://github.com/containers/podman-desktop-extension-ai-lab/issues/2847
- https://github.com/containers/podman-desktop-extension-ai-lab/issues/2886

axel7083:
Correcting some elements:

It does not makes sense to select inference runtime for the recipe, the RECIPE is deciding what it needs

axel7083:
> can't a recipe be compatible with multiple inference runtime?

Not in the way they are design, currently we have the strict table

- llamacpp => run GGUF
- whispercpp => run specific whisper models
- Openvino => run .ov models
- VLLM => run .safetensor models (might run GGUF, but not primary goal)

Recipes are declaring which backend they need

https://github.com/containers/podman-desktop-extension-ai-lab/blob/a3d4939358971a7489b5f771c336bbf3821a4e48/packages/backend/src/assets/ai.json#L59

so we filter the models listed, based on that, values allowed are `llama-cpp`, `whisper-cpp`, `openvino`, and `none` as of today.

PR #2893, opened 2025-04-16
feat: added option to select default inferencing runtime

### LINKED ISSUE #2611: Implementation of default inferencing runtime

There will be one default configuration out of the box. Right now, we'll be starting with `llamacpp`, but the goal will be to move to `vLLM` as soon as possible.

As a user, I should be able to configure the default inferencing runtime to be used from the settings of AI Lab.

Image

This issue will be the implementation of the UX described in https://github.com/containers/podman-desktop-extension-ai-lab/issues/2501

_No response_

### Additional context

_No response_

gastoner:
### What does this PR do?
Adds settings to chose default inference runtime to preferences 
Implements mockup in the issue

### Screenshot / video of UI

[Screencast_20250428_150354.webm](https://github.com/user-attachments/assets/9ed74cd7-276a-40e1-878f-b70460a76f43)

Closes #2611 

### How to test this PR?
Select different inference runtime -> models should be filtered out


packages/backend/package.json
```diff
@@ -51,6 +51,16 @@
           "maximum": 65535,
           "description": "Port on which the API is listening (requires restart of extension)"
         },
+        "ai-lab.inferenceRuntime": {
+          "type": "string",
+          "default": "llama-cpp",
```
jeffmaury, line 56: Default should be empty

packages/backend/package.json
```diff
@@ -51,6 +51,16 @@
...
+          "enum": [
+            "llama-cpp",
```
jeffmaury, line 58: Is there a way we can have this dynamically generated ?

packages/frontend/src/lib/select/ModelSelect.svelte
```diff
@@ -44,6 +46,16 @@ let selected: (ModelInfo & { label: string; value: string }) | undefined = $deri
 function handleOnChange(nValue: (ModelInfo & { label: string; value: string }) | undefined): void {
   value = nValue;
 }
+
+let defaultRuntime: string = 'llama-cpp';
```
jeffmaury, line 50: Should be undefined

packages/frontend/src/lib/select/ModelSelect.svelte
```diff
@@ -54,6 +66,7 @@ function handleOnChange(nValue: (ModelInfo & { label: string; value: string }) |
   onchange={handleOnChange}
   placeholder="Select model to use"
   items={models
+    .filter(model => model.backend === defaultRuntime)
```
jeffmaury, line 69: Then if default runtime is undefined we should not filter

jeffmaury: Also would be good to have it as an helper function/method

packages/frontend/src/pages/NewInstructLabSession.spec.ts
```diff
@@ -45,6 +45,14 @@ vi.mock('../utils/client', async () => ({
   studioClient: {
     openURL: vi.fn(),
     openDialog: vi.fn(),
+    getExtensionConfiguration: vi.fn(),
```
jeffmaury, line 48: Why do we need this change ?

packages/frontend/src/pages/Recipes.svelte
```diff
@@ -100,6 +102,14 @@ const filtersComponents: { label: string; key: CatalogFilterKey }[] = [
 function openContribution(): void {
   studioClient.openURL('https://github.com/containers/ai-lab-recipes/blob/main/CONTRIBUTING.md').catch(console.error);
 }
+
+let defaultRuntime: string = $state('llama-cpp');
```
jeffmaury, line 106: Should be undefined

axel7083 (changes requested):
We cannot have a default for inference server not dealing with the same kind of files, like you cannot default to llamacpp instead of whispercpp as model handled by whispercpp cannot be handled by llamacpp.

This need to be specific to GGUF models, or safetensors. Choice should be `llamacpp`, `openvino`, `vllm`.

Meaning this PR should be delayed after the following PRs

- https://github.com/containers/podman-desktop-extension-ai-lab/pull/2900
- https://github.com/containers/podman-desktop-extension-ai-lab/pull/2848

In the meantime we cannot have a default, as the inference provided depends on the models, not the other way around

packages/frontend/src/lib/select/ModelSelect.svelte
gastoner: updated

packages/frontend/src/lib/select/ModelSelect.svelte
gastoner: updated

packages/backend/package.json
gastoner: fixed

packages/frontend/src/pages/Recipes.svelte
gastoner: fixed

packages/backend/package.json
axel7083: no there is not

packages/frontend/src/pages/NewInstructLabSession.spec.ts
gastoner: Because the NewInstructLabSession.svelte is rendering ModelSelect which uses the configuration

packages/backend/src/registries/ConfigurationRegistry.ts
```diff
@@ -26,14 +26,15 @@ const CONFIGURATION_SECTIONS: string[] = [
...
+  'inferenceRuntime',
   'experimentalTuning',
   'modelUploadDisabled',
   'showGPUPromotion',
   'appearance',
 ];
 
 const API_PORT_DEFAULT = 10434;
-
+const INFERENCE_RUNTIME_DEFAULT = 'llama-cpp';
```
jeffmaury, line 37: We should not provide this default

jeffmaury (changes requested):

packages/backend/src/registries/ConfigurationRegistry.ts
gastoner: Forgot to remove this :D

packages/frontend/src/lib/select/ModelSelect.svelte
```diff
@@ -44,6 +46,23 @@ let selected: (ModelInfo & { label: string; value: string }) | undefined = $deri
...
+
+onMount(() => {
+  return configuration.subscribe(values => {
+    if (values?.inferenceRuntime) {
+      defaultRuntime = values.inferenceRuntime;
+    }
+  });
+});
+
+function filterModel(model: ModelInfo): boolean {
```
jeffmaury, line 60: The issue that I see here is that if user select a preferred runtime then only models for that runtime will be available for selection where it should be the default but we should be able to see also other models.
So I would rather undo this change and add a control to select later (like in recipes page)

gastoner: I've added a checkbox that will filter out the models based on default runtime

jeffmaury (changes requested):
I did not configure anything but when I go to the settings page the settings tell me that llama-cpp is the prefered (I would expect to be none) but if I go to the recipes page I see only one recipe (object detection)
So I think we are missing the no value case

packages/backend/package.json
```diff
@@ -51,6 +51,15 @@
           "maximum": 65535,
           "description": "Port on which the API is listening (requires restart of extension)"
         },
+        "ai-lab.inferenceRuntime": {
+          "type": "string",
+          "enum": [
```
jeffmaury, line 56: So we need an All value as we cannot have no value for a dropdown. Discussed with @benoitf

gastoner: Ok, I'll add this

jeffmaury (commented):
LGTM.

However, I though you said you add a select in ModelSelect so that we can select models even if they are not for the preferred runtime ?

gastoner:
> LGTM.
> 
> However, I though you said you add a select in ModelSelect so that we can select models even if they are not for the preferred runtime ?

In the ModelSelect there is a mechanism of filtering the models based on selected Recipe ~ runime for this recipe

gastoner:
> > > LGTM.
> > > However, I though you said you add a select in ModelSelect so that we can select models even if they are not for the preferred runtime ?
> > 
> > 
> > In the ModelSelect there is a mechanism of filtering the models based on selected Recipe ~ runime for this recipe
> 
> I think ModelSelect is used in other contexts ?

Yes it is used in 3 or 4 places I think. Should I remove the filter from ModelSelect then?

gastoner:
Should I now also add the openvino filter? since the openvino PR is now closed?

gastoner:
@axel7083 can you PTAL?

slemeur:
I think the behavior is fine. And we have a couple of PRs that are waiting to be merged after this one, which will make the experience more logical.

What we are proposing the user is to use different Inferencing Runtimes with AI Lab. With this ticket, it will default the entire scope of AI Lab to a specific runtime. 
For example, a user who is defaulting to Whisper, will get the list of recipes filtered to those who are compatible with Whisper, but the user will still be able to remove the filter and see the entire list. 

For other situation (and probably the most common case), the user are most likely going to change the inferencing runtime at the time they are going to start one. For example, when the user is starting a recipe, when the user is starting to serve a model or starting a playground. We have the corresponding tickets for that: https://github.com/containers/podman-desktop-extension-ai-lab/issues/2612, https://github.com/containers/podman-desktop-extension-ai-lab/issues/2613

So the current behavior is good from my point of view.

Issue #2884, opened 2025-04-16
Add MCP interaction in the playground


jeffmaury:
### Is your enhancement related to a problem? Please describe

Playground should present MCP server as tools and handles tools calling based on the model response

### Describe the solution you'd like

- Add a new setting for a MCP server configuration file
- Handles MCP interaction when calling a model

manusa:
## Tasks _(WIP)_

- Create MCP Server Manager
  - Manager to store settings in `$appUserDirectory/mcp-settings.json`
  - Server settings stored in `{"server":{}}`

Format for the settings `mcp-settings.json` `"server"` object:
| Field | Description | Example |
| ----- | ----------- | ------- |
| `enabled` | Enable/disable the server | `true`, `false` |
| `type` | Server connection (`sse`/`stdio`). | `"sse"`, `"stdio"` |
| `command` | (STDIO) Command to start the server executable. | `"npx"`, `"uv"`, `"docker"` |
| `args` | (STDIO) Array of arguments passed to the command. | `["-y", "kubernetes-mcp-server"]` |
| `url` | (SSE) URL of the server | `"http://example.com/sse"` |
| `headers`| (SSE) HTTP headers for the server | `{"API_KEY": "THE_API_KEY"}` |

PR #2916, opened 2025-04-23
feat(playground): create McpServerManager

manusa:
### What does this PR do?

Part of #2884

Implements the base manager to configure MCP servers.

Manager to be consumed by playground in order to load tools.
Signed-off-by: Marc Nuri <marc@marcnuri.com>

### DIFF
--- a/packages/backend/src/managers/playground/McpServerManager.spec.ts
+++ b/packages/backend/src/managers/playground/McpServerManager.spec.ts
@@ -1,0 +1,127 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import type { McpServer } from './McpServerManager';
+import { McpServerType, McpServerManager } from './McpServerManager';
+
+/* eslint-disable sonarjs/no-nested-functions */
+describe('McpServerManager', () => {
+  let appUserDirectory: string;
+  let consoleErrors: string;
+  let consoleWarnings: string;
+  beforeEach(async () => {
+    appUserDirectory = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'mcp-server-manager-test-'));
+    consoleErrors = '';
+    vi.spyOn(console, 'error').mockImplementation((...args) => {
+      consoleErrors += args.join(' ') + '\n';
+    });
+    consoleWarnings = '';
+    vi.spyOn(console, 'warn').mockImplementation((...args) => {
+      consoleWarnings += args.join(' ') + '\n';
+    });
+  });
+  afterEach(async () => {
+    await fs.promises.rm(appUserDirectory, { recursive: true });
+  });
+  describe('load', () => {
+    test('with no file, returns empty array', async () => {
+      const mcpServerManager = new McpServerManager(appUserDirectory);
+      const mcpServers = await mcpServerManager.load();
+      expect(mcpServers).toEqual([]);
+      expect(consoleErrors).toContain('McpServerManager: Failed to load MCP settings');
+      expect(consoleWarnings).toBe('');
+    });
+    test('with empty file, returns empty array', async () => {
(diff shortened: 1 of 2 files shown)


packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -0,0 +1,75 @@
...
+
+// TODO: Consider using JsonWatcher in case the file is going to be updated externally
+export class McpServerManager implements Disposable {
+  private readonly settingsFile: string;
+
+  constructor(private appUserDirectory: string) {
+    this.settingsFile = path.join(this.appUserDirectory, MCP_SETTINGS);
+  }
+
+  async load(): Promise<McpServer[]> {
```
axel7083, line 47: Shouldn't we keep them in memory? The current logic does not save them

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -0,0 +1,75 @@
...
+  name: string;
+  enabled: boolean;
+  type: McpServerType;
+  command: string;
+  args: string[];
+  url: string;
+  headers: Record<string, string>;
+}
+
+// TODO: Consider using JsonWatcher in case the file is going to be updated externally
```
axel7083, line 39: Should we consider this as a potential use case? Like the user would be doing the update until he found something that he want?

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+              },
+            },
+            'invalid-type': {
+              enabled: true,
+              type: 'invalid',
+              url: 'https://echo.example.com/sse',
+            },
+          },
+        };
+        fs.writeFileSync(mcpSettingsFile, JSON.stringify(mcpSettings));
```
axel7083, line 97: Instead of writing the file, we should mock the `node:fs/promise` package, and mock the readFile method used by the MCP server manager

axel7083, line 97: Instead of writing the file, we should mock the `node:fs/promise` package, and mock the readFile method used by the MCP server manager

axel7083 (commented):
Hey! Thanks for this contribution, left some comments!

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: Any reason why you prefer mocking to actually writing and reading a temporary file?

Mocking fs is implementation dependent (white box). ---> If we later switch to a different implementation, alter the current one, and so on, the test won't be valid.

In general, I try to avoid mocks due to the exposed reason (and others).
Nonetheless, if this is the way you usually proceed, I'm happy to change the test.

packages/backend/src/managers/playground/McpServerManager.ts
manusa: Still not sure about this one: https://github.com/containers/podman-desktop-extension-ai-lab/pull/2916#discussion_r2055923268

manusa: Just to be more clear. This is the initial implementation of the manager, so that I can actually bind it to the playground manager and enable tool calling.
Depending on how the tools are managed we should keep this in memory or expand the management capabilities.

manusa:
> I think the location of the config file should be a configuration setting so that we can reuse existing configuration files (Claude Desktop,...)

This is something I was considering as a feature (it's also how [VSCode is doing it](https://code.visualstudio.com/docs/copilot/chat/mcp-servers#_automatic-discovery-of-mcp-servers)).

Provide 2 options to add the servers:
1. Regular settings file managed by AI Lab: User is able to manage the servers through the Podman AI Lab UI
2. Automatic discovery of settings: User is able to opt-in into this feature and AI Lab will try to find servers from well-known locations (Claude desktop, Goose CLI, etc.)

I would certainly make 2 opt-in, since many power users will probably have multiple clients and might not want all their settings to converge.

In general, as an MCP tool user, you don't want to have them enabled all at once, especially in scenarios where you are using smaller models (like in Podman AI Lab)

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -0,0 +1,75 @@
...
+        mcpServer.name = entry[0];
+        if (!Object.values(McpServerType).includes(mcpServer.type)) {
+          console.warn(
+            `McpServerManager: Invalid MCP server type ${mcpServer.type} for server ${mcpServer.name}. Defaulting to STDIO.`,
+          );
+          continue;
+        }
+        mcpServers.push(mcpServer);
+      }
+    } catch (error) {
```
jeffmaury, line 66:
```suggestion
    } catch (error: unknown) {
```

jeffmaury (commented):
Not sure to get why the manager does not keep the servers in memory

manusa:
> Not sure to get why the manager does not keep the servers in memory

I'm working on the entire implementation as opposed to baby steps which is what I was trying with this PR.

I guess that once you see everything working it will make more sense.

As to your question, for each server definition, we'll eventually create an MCP client instance that needs to be started and stopped (maybe for each prompt, or maybe for the lifecycle of the extension, or the playground).
Depending on how we want to deal with that, it might make sense to add file watchers or other mechanisms to make sure everything is in sync.

I tend to prefer the simplistic approaches first. In this case, there's a file with the server definitions and maybe other MCP settings.
The file becomes the single source of truth (it's a really small file and is cheap to read and write on demand).

manusa:
Superseded by #2949

PR #2949, opened 2025-05-02
MCP integration in the playground

### LINKED ISSUE #2884: Add MCP interaction in the playground

manusa:
### What does this PR do?

Adds Model Context Protocol (MCP) support to the playground.

#### Notes

- The presence of the `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/mcp-settings.json` file acts as a feature flag.
  - Adding/removing the file or renaming it will enable or disable the feature.
- Streaming is not supported when using tool/function calling with the OpenAI compatible inference servers.
  - I've added streaming emulation, but the UX is not great in the current state.
  - I'd suggest to add some sort of animation in the chat that tells the user that the model is working.
    (I can implement this as a subsequent PR so we can comment on it).
- This feature will only work with the **Granite 3.3 model**.
  I can improve other models to support it too, but will require additional work.
- In the demos I'm using a very simple MCP server that gets the current time at different timezones: https://github.com/modelcontextprotocol/servers/tree/main/src/time
- Adding MCP servers that provide multiple tools/functions will not work since the model context is very limited and can't be configured at the moment.
- Even if the model context is improved, will still won't work as smooth as other inference servers (such as Ollama).
  I assume that the limited amount of memory while running the server in a Podman container is what affects the performance. The end result is that chat completions that take seconds in Ollama, end up taking minutes when run within the container.

#### With the feature enabled

https://github.com/user-attachments/assets/78cb8963-80b1-4d6f-845b-7d1fec495b5f

#### With the feature disabled (nothing should be broken)

https://github.com/user-attachments/assets/d3f9a7fd-4bfd-4371-afd3-0bc1b72a0033

- Fixes #2884
- Supersedes #2916

### How to test this PR?

- Create a file `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/mcp-settings.json`.
- Add the following content to the file:
  ```json
  {
    "servers": {
      "time": {
        "enabled": true,
        "type": "stdio",
        "command": "podman",
      }
    }
  }
  ```
- Create a new playground and select the `ibm-granite/granite-3.3-8b-instruct-GGUF` model.
- Write a prompt that makes use of one of the MCP server's functions:
  ```
  Get the time in the Paris timezone
  ```
- Continue to chat to ensure follow-ups work

### DIFF
--- a/packages/backend/src/managers/playground/McpServerManager.spec.ts
+++ b/packages/backend/src/managers/playground/McpServerManager.spec.ts
@@ -15,55 +15,39 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
-import fs from 'node:fs';
-import os from 'node:os';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { promises } from 'node:fs';
 import path from 'node:path';
-import type { McpServer } from './McpServerManager';
-import { McpServerType, McpServerManager } from './McpServerManager';
+import { type McpServer, McpServerType } from '../../models/mcpTypes';
+import { McpServerManager } from './McpServerManager';
+
+vi.mock('node:fs');
+vi.mock('node:fs/promises');
 
 /* eslint-disable sonarjs/no-nested-functions */
 describe('McpServerManager', () => {
   let appUserDirectory: string;
-  let consoleErrors: string;
-  let consoleWarnings: string;
   beforeEach(async () => {
-    appUserDirectory = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'mcp-server-manager-test-'));
-    consoleErrors = '';
-    vi.spyOn(console, 'error').mockImplementation((...args) => {
-      consoleErrors += args.join(' ') + '\n';
-    });
-    consoleWarnings = '';
-    vi.spyOn(console, 'warn').mockImplementation((...args) => {
-      consoleWarnings += args.join(' ') + '\n';
-    });
-  });
-  afterEach(async () => {
-    await fs.promises.rm(appUserDirectory, { recursive: true });
+    vi.resetAllMocks();
+    appUserDirectory = path.join('/', 'tmp', 'mcp-server-manager-test-');
   });
   describe('load', () => {
     test('with no file, returns empty array', async () => {
       const mcpServerManager = new McpServerManager(appUserDirectory);
       const mcpServers = await mcpServerManager.load();
       expect(mcpServers).toEqual([]);
-      expect(consoleErrors).toContain('McpServerManager: Failed to load MCP settings');
-      expect(consoleWarnings).toBe('');
     });
     test('with empty file, returns empty array', async () => {
       const mcpServerManager = new McpServerManager(appUserDirectory);
-      vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => '{}');
+      vi.mocked(promises.readFile).mockResolvedValue('{}');
       const mcpServers = await mcpServerManager.load();
       expect(mcpServers).toEqual([]);
-      expect(consoleErrors).toBe('');
-      expect(consoleWarnings).toBe('');
     });
     test('with invalid JSON, returns empty array', async () => {
       const mcpServerManager = new McpServerManager(appUserDirectory);
-      vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => '{invalid json}');
+      vi.mocked(promises.readFile).mockResolvedValue('{invalid json}');
       const mcpServers = await mcpServerManager.load();
       expect(mcpServers).toEqual([]);
-      expect(consoleErrors).toContain('McpServerManager: Failed to load MCP settings');
-      expect(consoleWarnings).toBe('');
     });
     describe('with valid JSON', () => {
       let mcpServers: McpServer[];
@@ -91,18 +75,21 @@ describe('McpServerManager', () => {
             },
           },
         };
-        vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => JSON.stringify(mcpSettings));
+        vi.mocked(promises.readFile).mockResolvedValue(JSON.stringify(mcpSettings));
         mcpServers = await new McpServerManager(appUserDirectory).load();
       });
       test('parses stdio server', async () => {
-        const stdioOk: McpServer | undefined = mcpServers.find(server => server.name === 'stdio-ok');
-        expect(stdioOk).toEqual({
-          name: 'stdio-ok',
-          enabled: true,
-          type: McpServerType.STDIO,
-          command: 'npx',
-          args: ['-y', 'kubernetes-mcp-server'],
-        });
+        expect(mcpServers).toEqual(
+          expect.arrayContaining([
+            {
+              name: 'stdio-ok',
+              enabled: true,
+              type: McpServerType.STDIO,
+              command: 'npx',
+              args: ['-y', 'kubernetes-mcp-server'],
+            },
+          ]),
+        );
       });
       test('parses sse server', async () => {
         const sseOk: McpServer | undefined = mcpServers.find(server => server.name === 'sse-ok');
@@ -117,7 +104,6 @@ describe('McpServerManager', () => {
       test('ignores invalid type', async () => {
         const invalidType: McpServer | undefined = mcpServers.find(server => server.name === 'invalid-type');
         expect(invalidType).toBeUndefined();
-        expect(consoleWarnings).toContain('McpServerManager: Invalid MCP server type invalid for server invalid-type.');
       });
     });
   });
--- a/packages/backend/src/managers/playground/McpServerManager.spec.ts
+++ b/packages/backend/src/managers/playground/McpServerManager.spec.ts
@@ -51,8 +51,7 @@ describe('McpServerManager', () => {
     });
     test('with empty file, returns empty array', async () => {
       const mcpServerManager = new McpServerManager(appUserDirectory);
-      const mcpSettingsFile = path.join(appUserDirectory, 'mcp-settings.json');
-      await fs.promises.writeFile(mcpSettingsFile, '{}');
+      vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => '{}');
       const mcpServers = await mcpServerManager.load();
       expect(mcpServers).toEqual([]);
       expect(consoleErrors).toBe('');
@@ -60,8 +59,7 @@ describe('McpServerManager', () => {
     });
     test('with invalid JSON, returns empty array', async () => {
       const mcpServerManager = new McpServerManager(appUserDirectory);
-      const mcpSettingsFile = path.join(appUserDirectory, 'mcp-settings.json');
-      await fs.promises.writeFile(mcpSettingsFile, '{invalid json}');
+      vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => '{invalid json}');
       const mcpServers = await mcpServerManager.load();
       expect(mcpServers).toEqual([]);
       expect(consoleErrors).toContain('McpServerManager: Failed to load MCP settings');
@@ -70,7 +68,6 @@ describe('McpServerManager', () => {
     describe('with valid JSON', () => {
       let mcpServers: McpServer[];
       beforeEach(async () => {
-        const mcpSettingsFile = path.join(appUserDirectory, 'mcp-settings.json');
         const mcpSettings = {
           servers: {
             'stdio-ok': {
@@ -94,7 +91,7 @@ describe('McpServerManager', () => {
             },
           },
         };
-        fs.writeFileSync(mcpSettingsFile, JSON.stringify(mcpSettings));
+        vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => JSON.stringify(mcpSettings));
         mcpServers = await new McpServerManager(appUserDirectory).load();
       });
       test('parses stdio server', async () => {
--- a/packages/backend/src/managers/playground/McpServerManager.spec.ts
+++ b/packages/backend/src/managers/playground/McpServerManager.spec.ts
@@ -1,0 +1,127 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import type { McpServer } from './McpServerManager';
+import { McpServerType, McpServerManager } from './McpServerManager';
+
+/* eslint-disable sonarjs/no-nested-functions */
+describe('McpServerManager', () => {
+  let appUserDirectory: string;
+  let consoleErrors: string;
+  let consoleWarnings: string;
+  beforeEach(async () => {
+    appUserDirectory = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'mcp-server-manager-test-'));
+    consoleErrors = '';
+    vi.spyOn(console, 'error').mockImplementation((...args) => {
+      consoleErrors += args.join(' ') + '\n';
+    });
+    consoleWarnings = '';
+    vi.spyOn(console, 'warn').mockImplementation((...args) => {
+      consoleWarnings += args.join(' ') + '\n';
+    });
+  });
+  afterEach(async () => {
+    await fs.promises.rm(appUserDirectory, { recursive: true });
+  });
+  describe('load', () => {
+    test('with no file, returns empty array', async () => {
+      const mcpServerManager = new McpServerManager(appUserDirectory);
+      const mcpServers = await mcpServerManager.load();
+      expect(mcpServers).toEqual([]);
+      expect(consoleErrors).toContain('McpServerManager: Failed to load MCP settings');
+      expect(consoleWarnings).toBe('');
+    });
+    test('with empty file, returns empty array', async () => {
+      const mcpServerManager = new McpServerManager(appUserDirectory);
+      const mcpSettingsFile = path.join(appUserDirectory, 'mcp-settings.json');
+      await fs.promises.writeFile(mcpSettingsFile, '{}');
+      const mcpServers = await mcpServerManager.load();
+      expect(mcpServers).toEqual([]);
+      expect(consoleErrors).toBe('');
+      expect(consoleWarnings).toBe('');
+    });
+    test('with invalid JSON, returns empty array', async () => {
+      const mcpServerManager = new McpServerManager(appUserDirectory);
+      const mcpSettingsFile = path.join(appUserDirectory, 'mcp-settings.json');
+      await fs.promises.writeFile(mcpSettingsFile, '{invalid json}');
+      const mcpServers = await mcpServerManager.load();
+      expect(mcpServers).toEqual([]);
+      expect(consoleErrors).toContain('McpServerManager: Failed to load MCP settings');
+      expect(consoleWarnings).toBe('');
+    });
+    describe('with valid JSON', () => {
+      let mcpServers: McpServer[];
+      beforeEach(async () => {
+        const mcpSettingsFile = path.join(appUserDirectory, 'mcp-settings.json');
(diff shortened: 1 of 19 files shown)


packages/backend/src/managers/playground/aiSdk.ts
```diff
@@ -126,15 +181,23 @@ export class AiStreamProcessor<TOOLS extends ToolSet> {
...
     // Ensure the last message is marked as complete to allow the user to resume the conversation
     if (this.currentMessageId !== undefined) {
-      // TODO, this doesn't seem very wise (using choices as partial state holder)
-      // Refactor to use this.conversationRegistry.update instead
       this.conversationRegistry.completeMessage(this.conversationId, this.currentMessageId);
     }
   };
 
-  stream = (model: LanguageModel, options?: ModelOptions): StreamTextResult<TOOLS, never> => {
+  private onFinish = (stepResult: Omit<StepResult<TOOLS>, 'stepType' | 'isContinued'>): void => {
```
gastoner, line 188: Can't we create a special type for 'stepType' | 'isContinued'?

packages/backend/src/managers/playgroundV2Manager.spec.ts
```diff
@@ -72,16 +75,19 @@ beforeEach(async () => {
   vi.mocked(rpcExtensionMock.fire).mockResolvedValue(true);
   vi.useFakeTimers();
   const aiSdkSpecShared = await import('./playground/aiSdk.spec');
+  appUserDirectory = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'mcp-server-manager-test-'));
```
gastoner, line 78: Just `mcp-server-manager-test` without the last `-` ?

packages/backend/src/registries/ConversationRegistry.ts
```diff
@@ -107,55 +110,72 @@ export class ConversationRegistry extends Publisher<Conversation[]> implements D
...
       ...conversation.messages[messageIndex],
-      choices: [...((conversation.messages[messageIndex] as PendingChat)?.choices || []), choice],
-    } as PendingChat);
+      content: ((conversation.messages[messageIndex] as AssistantChat).content ?? '') + delta,
+    } as AssistantChat);
+  }
+
+  /**
+   *
+   */
```
gastoner, line 158: Missing comment?

packages/backend/src/managers/playgroundV2Manager.spec.ts
manusa: The standard approach usually involves adding a trailing dash:
https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

End result is a more readable `mcp-server-manager-test-123456` directory as opposed to `mcp-server-manager-test123456`.

But I assume it's just a matter of taste. I can change if you want and this is how you usually proceed.

gastoner: Oh thanks for clarifying, no need to to this then

gastoner (approved):
Thanks for the changes! 
BTW I belive that also `MaziyarPanahi/Mistral-7B-Instruct-v0.3.Q4_K_M` should work with this PR
image

packages/backend/src/registries/ConversationRegistry.ts
```diff
@@ -107,55 +110,72 @@ export class ConversationRegistry extends Publisher<Conversation[]> implements D
...
-    if (messageIndex === -1)
-      throw new Error(`message with id ${messageId} does not exist in conversation ${conversationId}.`);
-
-    this.update(conversationId, messageId, {
-      ...conversation.messages[messageIndex],
-      usage: usage,
-    } as PendingChat);
+    if (!usage) {
+      return;
+    }
```
axel7083, line 130: if usage cannot be undefined, the check from line 128 to 130 is not needed anymore

packages/backend/src/registries/ConversationRegistry.ts
```diff
@@ -107,55 +110,72 @@ export class ConversationRegistry extends Publisher<Conversation[]> implements D
...
   /**
-   * Utility method to quickly add a choice to a given a message inside a conversation
+   * Utility method to quickly add a delta to a given a message inside a conversation
    * @param conversationId
    * @param messageId
-   * @param choice
+   * @param delta
    */
-  appendChoice(conversationId: string, messageId: string, choice: Choice): void {
+  textDelta(conversationId: string, messageId: string, delta: string): void {
```
axel7083, line 144: I am not sure to understand what a delta is, could you clarify?

packages/backend/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "esnext",
     "module": "esnext",
-    "moduleResolution": "Node",
+    "moduleResolution": "bundler",
```
axel7083, line 5: this seems odd ? why is this required ?

packages/shared/src/models/IPlaygroundMessage.ts
```diff
@@ -34,13 +34,13 @@ export interface ModelUsage {
...
   role: 'system' | 'user' | 'assistant';
-  content?: string;
-  usage?: ModelUsage;
+  content?: string | object;
 }
 
 export interface AssistantChat extends ChatMessage {
   role: 'assistant';
   completed?: number;
+  content?: string | ToolCall;
```
axel7083, line 43: Why not using a dedicated properties, like `tools` ? trying to put everything into content may lead to type issues, and force us to always check

packages/shared/src/models/IPlaygroundMessage.ts
```diff
@@ -81,6 +90,10 @@ export function isAssistantChat(msg: Message): msg is AssistantChat {
   return isChatMessage(msg) && msg.role === 'assistant';
 }
 
+export function isAssistantToolCall(msg: Message): msg is AssistantChat {
+  return isAssistantChat(msg) && (msg.content as ToolCall)?.type === 'tool-call';
+}
```
axel7083, line 95: follow up to https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949/files#r2073488474, we should now have a single type/interface with several meaning.

AssistantChat should have an optional `tools` property,

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      const result = within(container).getByTestId('tool-call-result');
+      expect(result?.textContent).toBe(
+        '{\n' +
+          '  "content": [\n' +
+          '    {\n' +
+          '      "type": "text",\n' +
+          '      "text": "The weather in Don Benito is sunny with a temperature of 25°C."\n' +
+          '    }\n' +
+          '  ]\n' +
+          '}',
```
axel7083, line 70:
```ts
expect(JSON.parse(result?.textContent)).toEqual({
  content: [...]
})
```

suggestion: probably easier to test like that, we don't know the line return etc. could change in the future when updating the component, let's only test for the important content

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+ ***********************************************************************/
+import { expect, test, beforeEach, describe } from 'vitest';
+import { fireEvent, render, within } from '@testing-library/svelte';
+import ToolCallMessage from '/@/lib/conversation/ToolCallMessage.svelte';
+
+describe('ToolCallMessage component', () => {
+  let container: HTMLElement;
+  describe('with result', () => {
+    beforeEach(() => {
+      container = render(ToolCallMessage, {
```
axel7083, line 26: remark: Here you don't need it because you don't mock anything, but it is always preferable to have `vi.resetAllMock()` to avoid forgetting when we need it.

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    });
+    test('Has toggle button', () => {
+      const toggleButton = within(container).getByTestId('tool-call-show-details');
+      expect(toggleButton).toBeDefined();
+    });
+    test('Clicking toggle button shows details', async () => {
+      const toggleButton = within(container).getByTestId('tool-call-show-details');
+      await fireEvent.click(toggleButton);
+      const result = within(container).getByTestId('tool-call-details');
+      expect(Array.from(result?.classList)).not.contains('h-0');
```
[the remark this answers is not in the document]
axel7083: Need https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949/files#r2073501005

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+            toolCallId: '1-1',
+            toolName: 'weather',
+            args: { location: 'Don Benito' },
+          },
+        },
+      }).container;
+    });
+    test('Has role', () => {
+      const role = within(container).getByTestId('tool-call-role');
+      expect(role?.textContent).toBe('Assistant');
```
axel7083, line 108: If you fix the import you will be able to use `expect(role).toHaveTextContent`

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+import type { McpServer } from './McpServerManager';
+import { McpServerType, McpServerManager } from './McpServerManager';
+
+/* eslint-disable sonarjs/no-nested-functions */
+describe('McpServerManager', () => {
+  let appUserDirectory: string;
+  let consoleErrors: string;
+  let consoleWarnings: string;
+  beforeEach(async () => {
+    appUserDirectory = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'mcp-server-manager-test-'));
```
axel7083, line 31: you should always include the `vi.resetAllMock()` to avoid having tests leaking with each other

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    consoleErrors = '';
+    vi.spyOn(console, 'error').mockImplementation((...args) => {
```
axel7083, line 33: I would not really recommend to test console.* functions, logging information could change be updated, and IHMO should not be use to test the flow of the tested code

axel7083: If you need to check for an error, if should be raised, if this is only warning, we don't need to test it, this is not core feature

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+      consoleErrors += args.join(' ') + '\n';
+    });
+    consoleWarnings = '';
+    vi.spyOn(console, 'warn').mockImplementation((...args) => {
+      consoleWarnings += args.join(' ') + '\n';
+    });
+  });
+  afterEach(async () => {
+    await fs.promises.rm(appUserDirectory, { recursive: true });
+  });
```
axel7083, line 43: I usually don't like that we do real IO in our testing, because if one test fails the afterEach is not ran

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+      const mcpServerManager = new McpServerManager(appUserDirectory);
+      vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => '{}');
+      const mcpServers = await mcpServerManager.load();
+      expect(mcpServers).toEqual([]);
+      expect(consoleErrors).toBe('');
+      expect(consoleWarnings).toBe('');
+    });
+    test('with invalid JSON, returns empty array', async () => {
+      const mcpServerManager = new McpServerManager(appUserDirectory);
+      vi.spyOn(fs.promises, 'readFile').mockImplementation(async () => '{invalid json}');
```
axel7083, line 62: I would rather we use `vi.mock` at the top of the file instead of spying one module functions.

Some explanation, when partially mocking some stuff with vi.spyOn, we make a module dirty (some stuff are mocked, others not), this is a readability problem, when I read the code I don't know what is real and what is not.

Mocking the full package is usually easier, you test your code, not the IO

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+      });
+      test('parses stdio server', async () => {
+        const stdioOk: McpServer | undefined = mcpServers.find(server => server.name === 'stdio-ok');
+        expect(stdioOk).toEqual({
+          name: 'stdio-ok',
+          enabled: true,
+          type: McpServerType.STDIO,
+          command: 'npx',
+          args: ['-y', 'kubernetes-mcp-server'],
+        });
```
axel7083, line 105: You can use [expect.arrayContaining(array)](https://jestjs.io/docs/expect#expectarraycontainingarray)

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -0,0 +1,125 @@
...
+  name: string;
+  enabled: boolean;
+  type: McpServerType;
+  command: string;
+  args: string[];
+  url: string;
+  headers: Record<string, string>;
+}
+
+export interface McpClient {
```
axel7083, line 47: Please move all the interfaces to the `packages/backend/src/models/*` either have a single `mcp-types.ts` or a file per interface

axel7083 (commented):
Honestly, great work! I left a lot of comments, sorry about that, mostly around the testing and where the code is, but the core logic is very good, kudos 👍

packages/backend/src/registries/ConversationRegistry.ts
manusa: In streaming mode, the inference server will send `text-delta` chunks as it completes the chat prompt.

Previously, we were reusing `choice` which is not what it's intended for (an LLM can provide multiple text responses as choices for the user to select the most appropriate).

This function is used to concatenate the new delta to the **text** content of the Assistant's response.

packages/shared/src/models/IPlaygroundMessage.ts
manusa: Because this is how the inference server returns the assistant's response.
I didn't want to diverge much from the upstream behaviors.

axel7083: If this is how the server respond I am good with it, forget my comments about it

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import { expect, test, beforeEach, describe } from 'vitest';
+import { fireEvent, render, within } from '@testing-library/svelte';
+import ToolCallMessage from '/@/lib/conversation/ToolCallMessage.svelte';
+
+describe('ToolCallMessage component', () => {
```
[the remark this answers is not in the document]
manusa: I never implement a test without a proper `describe` scope.
Besides making tests more descriptive, depending on the test runner, non-scoped test groups can lead to unexpected test results.

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
manusa: I will add it.
However, as I stated in the previous PR, striving to not use mocks usually leads to a healthier code-base ;)

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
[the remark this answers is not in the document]
manusa: :+1: wasn't working for me. I was missing https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949/files#r2073501005, thx!

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: [Remember ](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949#discussion_r2073535508)

Also note that you can set this as a global test setup in vitest if you're really concerned about it.

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: Well, here I'm testing it as part of the behavior.
If it's not a core feature, then I'd suggest not to add it.
Not sure if you've noticed, but I'm kind of orthodox and extremely strict with tests.
If there's something in the production code I will test it. If that's not part of the impl, then I'll remove it from the production code. But, in general, I won't have both.

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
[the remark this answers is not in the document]
axel7083: > Besides making tests more descriptive

We use the filename for the scope

> depending on the test runner, non-scoped test groups can lead to unexpected test results.

This is nt true with vitest

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
axel7083: > However, as I stated in the previous PR, striving to not use mocks usually leads to a healthier code-base ;)

That's an interesting opinion

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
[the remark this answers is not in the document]
axel7083: marking as resolved

packages/backend/src/managers/playground/McpServerManager.spec.ts
axel7083: Yeah and I really appreciate the quality of code you provide, but testing console.log makes the test too complex for just checking that something will appear in the terminal

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: Not sure I'm following, `afterEach` should **always** be run regardless of the test failures.
They're main purpose is mostly to clean up resources such as temporary files.

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: I will remove this, but note that I'm extremely uncomfortable doing so.

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
manusa: In this case I want to make sure that the text is displayed this way since it's within a `<pre>` block i.e. formatting is part of the expected behavior.

In fact, using the testing-library `toHaveTextContent` is actually tampering with the output too and I'm not able to use it and 100% verify that text is displayed as I want.

For both fields (arguments + result), I think it's better if we keep the test this way.
I'm not sure either if there will be someone from the design team reviewing this

packages/frontend/src/lib/conversation/ToolCallMessage.spec.ts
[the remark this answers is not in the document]
manusa: It actually is if your test file ends up imported in a different test suite.

Vitest will also have different behaviors depending on how the esbuild (bundler) ends up transpiling your code.

I'll remove it, but not feeling very comfortable about it.

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: I understand the difference between partial mocking and complete mocking.
As I said earlier, I'm completely against this, just doing it to comply with the code-base coding styles.

I believe many of the issues you have, especially those with the Windows filesystem, are in part because you are testing your production code this way.

I'm currently changing the test to use `vi.mock`.

packages/backend/src/managers/playground/McpServerManager.spec.ts
manusa: Anyway, removed in the scope of other comments :shrug:

manusa:
> Honestly, great work! I left a lot of comments, sorry about that, mostly around the testing and where the code is, but the core logic is very good, kudos 👍

Thanks for reviewing!

I think I managed to address all of requested changes except for two open points:

- https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949#discussion_r2073671033 (arguments+result pre output verification)
- https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949#discussion_r2073680230 (tsconfig moduleResolution)

packages/backend/tsconfig.json
axel7083: I confirm that building the extension, and installing from a registry works well, so this does not seems to affect the build output.

axel7083 (approved):
I was able to tests on Windows with the CUDA images, works as expected 👍 

> Thanks again for the changes, and explanation provided in this PR  

image

manusa:
> I was able to tests on Windows with the CUDA images, works as expected 👍
> 
> > Thanks again for the changes, and explanation provided in this PR
> 
> image

Awesome, thx!

slemeur:
oh fantastic!!!

Issue #2960, opened 2025-05-06
bug(ui): `state_unsafe_mutation`


axel7083:
### Bug description

While testing https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949 I had a svelte error raised

**We should not try to mutate a state inside a `$derived`**, here at line L63, we are updating the `errorMsg` state **inside** at derived state: this is wrong and make the component crash

https://github.com/containers/podman-desktop-extension-ai-lab/blob/dafb8807898380133ae8086c6aab1e0549ab2c8d/packages/frontend/src/pages/Playground.svelte#L53-L70

### Operating system

Windows 11

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

1. Have the inference server throw an error (E.g. `got exception: {"code":500,"message":"tools param requires --jinja flag","type":"server_error"}`)
2. assert the playground page crash

https://github.com/user-attachments/assets/5d3892d1-bbed-4135-8113-048d10991cf9

### Relevant log output

```shell
index-C-xhvNv8.js:2 Uncaught (in promise) Error: https://svelte.dev/e/state_unsafe_mutation
```

PR #2961, opened 2025-05-06
fix(ui): playground `state_unsafe_mutation`

axel7083:
### What does this PR do?

The tests were not covering the scenario where the server return an error, and this error was provided to the playground.

### Screenshot / video of UI

image

Fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2960

### How to test this PR?

- [x] test cover the issue and ensure no regression

**Manually**

This problem can be easily reproduced with a NVIDIA GPU.

1. assert you have an NVIDIA card
2. checkout https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949
3. start Podman Desktop & AI Lab
4. assert GPU enabled
5. start playground with `ibm-research/granite-3.2-8b-instruct-GGUF`
6. talk in the chat
7. assert error from the screenshot above

### DIFF
--- a/packages/frontend/src/pages/Playground.svelte
+++ b/packages/frontend/src/pages/Playground.svelte
@@ -9,6 +9,7 @@ import {
   isChatMessage,
   isErrorMessage,
   isAssistantToolCall,
+  type Message,
 } from '@shared/models/IPlaygroundMessage';
 import { catalog } from '../stores/catalog';
 import ContentDetailsLayout from '../lib/ContentDetailsLayout.svelte';
@@ -24,6 +25,7 @@ import { router } from 'tinro';
 import ConversationActions from '../lib/conversation/ConversationActions.svelte';
 import { ContainerIcon } from '@podman-desktop/ui-svelte/icons';
 import ToolCallMessage from '/@/lib/conversation/ToolCallMessage.svelte';
+import type { InferenceServer } from '@shared/models/IInference';
 
 interface Props {
   playgroundId: string;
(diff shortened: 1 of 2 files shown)


packages/frontend/src/pages/Playground.svelte
```diff
@@ -67,8 +70,14 @@ let sendEnabled = $derived.by(() => {
   return false;
 });
 
-let server = $derived(
-  $inferenceServers.find(is => conversation && is.models.map(mi => mi.id).includes(conversation?.modelId)),
+$effect(() => {
+  if (latest && isErrorMessage(latest)) {
+    errorMsg = latest.error;
+  }
+});
```
benoitf, line 77: hello, wondering why it's not also a derived state ?

something like
```
errorMsg = $derived(latest && isErrorMessage(latest) ? latest.error : undefined)
```

axel7083: because when calling the `submit` method, we put the potential error also inside this state

https://github.com/containers/podman-desktop-extension-ai-lab/blob/dafb8807898380133ae8086c6aab1e0549ab2c8d/packages/frontend/src/pages/Playground.svelte#L89-L91

gastoner (approved):
LGTM, tested on Linux

AFAK now in playground works only 2 models (`MaziyarPanahi/Mistral-7B-Instruct-v0.3.Q4_K_M` and `granite3.3` - those which have --jinja flag enabled) I think we should maybe disable other models from playground if they don't work in my opinion

manusa (approved):

Issue #2971, opened 2025-05-07
Playground streaming mode is not working anymore


jeffmaury:
### Bug description

Playground now takes a while after a request is sent and seems streaming is not enabled anymore

Seems to MCP enablement related

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

manusa:
Streaming **IS NOT** supported when using tools (MCP) (Stated in the [PR description](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949#issue-3035649998))

[Comment](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2961#issuecomment-2857718842) from @manusa in #2961:

> This might be because you have the mcp-servers.json file which enables the MCP tooling feature.
> By removing or renaming the file, everything should work as expected.
>
> **Maybe I can also add additional conditions to prevent MCP to be enabled on incompatible models ---> update the default catalog with some label/property?**

The mcp-settings.json is used as a feature flag.
I assume most of you are facing this issue because you still have the file after testing #2949

jeffmaury:
> Streaming **IS NOT** supported when using tools (MCP) (Stated in the [PR description](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949#issue-3035649998))
> 
> [Comment](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2961#issuecomment-2857718842) from [@manusa](https://github.com/manusa) in [#2961](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2961):
> 
> > This might be because you have the mcp-servers.json file which enables the MCP tooling feature.
> > By removing or renaming the file, everything should work as expected.
> > **Maybe I can also add additional conditions to prevent MCP to be enabled on incompatible models ---> update the default catalog with some label/property?**
> 
> The `mcp-settings.json` is used as a feature flag. I assume most of you are facing this issue because you still have the file after testing [#2949](https://github.com/containers/podman-desktop-extension-ai-lab/pull/2949)

No I don't but I have the feeling the issue is gone will check again

manusa:
I proposed somewhere else (I can't remember the PR or issue) to add some UI aid to show the user that the prompt is in progress.
I'll create a PR with a PoC so you can evaluate

Issue #2885, opened 2025-04-16
Allow playground to select tools/


jeffmaury:
### Is your enhancement related to a problem? Please describe

Allow the playground UI to select/unselect MCP servers loaded from the configuration file

### Describe the solution you'd like

Extends the settings panel with an MCP section allowing to select/unselect an MCP server

manusa:
Can I pick this one up (@bmahabirbu / @jeffmaury)?
I can  do it in the scope of improving the functionality of the McpServerManager that was introduced in #2949

PR #2972, opened 2025-05-07
feat(playground): mcp-settings.json watched for external changes

manusa:
### What does this PR do?

Enables file system watching of the `mcp-settings.json` file.

This is a required step to be able to complete #2885 easily.
After this one is merged, a subsequent step in #2885 can be to list in the UI what are the servers currently active (this can be iterated later on to actually display the list of available tools).

### Screenshot / video of UI

https://github.com/user-attachments/assets/5f9b6017-71f0-4e82-b1bd-ded6ad100abe

- Part of #2885
- Relates to #2971

### How to test this PR?

- Create or update a file `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/mcp-settings.json`.
- Set the following content to the file:
  ```json
  {
    "servers": {
      "time": {
        "enabled": true,
        "type": "stdio",
        "command": "podman",
      }
    }
  }
  ```
- Create a new playground and select the ibm-granite/granite-3.3-8b-instruct-GGUF model.
- Write a prompt that makes use of one of the MCP server's functions:
  ```
  Get the time in the Paris timezone
  ```
- A tool call should be performed, the assistant response **SHOULD NOT** be streamed as streaming is not supported yet when using tools.
- Change the value of the server in the `mcp-settings.json` file to disabled (or delete the file, or rename it).
- Write another prompt that doesn't make use of tools.
- The assistant response should be streamed as streaming **IS** supported when no tools are used.


packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -16,94 +16,178 @@
...
     vi.resetAllMocks();
+    vi.mocked(EventEmitter).mockImplementation(() => new TestEventEmitter() as unknown as EventEmitter<unknown>);
+    vi.mocked(fs.createFileSystemWatcher).mockImplementation(
+      () =>
+        ({
+          onDidChange: (fn: unknown) => (onDidChangeBinding = fn as () => void),
+          onDidCreate: (fn: unknown) => (onDidCreateBinding = fn as () => void),
+          onDidDelete: (fn: unknown) => (onDidDeleteBinding = fn as () => void),
+        }) as unknown as FileSystemWatcher,
+    );
```
axel7083, line 52: Here we are not focusing on the McpServerManager, but we are also testing the `JsonWatcher` which has its dedicated file for that.

Mixing both testing logic (mocking file system watcher & event emitter logic) makes the test pretty complicated.

To simplify, and only focus on the McpServerManager, maybe we could create a JsonWatcherMock class, and remove the node:fs & file system watcher logic.

manusa: That's why the original tests were not relying on mocks but on real files.

I'd seriously recommend not to rely on mocks and stubs for this kind of features.
If you mock everything you'll end up having extremely brittle code.

The tests should be very simple and behavior-oriented:
- With no mcp-settings file
  - getMcpSettings returns empty servers
  - When mcp-settings file created with valid servers
    - getMcpSettings returns list of servers
  - When mcp-settings file created with invalid content
    - getMcpSettings returns empty servers
- With mcp-settings file
  - getMcpSettings returns list of servers
  - When mcp-settings file deleted
    - getMcpSettings returns empty servers
  - When mcp-settings file updated
    - getMcpSettings returns list of updated servers

Notice how you can easily test and assert this by performing simple `fs` operations on the file (which are far more readable than mocking).
In addition, you'll ensure that this approach verifies the feature works regardless of the platform/system you run them on.

The only requirement is to create a test double for the `@podman-desktop/api` `fs` module which could then be reused across many other tests.


I don't really care about the `McpServerManager` or wherever the logic resides in.
I care about the user expectations and the according behavior. That's why we probably have such a different perspective on the approach to testing.

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -15,52 +15,67 @@
...
-// 1. As a user, I want to be able to modify the mcp-settings.json file and have the changes reflected in the application without restarting it.
-// 2. As a user, I want to be able to add/remove servers from the mcp-settings.json while also having the application to provide a UI to manage the servers.
-// 3. As a user, I don't want to tinker with the mcp-settings.json file and would like to have the application to provide a UI to manage the servers.
-// Cases 1 and 2 are covered by JsonWatcher, but case 3 is not.
-// For 2 and 3, we need to implement a UI to manage the servers.
-export class McpServerManager implements Disposable {
+export class McpServerManager extends Publisher<McpSettings> implements Disposable {
   private readonly settingsFile: string;
+  private mcpSettings: McpSettings;
+  private readonly _onUpdate: EventEmitter<McpSettings>;
```
axel7083, line 33: The `_onUpdate` do not have any listeners, you are firing events, but I don't anything subscribing to it.

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -15,52 +15,67 @@
...
+  readonly #jsonWatcher: JsonWatcher<McpSettings>;
```
axel7083, line 34: The jsonWatcher should be disposed in the `dispose()`

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -15,52 +15,67 @@
...
+  ) {
+    super(rpcExtension, MSG_MCP_SERVERS_UPDATE, () => this.getMcpSettings());
     this.settingsFile = path.join(this.appUserDirectory, MCP_SETTINGS);
+    this.mcpSettings = {
+      servers: {},
+    };
+    this._onUpdate = new EventEmitter<McpSettings>();
+    this.#jsonWatcher = new JsonWatcher<McpSettings>(this.settingsFile, { ...this.mcpSettings });
+    this.#jsonWatcher.onContentUpdated(this.onMcpSettingsUpdated.bind(this));
+    this.#jsonWatcher.init();
```
axel7083, line 48: we usually uses a `init()` method on managers class

https://github.com/containers/podman-desktop-extension-ai-lab/blob/878854ad6ddcc8d8551404b11fd124f595b81e96/packages/backend/src/managers/catalogManager.ts#L63

might be nice to init the JsonWatcher inside this method.

> The idea is that in the future, we only init classes when we need its content

packages/backend/src/managers/playground/McpServerManager.ts
```diff
@@ -15,52 +15,67 @@
...
   }
 
   dispose(): void {
     // NO OP
   }
+
+  override notify(): void {
+    super.notify();
+    this._onUpdate.fire(this.getMcpSettings());
+  }
```
axel7083, line 80: If another class need to subscribe to the MCPManager, to subscribe to it, maybe we should port https://github.com/podman-desktop/extension-podman-quadlet/pull/539 to AI Lab.

packages/backend/src/managers/playground/McpServerManager.spec.ts
jeffmaury: The pattern we use in Podman Desktop and Podman AI Lab in unit testing is to keep unit tests focused on the component under test and to mock sub components used by the component under test

Mocking JsonWatcher would make the test more simple and readable

manusa: Yes, I agreed with @axel7083 to keep everything inline with the current repo and org approach.

Nonetheless, I'd strongly recommend to invest some resources in creating a better test framework that provides testing infra for `@podman-desktop/api` and allows tests to be written without mocking every component's dependency. Tests would then be easier to write, maintain, and would probably end up being more reliable. If you're interested on learning more about this then let me know.

packages/backend/src/managers/playground/McpServerManager.ts
manusa: :+1: This is part of the upcoming feature. Removing it in scope of this PR

manusa:
Hi @axel7083 @jeffmaury 
All comments should have been addressed.
I've also changed the affected test suite to completely adapt to your org styles.
Please check if everything is aligned as per your expectations.

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -16,95 +16,106 @@
...
+let appUserDirectory: string;
+let mcpServerManager: McpServerManager;
+beforeEach(async () => {
+  vi.resetAllMocks();
+  mockJsonWatcher = {
+    init: vi.fn(),
+    dispose: vi.fn(),
+    onContentUpdated: vi.fn((fn: (mcpSettings: McpSettings) => void) => (update = fn)),
+  } as unknown as JsonWatcher<McpSettings>;
+  vi.mocked(JsonWatcher).mockImplementation(() => mockJsonWatcher);
```
axel7083, line 41:
```suggestion
  vi.mocked(JsonWatcher).mockReturnValue(mockJsonWatcher);
```

Here, when mocking a constructor, you may use mock return value directly

axel7083 (approved):

packages/backend/src/managers/playground/McpServerManager.spec.ts
```diff
@@ -16,95 +16,106 @@
...
-              headers: {
-                foo: 'bar',
-              },
-            },
-            'invalid-type': {
-              enabled: true,
-              type: 'invalid',
-              url: 'https://echo.example.com/sse',
-            },
+let mockJsonWatcher: JsonWatcher<McpSettings>;
```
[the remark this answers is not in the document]
manusa: In general, I'd strongly advise against that because you're not protecting the tests from side-effects.

Anyway, changed as per your request.

jeffmaury (approved):
LGTM

manusa:
Is there anything missing from my side?

PR #2978, opened 2025-05-08
fix(playground): duplicate chat messages can be displayed

manusa:
### What does this PR do?

When a Chat message contains lines with the same content (or multiple empty lines), Svelte fails because it contains duplicate keys.

This fix adds an artificial key based on the index of the paragraph. Maybe completely replacing getMessageParagraphs with something else can be considered in the future.

### Screenshot / video of UI

image

n/a

### How to test this PR?

Create a playground and ask the model to repeat something, for example `Repeat the following "Hello" 10 times, once per line`.

### DIFF
--- a/packages/frontend/src/lib/conversation/ChatMessage.svelte
+++ b/packages/frontend/src/lib/conversation/ChatMessage.svelte
@@ -35,7 +35,7 @@ function getMessageParagraphs(message: ChatMessage): string[] {
     class:bg-[var(--pd-content-card-inset-bg)]={isAssistantChat(message)}
     class:ml-8={isAssistantChat(message)}
     class:mr-8={isUserChat(message)}>
-    {#each getMessageParagraphs(message) as paragraph (paragraph)}
+    {#each getMessageParagraphs(message) as paragraph, idx (`${idx}-paragraph`)}
       <p>{paragraph}</p>
     {/each}
   </div>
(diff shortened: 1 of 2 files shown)


benoitf (approved):

jeffmaury (approved):
LGTM

packages/frontend/src/lib/conversation/ChatMessage.svelte
```diff
@@ -35,7 +35,7 @@ function getMessageParagraphs(message: ChatMessage): string[] {
     class:bg-[var(--pd-content-card-inset-bg)]={isAssistantChat(message)}
     class:ml-8={isAssistantChat(message)}
     class:mr-8={isUserChat(message)}>
-    {#each getMessageParagraphs(message) as paragraph (paragraph)}
+    {#each getMessageParagraphs(message) as paragraph, idx (`${idx}-paragraph`)}
```
axel7083, line 38:
```suggestion
    {#each getMessageParagraphs(message) as paragraph, index (index)}
```

Probably just use the index as key (we should not introduce new name such as idx, we usually use `index`).

packages/frontend/src/pages/Playground.spec.ts
```diff
@@ -314,7 +314,7 @@ test('sending prompt should display the prompt and the response', async () => {
         {
           role: 'assistant',
           id: 'message-2',
-          content: 'a response from the assistant',
+          content: 'a response from the assistant\neat, sleep, code, repeat\neat, sleep, code, repeat',
```
axel7083, line 317: Thanks for the regression test 👍

packages/frontend/src/lib/conversation/ChatMessage.svelte
manusa: :flushed: there was a typo there, original intention was to have `${idx}-${paragraph}` to not diverge from the original.

Anyway, I've updated as per your suggestions.

packages/frontend/src/pages/Playground.spec.ts
manusa: np :wink: 

I was wondering to create a separate test, but wanted to get the fix in ASAP. Might come up in a future PR :shrug:

axel7083 (approved):

Issue #2614, opened 2025-03-04
Implementation of Displaying used Inferencing Runtime


slemeur:
#### Playgrounds

As a user, I should be able to identify what is the inferencing runtime used by a playground. 
We should be adding a "label" that will materialize.

<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/07cbf9cb-18bc-49df-a7bc-d40fa0ce9a2c" />

Using a similar pattern than when we are indicating the environment in the pods page of podman desktop. 

<img width="185" alt="Image" src="https://github.com/user-attachments/assets/f5f8ee06-7495-42a1-a70a-1bfa49963882" />

#### Model Service

As a user, I should be able to identify what is the inferencing runtime used by a model service. 
We need to extend the "Inference Server" to add indication about it.

##### Details of a model service
<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/380871d3-5a69-4133-b5f3-495bc3e2a50d" />

##### List of services

<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/72a6d705-bf2b-44f5-9ab6-de7199c9374f" />

#### Recipes

As a user, I should be able to identify what is the inferencing runtime used by a recipe. 
We should be adding a "label" that will materialize.

<img width="1435" alt="Image" src="https://github.com/user-attachments/assets/2b12e03e-ad47-458e-a7c9-fa75b976129f" />

Using a similar pattern than when we are indicating the environment in the pods page of podman desktop. 

<img width="185" alt="Image" src="https://github.com/user-attachments/assets/f5f8ee06-7495-42a1-a70a-1bfa49963882" />

### Describe the solution you'd like

Implementation of pattern defined in https://github.com/containers/podman-desktop-extension-ai-lab/issues/2503

feloy:
Setting as On hold, needs https://github.com/containers/podman-desktop-extension-ai-lab/pull/2893

MariaLeonova:
@vancura I think runtime is a valid thing to show, as the user should be aware of what they are using and be able to change it for separate entities. I believe it very likely that they would have a mix, but I think this is worth a discussion at a UX call.

PR #3005, opened 2025-05-14
display runtime for running inference servers

feloy:
### What does this PR do?

Display runtime for running inference servers

### Screenshot / video of UI

display-inference-server-runtime

Part of #2614 

### How to test this PR?

Start inference servers with different backends (llama.cpp / whisper.cpp), and check that the runtime is correctly displayed in the list

### DIFF
--- a/packages/frontend/src/lib/table/service/ServiceColumnRuntime.svelte
+++ b/packages/frontend/src/lib/table/service/ServiceColumnRuntime.svelte
@@ -1,0 +1,8 @@
+<script lang="ts">
+import { inferenceTypeLabel, type InferenceServer } from '@shared/models/IInference';
+import Badge from '../../Badge.svelte';
+
+export let object: InferenceServer;
+</script>
+
+<Badge content={inferenceTypeLabel(object.type)} />
--- a/packages/shared/src/models/IInference.ts
+++ b/packages/shared/src/models/IInference.ts
@@ -23,6 +23,19 @@ export enum InferenceType {
   NONE = 'none',
 }
 
+const InferenceTypeLabel = {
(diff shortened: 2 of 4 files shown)


packages/frontend/src/lib/table/service/ServiceColumnRuntime.svelte
```diff
@@ -0,0 +1,8 @@
+<script lang="ts">
+import { InferenceTypeLabel, type InferenceServer } from '@shared/models/IInference';
+import Badge from '../../Badge.svelte';
+
+export let object: InferenceServer;
```
gastoner, line 5: Cant we use Props here?

packages/shared/src/models/IInference.ts
```diff
@@ -23,6 +23,13 @@ export enum InferenceType {
   NONE = 'none',
 }
 
+export const InferenceTypeLabel = {
+  'llama-cpp': 'llamacpp',
+  'whisper-cpp': 'whispercpp',
+  openvino: 'OpenVino',
```
gastoner, line 29: Openvino is already supported?

feloy: wrong rebase, I'll remove it. thanks

packages/frontend/src/lib/table/service/ServiceColumnRuntime.svelte
```diff
@@ -0,0 +1,8 @@
...
+</script>
+
+<Badge content={InferenceTypeLabel[object.type]} />
```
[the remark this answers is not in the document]
feloy: right, I changed

gastoner (approved):
LGTM, tested on Linux

Issue #2937, opened 2025-04-30
Unexpected Can't find model info for local folder message at startup


jeffmaury:
### Bug description

When extension start models manager logs this message

### Operating system

Win11Pro

### Installation Method

None

### Version

next (development version)

### Relevant log output

```shell

```

PR #3013, opened 2025-05-15
feat: display application runtime

feloy:

### What does this PR do?

Display application runtime

The models manager needs to be initialized synchronously, so the app manager has an already initialized models catalog (or depending on race condition, it is possible that the model's backend is not found)

### Screenshot / video of UI
display-app-runtime

Part of #2614 

### How to test this PR?

Start a recipe and check that the runtime is correctly displayed
Restart (PD, extension, container runtime, etc) and check that the runtime is still correct

### DIFF
--- a/packages/backend/src/managers/application/applicationManager.ts
+++ b/packages/backend/src/managers/application/applicationManager.ts
@@ -54,6 +54,7 @@ import { VMType } from '@shared/models/IPodman';
 import { RECIPE_START_ROUTE } from '../../registries/NavigationRegistry';
 import type { RpcExtension } from '@shared/messages/MessageProxy';
 import { TaskRunner } from '../TaskRunner';
+import { getInferenceType } from '../../utils/inferenceUtils';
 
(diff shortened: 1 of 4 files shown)


gastoner (approved):
LGTM on Linux

jeffmaury (commented):
For me this PR addresses 2 concerns:
- models manager loading
- application runtime

The first part is not complete as it is caused by the caatlog tio be loaded asynchroniusly (see #2937)

Can we split it into 2 PRs ?

feloy:
> For me this PR addresses 2 concerns:
> 
> * models manager loading
> * application runtime
> 
> The first part is not complete as it is caused by the caatlog tio be loaded asynchroniusly (see #2937)
> 
> Can we split it into 2 PRs ?

Right, I removed the "models manager loading" part, I'll create a new PR with this part

feloy:
@jeffmaury is it ok as this?

jeffmaury (approved):
LGTM

PR #3015, opened 2025-05-15
feat: display runtime in service details

### LINKED ISSUE #2614: Implementation of Displaying used Inferencing Runtime

feloy:

### What does this PR do?

Display runtime in service details

### Screenshot / video of UI

display-service-details-runtime

Fixes #2614

### How to test this PR?

Start an inference server and check in the details page that the runtime is correct

### DIFF
--- a/packages/frontend/src/pages/InferenceServerDetails.spec.ts
+++ b/packages/frontend/src/pages/InferenceServerDetails.spec.ts
@@ -283,6 +283,15 @@ test('ensure models to be clickable', async () => {
   expect(a.getAttribute('href')).toBe('/model/dummyModelId');
 });
 
+test('runtime label is displayed', async () => {
+  render(InferenceServerDetails, {
+    containerId: 'dummyContainerId',
+  });
+
+  const span = screen.getByText('llamacpp');
+  expect(span).toBeDefined();
+});
+
 describe('labels', () => {
   test('GPU label should display GPU Inference', async () => {
     mocks.getInferenceServersMock.mockReturnValue([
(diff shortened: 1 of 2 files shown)


gastoner (commented):
Question: Is it expected that when the service is stopped, the label is not visible?
I would also try to make it consistent with other labels using the tooltip and div combination if it is possible to avoid having two different height of the labels


[Screencast_20250515_101113.webm](https://github.com/user-attachments/assets/61e600f9-24a8-43e6-b2c8-2785e57f8e02)

jeffmaury (approved):
LGTM

feloy:
> Question: Is it expected that when the service is stopped, the label is not visible?

Yes, it is expected. It is based on the design, which displays it in the `Inference Endpoint URL` block, which is visible only when the service is running

I would also try to make it consistent with other labels using the tooltip and div combination if it is possible to avoid having two different height of the labels
> 
>  Screencast_20250515_101113.webm

I'll change the design

Issue #2624, opened 2025-03-04
LLama-stack recipes


slemeur:
### Epic domain

As Podman AI Lab provides samples implementation for a lot of use cases involving LLM, we would like to extend the catalog to showcase samples using LLama-stack

PR #3016, opened 2025-05-15
feat: start a recipe with llama-stack backend

feloy:

### What does this PR do?

This PR is a first step towards being able to run a recipe using llama stack as backend.

This PR covers:
- AI Lab supports recipes with `backend="llama-stack"
- for such recipe, no model is asked in the "Start Recipe" form
- when AI Lab deploys a recipe with llama-stack backend, it does not deploy any inference server
- the MODEL_ENDPOINT passed to the recipe is the endpoint of Llama Stack
- all models served by Llama stack are accessible to the app. It is the responsability of the app to choose with which model to work

Not covered by this PR:
- the user needs to start the llama-stack manually, it is not yet started during the Recipe startup
- once started, the recipe is not listed in the list of running recipes, and the details page is not accessible (the user needs to go to the container details to get the port on which to access the app)

### Screenshot / video of UI

https://github.com/user-attachments/assets/928feef2-64bc-44e4-a5b6-c239b2134547

Part of #2624 

### How to test this PR?

1. Add this recipe to your `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/user-catalog.json` file:

```
{
  "version": "1.0",
  "recipes": [
    {
      "id": "chatbot-llama-stack",
      "description": "This recipe provides a blueprint for developers to create their own AI-powered chat applications using Streamlit and llama-stack.",
      "name": "ChatBot using Llama Stack",
      "repository": "https://github.com/feloy/chatbot-llama-stack-recipe",
      "ref": "main",
      "icon": "natural-language-processing",
      "categories": ["natural-language-processing"],
      "basedir": "/",
      "readme": "# Chat Application using Llama Stack\n",
      "backend": "llama-stack",
      "languages": ["python"],
      "frameworks": ["streamlit", "llama-stack"]
    }
  ],
  "models": [
    {
      "id": "openvino-mistral-7B-instruct-v0.2-int4-ov",
      "name": "OpenVINO/mistral-7B-instruct-v0.2-int4-ov",
      "description": "OpenVINO/mistral-7B-instruct-v0.2-int4-ov",
      "url": "huggingface:/OpenVINO/mistral-7B-instruct-v0.2-int4-ov",
      "backend": "openvino"
    }
  ],
  "categories": []
}
```

2. Start the Llama Stack from AI Lab
3. Go to the Recipes catalog and start the "ChatBot using Llama Stack" recipe
4. Go to the containers details of the newly created container running the app
5. Get the port of the container
6. Navigate to `http://localhost:$port`


packages/backend/src/managers/application/applicationManager.ts
```diff
@@ -176,45 +179,52 @@ export class ApplicationManager extends Publisher<ApplicationState[]> implements
...
-    // upload model to podman machine if user system is supported
-    if (!recipeComponents.inferenceServer) {
-      modelPath = await this.modelsManager.uploadModelToPodmanMachine(connection, model, {
-        ...labels,
-        'recipe-id': recipe.id,
-        'model-id': model.id,
-      });
+    if (model) {
+      // upload model to podman machine if user system is supported
+      if (!recipeComponents.inferenceServer) {
```
axel7083, line 209: suggestion: use a single if `model && !recipeComponents.inferenceServer`

packages/backend/src/managers/application/applicationManager.ts
```diff
@@ -176,45 +179,52 @@ export class ApplicationManager extends Publisher<ApplicationState[]> implements
   private async initApplication(
     connection: ContainerProviderConnection,
     recipe: Recipe,
-    model: ModelInfo,
+    model?: ModelInfo,
     labels: Record<string, string> = {},
   ): Promise<PodInfo> {
+    const modelId = model ? model.id : '0';
```
axel7083, line 185: I don't really like that

feloy: Sure, me neither:
> Note: I'm not very happy with the way the applicationManager functions accept an optional model, depending on if we want to use llama-stack or not. Let's first discuss the high-level architecture of this solution, I can change these implementation details when we agree on the architecture

axel7083: Sorry I went to code too fast :pray:  

> Let's first discuss the high-level architecture of this solution,

What depends on the `modelId`? We use it for operation like `hasApplicationPod` & `removeApplication`. Maybe we should only use the recipeId as key?

feloy: I'm not very comfortable changing the labels, as the model label is also used for the tasks (the user can start the same recipe with different models, and we want to differentiate the tasks for them).

I have pushed a new commit changing how the parameters are passed: having explicit options inferModel / startLlamaStack, which enforce the other parameters (the `model` parameter for the moment). I also changed this arbitrary `0` model Id to `<none>`

gastoner:
Please add "llama-stack" here https://github.com/containers/podman-desktop-extension-ai-lab/blob/main/packages/frontend/src/lib/RecipeCardTags.ts#L25 so it works with other tags (I think this should be only change to make it work)

packages/backend/src/managers/recipes/RecipeManager.ts
```diff
@@ -96,73 +96,70 @@ export class RecipeManager implements Disposable {
...
   }
 
-  public async buildRecipe(
-    connection: ContainerProviderConnection,
-    recipe: Recipe,
-    model: ModelInfo,
-    labels?: { [key: string]: string },
-  ): Promise<RecipeComponents> {
-    const localFolder = path.join(this.appUserDirectory, recipe.id);
+  public async buildRecipe(options: ApplicationOptions, labels?: { [key: string]: string }): Promise<RecipeComponents> {
```
axel7083, line 99: why not including the labels in the options? feel weird to move everything inside ApplicationOptions but the labels?

packages/backend/src/studio-api-impl.ts
```diff
@@ -240,7 +241,26 @@ export class StudioApiImpl implements StudioAPI {
 
     if (!connection) throw new Error('no running container provider connection found.');
 
-    return this.applicationManager.requestPullApplication(connection, recipe, model);
+    let model: ModelInfo | undefined;
+    if (isRecipePullOptionsWithModelInference(options)) {
```
axel7083, line 245: why do we need `isRecipePullOptionsWithLlamaStack` and `isRecipePullOptionsWithModelInference` when you introduced the `startLlamaStack` boolean ?

> isRecipePullOptionsWithLlamaStack seems to have no usage?

packages/backend/src/models/ApplicationOptions.ts
```diff
@@ -0,0 +1,51 @@
...
+  inferModel: boolean;
+  startLlamaStack: boolean;
+  connection: ContainerProviderConnection;
+  recipe: Recipe;
+}
+
+export interface ApplicationOptionsWithModelInference extends ApplicationOptionsCommon {
+  inferModel: true;
+  model: ModelInfo;
+}
```
axel7083, line 35:
```suggestion
type ApplicationOptionsWithModelInference = ApplicationOptionsCommon & {
  inferModel: true;
  model: ModelInfo;
}
```
Feel like this should be a type, not really fan of extending custom interface

packages/backend/src/managers/recipes/RecipeManager.ts
feloy: First reason is because labels is not part of the parameters of `requestPullApplication`, and I would like to keep the same options structure everywhere

feloy:
> This is pretty hard to make the difference between the refactoring required, and the actual code needed to support llama-stack
> 
> You updated the signature of the applicationManager class, but this would probably ease the review to have this in a separate refactor PR 🙏

OK, I'll make a first PR with refactoring only

Issue #2563, opened 2025-02-20
Add Function Calling Demo AI App installation test and interaction


ScrewTSW:
### Is your enhancement related to a problem? Please describe

Currently this test case has to be performed manually

### Describe the solution you'd like

* Add Function Calling Demo AI App to installation and verification list
* Implement custom interaction with the AI App service

PR #3046, opened 2025-05-20
chore(test): function calling test

amisskii:
### What does this PR do?

 Adds Function Calling Demo AI App to installation and verification list
 Implements custom interaction with the AI App service

https://github.com/containers/podman-desktop-extension-ai-lab/issues/2563

### DIFF
--- a/tests/playwright/src/ai-lab-extension.spec.ts
+++ b/tests/playwright/src/ai-lab-extension.spec.ts
@@ -418,7 +418,7 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
     });
   });
 
-  ['Audio to Text', 'ChatBot', 'Summarizer', 'Code Generation', 'RAG Chatbot'].forEach(appName => {
+  ['Audio to Text', 'ChatBot', 'Summarizer', 'Code Generation', 'RAG Chatbot', 'Function calling'].forEach(appName => {
     test.describe.serial(`AI Recipe installation`, () => {
       test.skip(
         !process.env.EXT_TEST_RAG_CHATBOT && appName === 'RAG Chatbot',
@@ -426,7 +426,7 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
       );
       let recipesCatalogPage: AILabRecipesCatalogPage;
 
-      test.beforeEach(`Open Recipes Catalog`, async ({ runner, page, navigationBar }) => {
+      test.beforeAll(`Open Recipes Catalog`, async ({ runner, page, navigationBar }) => {
         aiLabPage = await reopenAILabDashboard(runner, page, navigationBar);
         await aiLabPage.navigationBar.waitForLoad();
 
@@ -441,8 +441,53 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
         await demoApp.startNewDeployment();
       });
 
-      test.afterEach(`Stop ${appName} app`, async ({ navigationBar }) => {
+      test(`Verify that model service for the ${appName} is working`, async ({ request }) => {
+        test.skip(appName !== 'Function calling');
+        test.setTimeout(600_000);
+
+        const modelServicePage = await aiLabPage.navigationBar.openServices();
+        const serviceDetailsPage = await modelServicePage.openServiceDetails(
+          'ibm-granite/granite-3.3-8b-instruct-GGUF',
+        );
+
+        await playExpect
+          // eslint-disable-next-line sonarjs/no-nested-functions
+          .poll(async () => await serviceDetailsPage.getServiceState(), { timeout: 60_000 })
+          .toBe('RUNNING');
+        const port = await serviceDetailsPage.getInferenceServerPort();
+        const url = `http://localhost:${port}/v1/chat/completions`;
+
+        const response = await request.post(url, {
+          data: {
+            messages: [
+              {
+                content: 'You are a helpful assistant.',
+                role: 'system',
+              },
+              {
+                content: 'What is the capital of Czech Republic?',
+                role: 'user',
+              },
+            ],
+          },
+          timeout: 600_000,
+        });
+
+        playExpect(response.ok()).toBeTruthy();
+        const body = await response.body();
+        const text = body.toString();
+        playExpect(text).toContain('Prague');
+      });
+
+      test(`Stop ${appName} app`, async () => {
         test.setTimeout(150_000);
+        await stopAndDeleteApp(appName);
+        await cleanupServiceModels();
+      });
+
+      test.afterAll(`Ensure cleanup of "${appName}" app, related service, and images`, async ({ navigationBar }) => {
+        test.setTimeout(150_000);
+
         await stopAndDeleteApp(appName);
         await cleanupServiceModels();
         await deleteUnusedImages(navigationBar);
@@ -465,6 +510,10 @@ async function cleanupServiceModels(): Promise<void> {
 async function stopAndDeleteApp(appName: string): Promise<void> {
   const aiRunningAppsPage = await aiLabPage.navigationBar.openRunningApps();
   await aiRunningAppsPage.waitForLoad();
+  if (!(await aiRunningAppsPage.appExists(appName))) {
+    console.log(`"${appName}" is not present in the running apps list. Skipping stop and delete operations.`);
+    return;
+  }
   await playExpect.poll(async () => await aiRunningAppsPage.appExists(appName), { timeout: 10_000 }).toBeTruthy();
   await playExpect
     .poll(async () => await aiRunningAppsPage.getCurrentStatusForApp(appName), { timeout: 60_000 })
--- a/tests/playwright/src/model/ai-lab-model-service-page.ts
+++ b/tests/playwright/src/model/ai-lab-model-service-page.ts
@@ -21,6 +21,7 @@ import type { Locator, Page } from '@playwright/test';
 import { AILabBasePage } from './ai-lab-base-page';
 import { handleConfirmationDialog, podmanAILabExtension } from '@podman-desktop/tests-playwright';
 import { AILabCreatingModelServicePage } from './ai-lab-creating-model-service-page';
+import { AILabServiceDetailsPage } from './ai-lab-service-details-page';
 
 export class AiModelServicePage extends AILabBasePage {
   readonly additionalActions: Locator;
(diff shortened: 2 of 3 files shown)


tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -461,7 +461,43 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
         await demoApp.startNewDeployment();
       });
 
-      test.afterEach(`Stop ${appName} app`, async ({ navigationBar }) => {
+      /* eslint-disable sonarjs/no-nested-functions */
+      if (appName === 'Function calling') {
+        test(`Verify that model service for the ${appName} is working`, async ({ request }) => {
+          test.setTimeout(600_000);
+          const modelServicePage = await aiLabPage.navigationBar.openServices();
+          const serviceDetailsPage = await modelServicePage.openSercviceDetails(
```
gastoner, line 469: openSercviceDetails => openServiceDetails ?

tests/playwright/src/model/ai-lab-model-service-page.ts
```diff
@@ -66,6 +67,27 @@ export class AiModelServicePage extends AILabBasePage {
...
+      throw new Error(`Model [${modelName}] service doesn't exist`);
+    }
+    const serviceRowName = serviceRow.getByRole('cell').nth(3);
+    await serviceRowName.click();
+    return new AILabServiceDetailsPage(this.page, this.webview);
+  }
+
+  async getServiceByModel(modelName: string): Promise<Locator | undefined> {
+    const rows = await this.getAllTableRows();
+    for (let rowNum = 1; rowNum < rows.length; rowNum++) {
```
gastoner, line 82: Is there a reason why we start from 1 and not from 0? I would rather start from 0 and had +1 somewhere else :D but not required

tests/playwright/src/model/ai-lab-service-details-page.ts
```diff
@@ -56,4 +56,12 @@ export class AILabServiceDetailsPage extends AILabBasePage {
     const port = split ? split[split.length - 1].split('/')[0] : '';
     return port;
   }
+
+  async getServiceState(): Promise<string> {
+    const serviceState = await this.webview.getByRole('status').getAttribute('title');
+    if (serviceState === 'RUNNING') {
+      return 'RUNNING';
+    }
+    return serviceState ?? 'UNKNOWN';
```
gastoner, line 65: Isn't this `if` redundant? Like: 
1. service state === "RUNNING" => I get "RUNNING"
2. service state === "FOO" => I get "FOO",
3. service state === undefined => I get "UNKNOWN"

I think you can skip the whole if statement

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -437,15 +437,15 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
...
+  ['Audio to Text', 'ChatBot', 'Summarizer', 'Code Generation', 'RAG Chatbot', 'Function calling'].forEach(appName => {
     test.describe.serial(`AI Recipe installation`, () => {
       test.skip(
         !process.env.EXT_TEST_RAG_CHATBOT && appName === 'RAG Chatbot',
         'EXT_TEST_RAG_CHATBOT variable not set, skipping test',
       );
       let recipesCatalogPage: AILabRecipesCatalogPage;
 
-      test.beforeEach(`Open Recipes Catalog`, async ({ runner, page, navigationBar }) => {
+      test(`Navigate to Recipes Catalog for ${appName}`, async ({ runner, page, navigationBar }) => {
```
cbr7, line 448: Why was this change needed?

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -461,7 +461,43 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
         await demoApp.startNewDeployment();
       });
 
-      test.afterEach(`Stop ${appName} app`, async ({ navigationBar }) => {
+      /* eslint-disable sonarjs/no-nested-functions */
+      if (appName === 'Function calling') {
+        test(`Verify that model service for the ${appName} is working`, async ({ request }) => {
```
cbr7, line 466: If this is a test is that only for function calling app then please do not nest it inside an if statement, what you should do is make it a normal test and check the validation condition inside a test.skip() statement.

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -461,7 +461,43 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
...
+
+          playExpect(response.ok()).toBeTruthy();
+          const body = await response.body();
+          const text = body.toString();
+          playExpect(text).toContain('Prague');
+        });
+      }
+      /* eslint-enable sonarjs/no-nested-functions */
+
+      test(`Stop ${appName} app`, async ({ navigationBar }) => {
```
cbr7, line 500: Again as with the beforeEach, changing the afterEach to a test has functional consequences, is there a particular reason you've changed this structure?

tests/playwright/src/model/ai-lab-service-details-page.ts
cbr7: Agreed with @gastoner, the if is not necessary.

cbr7 (changes requested):
Left some comments above.

tests/playwright/src/model/ai-lab-model-service-page.ts
amisskii: There, I skip the header row in the table, as it doesn't contain any model services. I believe we use the same approach in the podman-desktop repository.

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -485,6 +527,9 @@ async function cleanupServiceModels(): Promise<void> {
 async function stopAndDeleteApp(appName: string): Promise<void> {
   const aiRunningAppsPage = await aiLabPage.navigationBar.openRunningApps();
   await aiRunningAppsPage.waitForLoad();
+  if (!(await aiRunningAppsPage.appExists(appName))) {
+    console.log(`"${appName}" is not present in the running apps list. Skipping stop and delete operations.`);
+  }
   await playExpect.poll(async () => await aiRunningAppsPage.appExists(appName), { timeout: 10_000 }).toBeTruthy();
```
gastoner, line 533: I see 2 times `await aiRunningAppsPage.appExists(appName)` can't we extract it to const above the `if` and use it in 530 and 533?

cbr7: @gastoner that would not actually work, the assert statement (the one with playExpect.poll requires the actual method because what it does it cyclically call it in order to check it's value until the value changes if we'd assign the result to constant and check that in the assert that value would never change the it would completely break the assert functionality.

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -485,6 +527,9 @@ async function cleanupServiceModels(): Promise<void> {
 async function stopAndDeleteApp(appName: string): Promise<void> {
   const aiRunningAppsPage = await aiLabPage.navigationBar.openRunningApps();
   await aiRunningAppsPage.waitForLoad();
+  if (!(await aiRunningAppsPage.appExists(appName))) {
+    console.log(`"${appName}" is not present in the running apps list. Skipping stop and delete operations.`);
```
cbr7, line 531: is there a `return` statement that is missing from this if block? Because otherwise I don't see where the skip that the console.log is talking about is happening.

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -461,10 +461,52 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
...
+        playExpect(text).toContain('Prague');
+      });
+
+      test(`Stop ${appName} app`, async () => {
         test.setTimeout(150_000);
         await stopAndDeleteApp(appName);
         await cleanupServiceModels();
+      });
+
+      test.afterAll(`Ensure cleanup of "${appName}" app, related service, and images`, async ({ navigationBar }) => {
```
cbr7, line 505: This does not seem correct afterAll gets only run once after all the test instances, unlike afterEach which gets run after each test, this will only clean up the last test in reality, not all the apps that were deployed. Pleae revert to afterEach.

tests/playwright/src/ai-lab-extension.spec.ts
gastoner: Thanks for explain, didnt know that :D

tests/playwright/src/ai-lab-extension.spec.ts
```diff
@@ -437,15 +437,15 @@ test.describe.serial(`AI Lab extension installation and verification`, () => {
...
+  ['ChatBot', 'Function calling'].forEach(appName => {
     test.describe.serial(`AI Recipe installation`, () => {
       test.skip(
         !process.env.EXT_TEST_RAG_CHATBOT && appName === 'RAG Chatbot',
         'EXT_TEST_RAG_CHATBOT variable not set, skipping test',
       );
       let recipesCatalogPage: AILabRecipesCatalogPage;
 
-      test.beforeEach(`Open Recipes Catalog`, async ({ runner, page, navigationBar }) => {
+      test.beforeAll(`Open Recipes Catalog`, async ({ runner, page, navigationBar }) => {
```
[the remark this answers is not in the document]
odockal: This should be fine, no?
We should end up with:
```
iterationX (X = [chatbot, funtion call]):
  describe.serial start
    beforeAll
      testInstall
      testService (optional)
      testStop
    afterAll
  describe end
  ```
  
  Doing it with before/afterEach would do something like:
  ```
  iterationX (X = [chatbot, funtion call]):
  describe.serial start
    beforeEach
      testInstall
    AfterEach
    BeforeEach
      testService (optional)
    AfterEach
    BeforeEach
      testStop
    afterEach
  describe end
  ```
  
  Which wont work?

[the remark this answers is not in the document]
Comment (cbr7) [reply tests/playwright/src/ai-lab-extension.spec.ts:448]: I'm honestly not sure what flow you are describing, but just taking a quick look at the afterAll, all it will do is uninstall the last app, any previous app will not be uninstalled, unless I am missing something.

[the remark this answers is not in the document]
Comment (amisskii) [reply tests/playwright/src/ai-lab-extension.spec.ts:448]: The `afterAll `hook will run for each AI app because it is defined inside a `test.describe.serial` block that is created separately for each app using .`forEach`. The `afterEach` hook runs after every test case inside` test.describe.serial` and will attempt to delete the AI app after each test — which we don't want. This approach worked previously because we only had beforeEach, a single install test, and afterEach.

File: tests/playwright/src/model/ai-lab-model-service-page.ts [hunk 11]
```diff
@@ -66,6 +67,28 @@ export class AiModelServicePage extends AILabBasePage {
     return (await this.getAllTableRows()).length;
   }
 
+  async openServiceDetails(modelName: string): Promise<AILabServiceDetailsPage> {
+    const serviceRow = await this.getServiceByModel(modelName);
+    if (serviceRow === undefined) {
+      throw new Error(`Model [${modelName}] service doesn't exist`);
+    }
+    const serviceRowName = serviceRow.getByRole('cell').nth(3);
```
odockal, line 75: Please open follow up issue to address missing aria-labels in the row definition. We should not rely on order of the cells. This is out of scope of this PR.

PR #3065, opened 2025-05-22
feat(playground): mcp servers listed

manusa:
### What does this PR do?

Adds a UI element to be able to (later) activate/deactivate MCP Servers and their specific tools (even later).

This is the first iteration just to validate that the UI is aligned with the current expectation.
The next immediate iteration would be to add functionality to the checkboxes of the servers.
With this regard it would be good to know what effect we want for them.
For example, do we want them to be mapped to the file entries and update them there?

### Screenshot / video of UI

https://github.com/user-attachments/assets/0597fa21-759f-42fe-a867-60158f22cec8

Part of #2885

### How to test this PR?

- Create or update a file `~/.local/share/containers/podman-desktop/extensions-storage/redhat.ai-lab/mcp-settings.json`.
- Set the following content to the file:
  ```json
  {
    "servers": {
      "time": {
        "enabled": true,
        "type": "stdio",
        "command": "podman",
      }
    }
  }
  ```
- Create a new playground.
- A new button with label `MCP Servers` should be visible in the Settings pane.
- Press the button
- A modal window should be visible with the list of servers and a checkbox nest to each entry.
  - The checkbox has no effect at the moment.
- By modifying the file: enabling/disabling entries, adding new entries, etc.
  - Entries in the list should get automatically updated
  - If no entries are enabled, the button shouldn't be visible.


packages/backend/src/studio.ts
```diff
@@ -213,6 +213,7 @@ export class Studio {
      */
     this.#catalogManager = new CatalogManager(this.#rpcExtension, appUserDirectory);
     this.#catalogManager.init();
+    this.#extensionContext.subscriptions.push(this.#catalogManager);
```
manusa, line 216: This was missing, it's my understanding that it should be pushed to the subscriptions for disposal.

jeffmaury: It is now optional as every Disposable it automatically disposed when extension is stopped. Should be moved to a separate PR

manusa: I've removed the _missing_ entries.
However, I'm having a hard time understanding how are the `Disposable`s discovered by the extension for automatic disposal.

axel7083: Yeah, for extension-defined disposable, it should be disposed by the extension. The disposable provided by the extension-api are disposed by the core, but I think it is always a good idea to do it, for good practice.

bmahabirbu:
LGTM codewise as well! probs needs a rebase

manusa:
Hi, sorry for the delayed reply, it's been a really busy week

> the style for the MCP servers is not aligned with other sections titles in the left pane

image
Maybe an optical effect? or are you referring to something else?

> when clicking on MCP servers, the UI is displayed in the middle of the page which seems strange to me

Where would you want to locate it. Maybe top of the screen?

> once the MCP servers are displayed, clicking Escape to close will close the playground and return to the list of playgrounds

Looking into it.

PR #3156, opened 2025-06-10
feat: Adds the ability to select the runtime and inference service for playground

bmahabirbu:
### What does this PR do?
see above

thanks to @axel7083 for his work so far in https://github.com/containers/podman-desktop-extension-ai-lab/pull/3077
I think ill be able to use [InferenceRuntimeSelect.svelte](https://github.com/containers/podman-desktop-extension-ai-lab/pull/3077/files#diff-35d0667062ebbcec62d01a2e16b25efb5764007c3b081e8bc5f038b88b3debce) to help select the runtime and pass it to the playground creation IF there is no model services available to select. If so i think ill just point the user to create the service in the model services page

### Screenshot / video of UI

so far these are pics of selecting the inference provider
Screenshot 2025-06-10 011247

fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2613

### How to test this PR?

create multiple services with the same model and see if you can choose from it in the playground

### DIFF
--- a/packages/backend/src/managers/playgroundV2Manager.ts
+++ b/packages/backend/src/managers/playgroundV2Manager.ts
@@ -33,6 +33,7 @@ import { AiStreamProcessor } from './playground/aiSdk';
 import { type McpServerManager } from './playground/McpServerManager';
 import type { ToolSet } from 'ai';
 import { simulateStreamingMiddleware, wrapLanguageModel } from 'ai';
+import type { InferenceServer } from '@shared/models/IInference';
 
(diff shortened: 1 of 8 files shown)


bmahabirbu:
Thanks @jeffmaury! Agreed I'll work on what you described in a separate pr.

Once that is done maybe we can revisit this idea later down the line once each model has multiple runtimes to pick from.

This maybe useful if a user has multiple services with different runtimes for a specific model they wanted to pick from.

PR #3171, opened 2025-06-12
feat: added ability to select runtime for playground

bmahabirbu:
### What does this PR do?
adds the ability to select the runtime for playground

### Screenshot / video of UI
runtime

fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2613


gastoner (commented):
I think we are missing `none` option. 
The facebook model has `none` runtime but the filter for it is not there
image

bmahabirbu:
@gastoner good catch I assumed "none" in this case was simply to have no selection. I did have a playground with resnet setup but since the model is used for object detection I didnt get the chat to work. Maybe we can keep it exluded

axel7083 (approved):
LGTM, tested on Windows, works as expected 👍

bmahabirbu:
@jeffmaury did we want to have a defaut model selected when the runtime is all (e.i no filter)? 

My current implementation doesnt have the model selected so I had to edit one of the e2e tests

axel7083:
PR checks need to be fixed

bmahabirbu:
image

I can use the preferred runtime settings instead for runtime select but in the inference runtime dropdown, I dont see OpenVINO. Shouldn't we have it as a runtime backend? We can download the openvino model and create a service for it so we should have it for the playground too!

packages/frontend/src/pages/PlaygroundCreate.svelte
```diff
@@ -14,9 +14,29 @@ import type { Unsubscriber } from 'svelte/store';
...
 import ModelSelect from '/@/lib/select/ModelSelect.svelte';
 import { InferenceType } from '@shared/models/IInference';
+import InferenceRuntimeSelect from '/@/lib/select/InferenceRuntimeSelect.svelte';
+import { configuration } from '../stores/extensionConfiguration';
+
+// Get recommended runtime
+let runtime: InferenceType | undefined = undefined;
+
+// Exlude certain runtimes from selection
+export let exclude: InferenceType[] = [InferenceType.NONE, InferenceType.WHISPER_CPP, InferenceType.OPENVINO];
```
jeffmaury, line 24: Why is OpenVINO be excluded ?

bmahabirbu: We don't want openvino for mac so I exclude it till we have a pr that updates the runtime select in settings with openvino

Issue #2925, opened 2025-04-25
AI Lab model's catalog UI doesn't reflect new model after successful API Pull


amisskii:
### Bug description

After successfully pulling a model using the OpenAI API, it does not appear as downloaded in the Podman Desktop application's AI Lab Catalog page. The model only becomes visible after restarting the application.

### Operating system

Fedora

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

-  Pull a model using the API endpoint:

   ```bash
   curl -X POST http://127.0.0.1:10434/api/pull \
     -H "Content-Type: application/json" \
     -H "Accept: application/json" \
     -d '{
       "model": "instructlab/granite-7b-lab-GGUF",
       "insecure": false,
       "stream": false
     }'

-  Open the AI Lab Catalog tab in the Podman Desktop application.
-  Observe that the model is not listed as downloaded.
-  Restart the Podman Desktop application.
- Open the AI Lab Catalog again — the model now appears as downloaded.

### Relevant log output

```shell

```

PR #3250, opened 2025-07-02
fix: update api/pull to update pd model catalog

bmahabirbu:
### What does this PR do?

The pr fixed api/pull when called outside podman desktop where it doesnt update the model catalog in real time

fixes https://github.com/containers/podman-desktop-extension-ai-lab/issues/2925

### How to test this PR?

Input this command and watch podman desktop ai lab extensions model catalog update to include the model
We shouldn't see the downloading circle either since its an api call 

Windows

```
curl --no-buffer -X POST http://127.0.0.1:10434/api/pull -H "Content-Type: application/json" -H "Accept: application/json" -d "{\"model\": \"facebook/detr-resnet-101\", \"insecure\": false, \"stream\": true}"
```
Linux/Mac
```
curl --no-buffer -X POST http://127.0.0.1:10434/api/pull \
  -H "Content-Type: application/json" \
  -H "Accept: application/json" \
  -d '{
    "model": "facebook/detr-resnet-101",
    "insecure": false,
    "stream": true
  }'
```


bmahabirbu:
We could also ```  private async updateModelInfos(): Promise<void> {```  remove the private for this as well but I beleive the private status is there for a reason so I chose to use the slightly more resource intensive loadLocalModels to update. 

At first, I tried using requestdownload instead of createdownloader, but we lose the streaming functionality and would require more time to fix testing but can be done as well! 

I thought this way might be the easiest

packages/backend/src/managers/apiServer.ts
```diff
@@ -342,7 +342,8 @@ export class ApiServer implements Disposable {
 
     downloader
       .perform(modelName)
-      .then(() => {
+      .then(async () => {
+        await this.modelsManager.loadLocalModels();
```
jeffmaury, line 346: loadLocalModels is a heavy operation that will scan the local disk and check the manager is update. I don't think this is required here, maybe calling `sendModelsInfo()` is enough ?

bmahabirbu: I tried sendModelsInfo but it didnt work. 

In modelsmanager downloadModel calls  updateModelInfos which calls getLocalModelsFromDisk for downloading a single model. so I kept the same functionality here as well!

Issue #2891, opened 2025-04-16
Update MAPT version in AI Lab pipelines


ScrewTSW:
### Is your enhancement related to a problem? Please describe

Currently MAPT version is outdated

### Describe the solution you'd like

Update to latest version

### Describe alternatives you've considered

N/A

### Additional context

N/A

odockal:
Better solution to replace the mapt version with repository variable: https://github.com/podman-desktop/e2e/issues/354.

PR #3276, opened 2025-07-09
chore(ci): update MAPT version to 0.9.3

ScrewTSW:
### What does this PR do?
* Udates MAPT from 0.7.4 to 0.9.3
* Pre-creates `.pulumi` folder with RWX permissions - was causing failure with permission denied for folder creation from within `podman run`

#2891

### DIFF
--- a/.github/workflows/ai-lab-e2e-nightly-windows.yaml
+++ b/.github/workflows/ai-lab-e2e-nightly-windows.yaml
@@ -87,7 +87,7 @@ jobs:
     name: windows-${{ matrix.windows-version }}-${{ matrix.windows-featurepack }}
     runs-on: ubuntu-latest
     env:
-      MAPT_VERSION: v0.7.4
+      MAPT_VERSION: v0.9.3
       MAPT_IMAGE: quay.io/redhat-developer/mapt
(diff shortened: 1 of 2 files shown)


odockal (changes requested):
Switching to repo variable is better solution long term. Otherwise ok.

cbr7 (approved):

.github/workflows/ai-lab-e2e-nightly-windows.yaml
```diff
@@ -81,15 +76,16 @@ on:
           - Standard_D8s_v4
           - Standard_E8as_v5
           - Standard_E4as_v5
+      mapt_params:
+        default: 'IMAGE=quay.io/redhat-developer/mapt,VERSION_TAG=v0.9.3,EXCLUDED_REGIONS=westindia,centralindia,southindia,australiacentral,australiacentral2,australiaeast,australiasoutheast,southafricanorth,southafricawest'
```
odockal, line 80: you probably need to wrap regions into double quotes.

odockal (changes requested):
only one required change in env_vars value for regions.

.github/workflows/ai-lab-e2e-nightly-windows.yaml
ScrewTSW: good point

odockal (approved):

PR #3641, opened 2025-09-29
chore(ci): windows pipelines apply defaults if inputs from workflow_dispatch are empty

ScrewTSW:
### What does this PR do?
* Fixes `workflow_dispatch` event to use `DEFAULTS` when the inputs are empty

N/A hotfix


jeffmaury (commented):
Wondering if you can use ${!VAR_NAME:-$DEFAULT_VALUE} notation ?

PR #3725, opened 2025-10-14
fix: addressed flakey e2e test for restart app

bmahabirbu:
There was a race condition for handling the dialog on restart for windows which caused the progressbar loading to be missed stalling the test locally I assume this is what is causing the flakey e2e tests on ci recently


gastoner (approved):
Codewise LGTM

PR #3726, opened 2025-10-14
fix: stop using nightly for e2e-main test

bmahabirbu:
instead of testing the pr on an older nightly build we should instead build the local ai lab image and install it for testing e2e that way we get e2e tests that are accurate to the pr wanting to change things


bmahabirbu:
@Mergifyio backport 1.9.x

PR #3742, opened 2025-10-17
fix(backport): backport fixes for 1.9.1 release

bmahabirbu:
### What does this PR do?

back ports 4 commits to fix 1.9.0 errors for e2e and granite model downloads


jeffmaury (changes requested):
Should be done with mergify and commit per commit

Issue #3990, opened 2025-12-19
Extension should check apiVersion if available instead of version


jeffmaury:
### Is your feature request related to a problem? Please describe

The extension checks for a minimal version using the version field. Core will add apiVersion so we should update the check

### Describe the solution you'd like

use apiVersion if available

PR #4137, opened 2026-01-27
fix: removed version check

bmahabirbu:
### What does this PR do?

This pr aims to fix the issue with the red hat build of podman desktop to use apiversion where available

Fixes: https://github.com/containers/podman-desktop-extension-ai-lab/issues/3990

### DIFF
--- a/packages/backend/src/studio.spec.ts
+++ b/packages/backend/src/studio.spec.ts
@@ -18,9 +18,9 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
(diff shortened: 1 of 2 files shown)


bmahabirbu:
@jeffmaury updated! 

We no longer needed the podmanversion check either, since it was used for css styling for the labels in the recipe card for an older version of pd

jeffmaury (commented):
Looks to me that there is some kind of refactoring done in RecipeCardsTags.ts that is unrelated ?

PR #4446, opened 2026-05-13
fix(ci): remove git change check on pr-check that was blocking linting on Linux

ScrewTSW:
### What does this PR do?
* Fixes `pnpm install` changing lockfile on every build and causing linux lint pipeline to fail

hotfix

https://github.com/containers/podman-desktop-extension-ai-lab/actions/runs/25788102500/job/75746395536
https://github.com/containers/podman-desktop-extension-ai-lab/actions/runs/25339379944/job/74292729290

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -59,15 +59,6 @@ jobs:
       - name: Run svelte check
         run: pnpm svelte:check
 
-      # Check we don't have changes in git
-      - name: Check no changes in git
-        if: ${{ matrix.os=='ubuntu-22.04'}}
-        run: |
-          if ! git diff --exit-code; then
-            echo "Found changes in git"
-            exit 1
-          fi
-
   e2e-pr-check:
     name: e2e tests smoke
     runs-on: ubuntu-24.04


jeffmaury (commented):
Other extension do not use this flag, why do we need it ?

ScrewTSW:
> Other extension do not use this flag, why do we need it ?

because we have a test on linux that is a required blocking test, that verifies whether there are any changes in git before build.
running plain `pnpm install` updates the lockfile every single time new dependency gets published, because we are including them via `^` instead of exact version. So it's either frozen-lockfile or concrete dependencies, otherwise this prcheck linter test will keep on failing every time. There is a linked run where this happens and it's been an issue with all of my past open PRs.

jeffmaury:
> > Other extension do not use this flag, why do we need it ?
> 
> because we have a test on linux that is a required blocking test, that verifies whether there are any changes in git before build. running plain `pnpm install` updates the lockfile every single time new dependency gets published, because we are including them via `^` instead of exact version. So it's either frozen-lockfile or concrete dependencies, otherwise this prcheck linter test will keep on failing every time. There is a linked run where this happens and it's been an issue with all of my past open PRs.

Yes but why do we need this here and not in other extensions ? Should it be the case for other extensions as well ?

ScrewTSW:
> > > Other extension do not use this flag, why do we need it ?
> > 
> > 
> > because we have a test on linux that is a required blocking test, that verifies whether there are any changes in git before build. running plain `pnpm install` updates the lockfile every single time new dependency gets published, because we are including them via `^` instead of exact version. So it's either frozen-lockfile or concrete dependencies, otherwise this prcheck linter test will keep on failing every time. There is a linked run where this happens and it's been an issue with all of my past open PRs.
> 
> Yes but why do we need this here and not in other extensions ? Should it be the case for other extensions as well ?

I assume that other extensions do not have this test at all https://github.com/containers/podman-desktop-extension-ai-lab/actions/runs/25788102500/workflow#L63

bmahabirbu:
Lets remove the test instead! that should fix the errors for now

odockal:
you have broken pnpm-lock, judging by 14 days failing Dependabot updates runs. I have opened an issue: https://github.com/containers/podman-desktop-extension-ai-lab/issues/4447

ScrewTSW:
> you have broken pnpm-lock, judging by 14 days failing Dependabot updates runs. I have opened an issue: #4447

pnpm-lock updates when you do pnpm install so this is not blocking, but it's a good point.

benoitf (changes requested):
should be kept

ScrewTSW:
fixed by https://github.com/containers/podman-desktop-extension-ai-lab/pull/4448

PR #4395, opened 2026-04-14
fix: resolve #2010 — Misleading message

ed-thuando:
## Summary

fix: resolve #2010 — Misleading message

## Problem

**Severity**: `Medium` | **File**: `packages/backend/src/managers/GPUManager.ts`

The "Update your Podman Machine to improve performance" message is shown unconditionally, but Podman Machine only exists on macOS and Windows. On Linux, users run Podman directly without a VM, making this message misleading. Need to add a check that only shows this message on platforms where Podman Machine is actually available.

## Changes

- `packages/backend/src/managers/GPUManager.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
*Generated by [ContribAI](https://github.com/tang-vu/ContribAI) v6.0.0*

PR #4680, opened 2026-08-15
fix: don't show Podman Machine message on Linux when a connection is provided

### LINKED ISSUE #2010: Misleading message
### Bug description

When I try to run the RAG AI lab, the following message is displayed:

'Update your Podman Machine to improve performance' 

Sounds reasonable, but I'm on linux and not running a podman machine.

### Operating system

Fedora 40

### Installation Method

Flathub (Linux)

### Version

1.12.x

### Steps to reproduce

Dunno if it's unique to my setup, but i saw it after running RAG AI lab

### Relevant log output

_No response_

### Additional context

_No response_

xhon-pelushi:
### What does this PR do?
On Linux, `checkContainerConnectionStatusAndResources()` only short-circuited to `native` status when no connection was explicitly passed. When a caller (e.g. the RAG AI Lab's model-checker) passes an explicit connection on Linux, it fell through to the podman-machine CPU/memory checks and could surface "Update your Podman Machine to improve performance" — a message that doesn't apply on Linux, where Podman runs natively.

Narrowed the guard to trigger on `env.isLinux` alone, so Linux always reports `native` regardless of whether a connection was specified.

Kept the diff minimal: did not add `'native'` to `ContainerConnectionInfoStatus` in the shared types, since it's unused anywhere in the codebase — an unrelated change that got a previous attempt (#4395) maintainer pushback ("How come is this new method changing something").

Fixes #2010

### How to test this PR?
- `pnpm run test:backend` — 531 passed, including a new regression test (`'return native on Linux even when a connection is explicitly provided'`) covering this exact path.
- `pnpm run typecheck:backend` — clean.
- `npx eslint` on the changed files — clean.

### DIFF
--- a/packages/backend/src/managers/podmanConnection.spec.ts
+++ b/packages/backend/src/managers/podmanConnection.spec.ts
@@ -620,6 +620,27 @@ describe('checkContainerConnectionStatusAndResources', () => {
     });
   });
 
+  test('return native on Linux even when a connection is explicitly provided', async () => {
+    const manager = new PodmanConnection(rpcExtensionMock);
+    vi.mocked(env).isLinux = true;
+
+    const result = await manager.checkContainerConnectionStatusAndResources({
+      model: modelMock,
+      context: 'inference',
+      connection: {
+        providerId: 'podman',
+        name: 'podman',
+        type: 'podman',
+        status: 'started',
+        vmType: VMType.UNKNOWN,
+      },
+    });
+    expect(result).toStrictEqual({
+      status: 'native',
+      canRedirect: expect.any(Boolean),
+    });
+  });
+
   test('return noMachineInfo if there is no running podman connection', async () => {
     const manager = new PodmanConnection(rpcExtensionMock);
     vi.mocked(env).isLinux = false;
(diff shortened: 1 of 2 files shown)


xhon-pelushi:
Gentle ping — this has been open since 15 August. The one red check is not caused by this change.

The failing job is `e2e tests smoke`, and the single failing case is

```
✘ src/ai-lab-extension.spec.ts:407 › ... › Run Playground tests for ibm-granite/granite-4.0-micro-GGUF
    › Set system prompt, submit user input, and verify assistant response is visible
expect(locator).toBeEnabled() failed
```

i.e. a 1.3-minute timeout waiting for a local LLM to answer on a GPU-less hosted runner. Every
other check on this PR is green, including the unit tests that actually cover the change.

That suite is red on `main` continuously, not just here — `e2e-tests-main` has failed on every
daily run I can see, 3 September through 16 September without a single pass, always in the same
granite playground block. This morning's `main` run fails on `Delete AI Lab playground` in that
same block, 1 failed / 12 passed / 47 skipped.

The change itself is confined to `checkContainerConnectionStatusAndResources()` in
`packages/backend/src/managers/podmanConnection.ts` plus a test for it. It touches no model,
inference, playground or recipe code, so there is no route from it to a playground response
timing out.

Happy to rebase if that would help, or to split the test out if you would rather review it
separately.

bmahabirbu (approved):
LGTM 

thanks. Ya that test seems to be failing as of recently a fix maybe in the works

